Issue #23, opened 2017-12-04
Need to decide on naming convention for properties: camelCase versus snake_case


trieloff:
In [PR82](https://git.corp.adobe.com/AdobeCloudPlatform/xdm/pull/82) there is a discussion on use of camelCase versus snake_case for property names.  The README clearly calls out use of camelCase, except when the schema is based on an existing standard that uses a different convention.  But it appears that several schemas, including the asset models, use snake_case.

We need to make a decision on which way to go here.  I personally feel very strongly we use camelCase, since that is the generally used convention for JSON, and is the one used by both CDM and schema.org schemas.  It is also what is currently used by all the experience/DMa models.

## What are the schemas that are affected by the issue

All

## What are examples of products that are impacted by the issue

All

trieloff:
Decision has been made in https://git.corp.adobe.com/AdobeCloudPlatform/xdm/issues/90 to use camelCase

PR #33, opened 2017-12-06
Adding Emitted Event

theManikJindal:
## Author Notes
1. Adding `emitted event` to enable the transport of XDM Event Envelope compliant custom OSGI events from AEM. 
2. There's one concern though, which is that the generic nature of the `emitted` event allows for any activity to modeled as a simple `message emitted`. This, however, is a trap that the publishers of the XDM Event Envelope **must** steer away from.

@trieloff I am working on an example of this event type and will add to the PR very soon. I wanted to get the group's initial thoughts on this though, and hence, I am raising this PR.


trieloff:
I agree with the concern that `emitted` is something of a black box and will make it hard to filter what exactly is happening.

Maybe one solution would be to call it `emitted-<whatevertheosgieventnameis>` so that we still have the flexibility of passing anything from OSGi to Adobe Cloud Platform, without having consumers of `emitted` events to do most of the filtering on the client-side.

I'd love to hear @fmeschbe's take on that, too.

theManikJindal:
@trieloff see this - [XDM Event Envelope - Noun and Verb](https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=AdobeCloudPlatform&title=XDM+Event+Envelope#XDMEventEnvelope-NounandVerb)

theManikJindal:
On more thought, I think it would be agreeable to limit the range of the [`objectType`](https://github.com/adobe/xdm/blob/master/schemas/common/eventenvelope.schema.json#L15) field in the `emitted event` to certain predefined values (like an enum). 

But alas, the `objectType` may not be known to or present in XDM and thus it would be grossly limiting. But maybe that is an intended effect?!

trieloff:
I take back what I wrote earlier: yes, we should have fewer verbs, and `objectType` gives us a way to determine, what was wrapped. Carrying OSGi events inside an `EventEnvelope` is a bit like having one envelope inside another: not unheard of, but also not super resource-effective.

theManikJindal:
> Carrying OSGi events inside an EventEnvelope is a bit like having one envelope inside another: not unheard of, but also not super resource-effective.

So, what are you then proposing @trieloff?

trieloff:
Not everything has to be super resource-effective, if it gets the job done and deals nicely with the constraints of established infrastructure.

theManikJindal:
Okay, so then what I am hearing is - restrict the `objectType` to a URI that refers to an OSGI custom event, add an example, and we'd be ready to merge? :)

trieloff:
Just add the example. objectType restriction is optional.

theManikJindal:
@trieloff I added the example for the `emitted` event. 

However, I ran into an issue which needs consideration. The URI for a custom OSGI event was not available, and so I added the URL to its official javadoc. Let me know if this is workable or would you want something else.

trieloff:
One minor thing: given that the description is quite long (good) put it into an external `description.md` for easier authoring.

theManikJindal:
Thanks for reviewing @trieloff!

> OSGi topics and our event types are roughly comparable

I wasn't aware of this, is it so? Also, aren't topics custom? If so, could it give rise to situations where the `topic` name matches in two AEM repositories but they are actually different things in the real world?

trieloff:
Yes, there is a theoretical risk of collisions, but given that OSGi is using Java's naming conventions, and it is best practice to put your domain name into the first two levels of the package name, it is quite unlikely that we will experience collision in practice.

theManikJindal:
Done! Please review @trieloff :)

trieloff:
Ah. I should not have merged this before I've completed the work on #23 – now I have to adjust one more schema and example.

PR #35, opened 2017-12-08
Camel case for #23

trieloff:
Here are the changes to implement #23


trieloff:
@lrosenthol please take a look, too

lrosenthol:
Looks good to me, @trieloff  - thanks for all the hard work on these edits!

trieloff:
@lrosenthol we might want to have another look at some of the XMP properties. If I remember correctly, some start with capital letters, which is not according to style guide, but for consistency's sake I'd follow XMP's lead for properties where XMP has been inconsistent. WDYT?

lrosenthol:
yes, I saw a few of those (eg. ICCProfile) while reviewing, but we definitely want to keep them compatible with XMP (or any other existing schema we are taking from)

trieloff:
ICCProfile rolls just off the tongue, but dmaID isn't bad either ;)

Issue #25, opened 2017-12-04
Missing examples for activity stream schemas


trieloff:
## What are the schemas that are affected by the issue

schemas/external/activity-streams-2/xsd-anyURI.schema.json has 0 examples
schemas/external/activity-streams-2/link-rel.schema.json has 0 examples
schemas/external/activity-streams-2/activity.schema.json has 0 examples
schemas/external/activity-streams-2/mime-media-type.schema.json has 0 examples
schemas/external/activity-streams-2/collection.schema.json has 0 examples
schemas/external/activity-streams-2/xsd-duration.schema.json has 0 examples
schemas/external/activity-streams-2/link.schema.json has 0 examples
schemas/external/activity-streams-2/xsd-nonNegativeInteger.schema.json has 0 examples
schemas/external/activity-streams-2/object.schema.json has 0 examples
schemas/external/activity-streams-2/rdf-langString.schema.json has 0 examples
schemas/external/activity-streams-2/date-time.schema.json has 0 examples
schemas/external/activity-streams-2/xsd-dateTime.schema.json has 0 examples
schemas/external/activity-streams-2/uri.schema.json has 0 examples
schemas/external/activity-streams-2/person.schema.json has 0 examples
schemas/external/activity-streams-2/service.schema.json has 0 examples
schemas/external/activity-streams-2/group.schema.json has 0 examples
schemas/external/activity-streams-2/image.schema.json has 0 examples
schemas/external/activity-streams-2/type.schema.json has 0 examples
schemas/external/activity-streams-2/page.schema.json has 0 examples
schemas/external/activity-streams-2/xsd-string.schema.json has 0 examples
schemas/external/activity-streams-2/document.schema.json has 0 examples
schemas/external/activity-streams-2/id.schema.json has 0 examples
schemas/external/activity-streams-2/organization.schema.json has 0 examples
schemas/external/activity-streams-2/language-tag.schema.json has 0 examples
schemas/external/activity-streams-2/collection-page.schema.json has 0 examples
schemas/external/activity-streams-2/context.schema.json has 0 examples
schemas/external/activity-streams-2/application.schema.json has 0 examples

Issue #14, opened 2017-12-04
Activity Steams Schemas are Messy


trieloff:
- missing titles
- missing descriptions
- circular references (making @ankaur's life difficult)

@majindal can you help improving this?

trieloff:
@theManikJindal once you've done this, I'm happy to take a PR that rejects PRs without examples

theManikJindal:
@trieloff I am just on the verge of raising a PR that fixes this issue. However, there are some places where the tooling is failing. 

For example, if one defines a schema having an array at the root, the tooling throws an error. Tooling wants only `objects` to be at the root level. 

Example schema: 
```
{
    "type": "array",
}
```

Example instance:
```
["fails", "to", "validate", "this"]
```

trieloff:
Change the meta schema to allow arrays.

Issue #5, opened 2017-12-04
Clean up double-definitions in Activity Streams


trieloff:
@majindal in the `activity-streams` schemas, there are a number of definitions imported from RDF and XSD that largely mirror existing JSON Schema constructs. We would benefit from having this simplified by directly using the corresponding JSON Schema constructs and to remove the superfluous type definitions.

PR #36, opened 2017-12-08
Fixing Activity Stream 2 schemas

theManikJindal:
### Fixes all issues related to AS2 - 
+ https://github.com/adobe/xdm/issues/5 - Superfluous definitions
+  https://github.com/adobe/xdm/issues/14 - Missing titles and descriptions
+  https://github.com/adobe/xdm/issues/25 - Missing examples

### Notes:
1. Some schemas still do not have examples, because they are not schemas for objects. Rather they are schemas for array and strings with patterns.
2. Have removed a few schema files that were not being used. This is separate from issue#5.

_Pro Tip: Add query parameter `w=1` to the URL of the PR to show only non-whitespace diff._


trieloff:
Almost:

```bash
$ find . -name "*.schema.json" | xargs grep -e "description\": \"http" | wc -l
46
```

46 cases of descriptions that are nothing more than a URL, not even linked.

theManikJindal:
Oh, I missed these. They were not at the root level. Let me add those ~tomorrow~ on Monday.

Issue #15, opened 2017-12-04
Patterns for Schema extensibility/inheritance


trieloff:
While we are waiting for #43 to come to fruition, I wanted to gather my thoughts for the second class of extensibility that is not covered in #43: introducing new schemas that "extend" core XDM schemas, allowing XDM consumers to know that a certain schema is compatible with the extended schema.

- `definitions` work like mixins: you can combine multiple to build a complex schema
- the `properties` at the root of a schema define a concrete schema, but limit re-usability
- best practice for extensible schemas: use `definitions` and `$ref` them using `allOf`
- best practice for non-extensible schemas: don't use `definitions`, put everything into the root of the schema
- best practice for schema fragments (not directly instantiable): only use `definitions` but don't `$ref` them in the current schema
- adjust the tooling to discover the three patterns and warn about deviations

trieloff:
@kstreeter:
> Hi @trieloff, thank you for opening this issue.  Is it true to say that when we say "best practice", it is not a normative requirement for a valid model, and just a recommendation on how to structure the model for style and maintainability?
>
> I ask this because, as I mentioned in the comments for [data lineage](https://git.corp.adobe.com/AdobeCloudPlatform/xdm/commit/002c6d59cb3e5e67b13aba9c806364831f1912a6) a problem with assuming a particular normalized structure for the JSON Schema is that it prevents services from denormalizing the schema on behalf of a client.  
> 
> Our XDM registry implementation does this today: pulls all the references together into a single schema, so that the client does not have to do that operation, which is non-trivial.  This is why we have proposed to have explicit metadata that describes the lineage/compatibility of a model.
> 
> I feel like in general we should stay away from making a lot of normative requirements about the structure of the JSON Schema.  We can certainly propose best practices and style, but encoding semantic meaning into a specific way we use JSON Schema seems very fragile.  Hopefully that is not the intention here.

trieloff:
My approach would indeed have gone down the way of restricting the patterns we endorse for schema authors and apply implicit semantics to it. 

I see the introduction of custom keywords like `meta:derived_from` as another way to achieve the same goal and I'm not partial to solving it in any specific way.

In the `jsonschema2md` project we already have some schema-pre-processing to handle things like `description.md` files, and I expect this to grow, so that in the end we author one type of schema files with specific conventions and some custom keywords, which are then published as plain vanilla JSON Schema files.

[...]

PR #39, opened 2017-12-12
#15 Extensibility through new types

trieloff:
This proposal is a result of @kstreeter and my ideas on extensibility. We explicitly declare extensibility and intent to extend through the `meta:extends` keyword and then use the `meta.schema.json` to ensure that this intent is implemented through the correct JSON Schema constructs, so that every extended instance will validate against all extending schemas.


trieloff:
Requesting reviews from @lrosenthol and @kstreeter

kstreeter:
Hi @trieloff, I have been away and am getting back to this.  I think the idea of explicitly declaring both "extensible" and "extends" in addition to the JSON-schema level idioms to describe that situation is great.  This will allow us to build tooling that checks the schema-level constructs and ensures they reflect the meta-level declarations.  Also, great examples!

One thing that did come up during conversations I had with a few folks regarding this mechanism was how it related to schema versioning.  I had claimed that we don't need a specific "version" for schemas, as long as they followed the "addition only" rule and were marked as extending the previous version of the schema.

The concern with this is that you could potentially use the "extends" mechanism to build graphs, branching from a common schema and then branching again.  This is not nearly as clear as simply saying "this is version X of schema Y".

I think this is a valid point; you could easily introduce complex relationships using this mechanism. 
 To address this, we may want to define a guidelines/best practice around only extending our standard models linearly, and strongly discouraging creating graphs.

trieloff:
You might have seen that I've spent some time with the `npm run lint` command which is getting increasingly opinionated (as @theManikJindal can tell). I believe that we can use this to discourage the use of constructs that we find difficult to handle until we have gathered more experience with this.

In addition, `npm run lint` is now doing a primitive version check, forcing you to increase the version number of XDM if you change the number of schemas included, but sooner or later we will have to think about how different schema versions will be published and referenced.

The biggest risk in my mind are extensions that are incompatible with prior versions of themselves. Building tooling that compares two JSON Schemas and asserts whether *A includes B* or *B includes A* or both would allow us to help extension authors and build more robust systems that deal with incompatible extensions, so that we can either reject the new schema version, wipe old data, or come up with more intelligent migration strategies.

PR #48, opened 2017-12-18
Adding the `Rejected` Event Type

theManikJindal:
## Author Notes

On the same lines as the passive CRUD events added in [a previous PR](https://git.corp.adobe.com/AdobeCloudPlatform/xdm/pull/81), I am adding the `Rejected` event type.

The `Rejected` event is a past tense form of the [`Reject` Activity](https://www.w3.org/TR/activitystreams-vocabulary/#dfn-reject) and represents the act of an `actor` rejecting the `object`.

### DIFF
--- a/schemas/common/event/rejected.example.1.json
+++ b/schemas/common/event/rejected.example.1.json
@@ -1,34 +1,34 @@
 {
-    "@id": "https://events.adobe.io/events/id/82235bac-2b81-4e70-90b5-2bd1f04b5c7b",
-    "activitystreams:published": "2016-07-16T19:20:30+01:00",
-    "activitystreams:to": {
-        "https://ns.adobe.com/xdm-extensions/ims/user#id": "C74F69D7594880280A495D09@AdobeID",
-        "@type": "https://ns.adobe.com/xdm-extensions/ims/user"
+  "@id": "https://events.adobe.io/events/id/82235bac-2b81-4e70-90b5-2bd1f04b5c7b",
+  "activitystreams:published": "2016-07-16T19:20:30+01:00",
+  "activitystreams:to": {
+    "https://ns.adobe.com/xdm-extensions/ims/user#id": "C74F69D7594880280A495D09@AdobeID",
+    "@type": "https://ns.adobe.com/xdm-extensions/ims/user"
+  },
+  "@type": "https://ns.adobe.com/xdm/common/event/rejected",
+  "xdm:objectType": "https://stock.adobe.com/content/json/2017-08-31",
+  "activitystreams:actor": {
+    "name": "Adobe Stock Ingestion Service",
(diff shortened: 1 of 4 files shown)


trieloff:
Run `npm test` before you make a PR.

theManikJindal:
Fixed it now. It was not a validation error, but linting one.

schemas/common/event/rejected.schema.json
```diff
@@ -0,0 +1,26 @@
...
+    "Copyright 2017 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "$id": "https://ns.adobe.com/xdm/common/event/rejected",
+  "type": "object",
+  "title": "Rejected Event",
+  "description": "A `rejected event` follows semantics that are exactly equivalent to the semantics described by the [`reject activity`](https://www.w3.org/TR/activitystreams-vocabulary/#dfn-reject), except for one difference. The `reject activity` can be used in imperative contexts, however, the `rejected event` can only be used in passive contexts.",
```
trieloff: I don't get it. Is there a way to re-phrase the description? If it is getting longer, a separate file might be useful.

What are examples of objects that can be rejected? I can see the Adobe Stock example, but the description should list additional examples.

schemas/common/event/rejected.example.1.json
```diff
@@ -0,0 +1,34 @@
+{
+  "@id": "https://events.adobe.io/events/id/82235bac-2b81-4e70-90b5-2bd1f04b5c7b",
+  "activitystreams:published": "2016-07-16T19:20:30+01:00",
+  "activitystreams:to": {
+    "https://ns.adobe.com/xdm-extensions/ims/user#id": "C74F69D7594880280A495D09@AdobeID",
+    "@type": "https://ns.adobe.com/xdm-extensions/ims/user"
+  },
+  "@type": "https://ns.adobe.com/xdm/common/event/rejected",
+  "xdm:objectType": "https://stock.adobe.com/content/json/2017-08-31",
```
trieloff: Is that an asset? Should it be?

theManikJindal: It is not an asset. It is sort of a job specification file, whose schema and definition ~lives~ will live at the URL specified by the value of the `xdm:objectType` field.

schemas/common/event/rejected.schema.json
theManikJindal: I have added a better description. Please suggest any other changes, if needed.

theManikJindal:
This looks great! 👍

PR #49, opened 2017-12-18
Fix #46 - Typo in gpsAltitudeRef

### LINKED ISSUE #46: Typo in gpsAltitudeRef
https://github.com/adobe/xdm/blob/jsonschema2md-example/docs/reference/assets/asset.schema.md#exifgpsaltituderef-known-values

Missing a backtick in meta:enums

man0s:
- Fixed typo in gpsAltitudeRef

### DIFF
--- a/docs/reference/assets/asset.schema.md
+++ b/docs/reference/assets/asset.schema.md
@@ -1037,7 +1037,7 @@ The value of this property **must** be equal to one of the [known values below](
 | Value | Description |
 |-------|-------------|
 | `0` | The `gpsAltitude specifies a value above sea level` |
-| `1` | The `gpsAltitude specifies a value below sea level |
+| `1` | The `gpsAltitude specifies a value below sea level` |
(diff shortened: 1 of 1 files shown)


trieloff:
The files under `/docs/reference` get generated automatically from the schema files, in this case `asset.schema.json` – you need to apply the fix to that file.

The role of the backticks (`) is to indicate the start and end of a named property, in this case `gpsAltitudeRef`. The problem in #46 is that the backtick in the first value is at the end of the line, not at the end of the property value, and for the second line the closing backtick is missing.

trieloff:
@man0s in the future, you don't have to close a pull request that has gotten comments or requests for changes, you can just continue making changes in the branch that the pull request refers to. This is less work than re-opening a new PR for every iteration and it makes it easier to follow the conversation thread.

man0s:
@trieloff Noted..thanks!

PR #53, opened 2018-01-03
#Issue 44: dc:format is missing a type declaration

kgera:
Adding type to dc:format and required changes made to change type to string.

PR for :adobe#44 issue


trieloff:
This should go into master.

trieloff:
@kgera can you re-do this? Instead of fixing the output (Markdown) edit the schema. Then (if you want to) regenerate the documentation.

Issue #13, opened 2017-12-04
Is Asset Rating user editable?


trieloff:
Right now, it is marked as not user editable, but that conflicts with the description that says the asset rating has been set by a user.

lrosenthol:
Yes, it should be.  I'll try to submit a PR for the change

trieloff:
Cool. The semantics of being user editable are a bit hazy anyway, so clarifying this somewhere in `docs` would also be beneficial.

trieloff:
https://github.com/adobe/xdm/blob/master/schemas/assets/asset.schema.json#L285 if you are looking for a quick fix.

kgera:
submitted PR: https://github.com/adobe/xdm/pull/54 for review.

PR #54, opened 2018-01-03
[trivial] #13: Is Asset Rating user editable?

kgera:
Fix for #13 to make user rating editable.


trieloff:
Looks good. Thank you.

Issue #59, opened 2018-01-12
Add XDM schemas from data platform


jwen-adobe:
Add more xdm schemas to the "common" folder and create a new folder "experience" to hold experience schemas.

PR #60, opened 2018-01-12
#59 Add XDM schemas from data platform

jwen-adobe:
Link to the issue #59 https://github.com/adobe/xdm/issues/59

@trieloff @kstreete Please review and comment.

The following are included in this PR.

(1). Added more common schemas from platform to the "common" folder, these are referenced by the experience xdm models/schemas from (3).
(2). Moved person related schemas from "context" to "common" folder.
(3). Added new folder "experience" to hold experience xdm models.


lrosenthol:
1) a *LOT* of this stuff doesn't belong in common, as it is experience cloud-specific stuff.
2) there are a bunch of things here that are just not correct or valid outside of EC, for example an "application" as defined here is *not* what I would define one as.
3) this is a *HUGE* commit and will take a lot of time to review in detail.

jwen-adobe:
@lrosenthol ok, i know this is a HUGE commit, let me close this one for now and commit maybe five schemas a time for reviews.

Issue #63, opened 2018-01-12
Clean Up Audit Trail


trieloff:
I've made some cosmetic changes to Audit Trail already in  8e4de5f, but I think more is needed.

1. Audit Trail right now is not a trail, it's simply a last modified date (+etc.). This is an *inaccuracy*
2. Just like assets and content have the last modified date (+etc.) they should be part of the core entity. This is an *inconsistency*
3. Audit Trails are referencing a "batch ID", but what that batch ID points to and what a "batch" looks like is never specified. This is an *omission*.

We should fix (3) right away, by modeling a batch or bulk update (maybe treat the bulk update as an Event Envelope) and decide if we want to go down the route of (1), i.e. make AuditTrail an `array` or (2), i.e. drop the `xdm:audit` property and instead merge the audit trail properties into the schemas that currently have the `xdm:audit` property.

@kstreeter and @jwen-adobe, your feedback is appreciated.

## What are the schemas that are affected by the issue

In branch `experience-cloud-staging`:
- address
- audit trail
- person
- email address
- phone number
- profile

jwen-adobe:
@cmathis @cdegroot

kstreeter:
Hi @trieloff.  I agree that if it is not a trail we should not call it a trail.  Perhaps we just call it "audit record" or similar.  

I agree that having the modified time, etc in a different object is distinct from how that is handled for assets, but I think that is intentional.  For content, those properties tend to be integral: all content have them, and it would rarely make sense to have a representation of content that did not.  But for the data oriented models this is not true.  The audit data will tend to be layered on top of the original data by a different system then what generated it.  So it makes sense to layer this into a different node.  

Batch IDs are an abstract concept, and in the description they are URIs.  In the data platform, we will use that to refer to batches of data in the data lake, but other systems may have their own method of designating a batch.  We could provide more details on how to use this field, but as a URI it means implies it is an ID, and the ID space is managed as all URIs are.

trieloff:
Ok, then I would turn Audit Trail into an Audit Record mixin schema, that can optionally be added to all data entities that support batch ingestion.

In addition, I would add a `meta:batchable` flag to all schemas that use it, so that the `jsonschema2md` documentation can easily show schemas that support batch ingestion.

PR #69, opened 2018-01-15
Auditability schema (the artist formerly known as Audit Trail)

trieloff:
See #63.

@kstreeter this is a pretty common dependency for many `common` schemas, so if there are no objections, I'll merge this tomorrow.


trieloff:
Just noticed, that this depends on #73, too – I won't merge before #73 is in master.

Issue #28, opened 2017-12-04
Model JSON HAL (Hypertext Access Language)


trieloff:
## What are the schemas that are affected by the issue

The directory schema in #96 depends on the `_link` property, which is following https://tools.ietf.org/html/draft-kelly-json-hal-08 - JSON HAL is not modeled at the moment.

## What are examples of products that are impacted by the issue

- Creative Cloud Files

lucianfelix:
Hi,

I'm woking at the CICD api proposal and I need some guidance about how HAL applies to this.

How can we link to HAL schema definitions? Link an external one like [hyperschema.org](http://hyperschema.org/mediatypes/hal) or use our own (which would be proposed in a PR)?

Should the schemas also use _embedded part of HAL? There are schemas in the current model that use children as container for parent-child relations.

Are the links the only way of specifying relations or is it allowed to have redundant <rel_name>_id properties that point to liked object?

Thanks,
Lucian

trieloff:
Hi @lucianfelix 

> How can we link to HAL schema definitions? Link an external one like hyperschema.org or use our own (which would be proposed in a PR)?

We should have our own, residing in `schemas/external/hal` with proper description and examples. You can use an external definition as a starting point, if the license permits and is compatible.

trieloff:
> Should the schemas also use _embedded part of HAL? There are schemas in the current model that use children as container for parent-child relations.

Only model `_embedded` if you need it. As I see it right now, I would not replace all (or even most) of the nested documents we have in our current schemas with HAL `_embedded` relations.

trieloff:
> Are the links the only way of specifying relations or is it allowed to have redundant <rel_name>_id properties that point to liked object?

We do already have examples where we are linking (or pseudo-linking) through `somethingID` properties. Most of this is in `assets` where we point to old `versionID`s. For new stuff, I'd uniformly use HAL, so that this style of linking can be expected in most places.

trieloff:
This would also eliminate the need for the `reference` schema that got introduced with #59 and #60

PR #70, opened 2018-01-15
JSON HAL External Schema

trieloff:
PR for #28, allowing us to express relationships between resources in a more standardized way (and the same way the Adobe Cloud Platform is prescribing)


fmeschbe:
See also https://git.corp.adobe.com/pages/experience-platform/api-guidelines/6-json-standardizations.html#links

I wonder, whether it is a stable assumption to refer to and re-use an expired and archived draft ? Maybe we should just define our own `_links` specification (as down in above API guidelines) ?

lucianfelix:
Could you also add an example with ```_embedded```? It's interesting as it has implications on collection resources.

trieloff:
The link above is basically a subset of HAL, with the added disadvantage that we don't call it HAL, and that users won't get the idea of using things like https://github.com/mikekelly/hal_specification/wiki/Libraries that are compatible with it, whether we call it this or not.

trieloff:
@lucianfelix – I could, but I just noticed that the `_embedded` schema is too restrictive.

fmeschbe:
> The link above is basically a subset of HAL

@trieloff yes, it is derived from the API spec with some added properties.

In any case, that was just a reference. The main point is whether we want to actually refer to a defunct draft or whether it would be better to refer to the actual spec https://github.com/mikekelly/hal_specification (which also has not been updated in a long time) and work with the author to come up with a proper schema.

I am not feeling very confident with defining a JSON Schema in a foreign URL space.

trieloff:
Regarding the HAL status: https://groups.google.com/forum/#!topic/hal-discuss/WrI5YDQo610 (tl;dr: not dead)

There are a number of schemas for HAL already. I did not copy any of them for following reasons:

1. either they don't have a license
2. or the license is not CC (e.g. https://github.com/DaveJS/dave.schema.json is MPL)
3. and the schema not structured in a way that works well with our documentation

We have created JSON Schemas for a number of other standards, without implying that this JSON Schema would have a reach beyond XDM, so I'd assume to treat HAL the same.

trieloff:
Comment from @fmeschbe via Slack:

> was mich dann nur noch stört ist die ID, welche Du einem non-Adobe namespace hast…

> The only remaining thing I object to is that your `$id` points to a non-Adobe Namespace, although we only interpret HAL and don't use any official schema.

Issue #71, opened 2018-01-15
Create an Access Control Policy Schema


fmeschbe:
Reflect resource-based access control policies, currently only access control lists, in an XDM model. The access control policy, technically speaking is something like:

```
acl = "[" 1#ace "]"
ace = "{"  principal "," privilege "," relations ","" modifier "}"
```

PR #72, opened 2018-01-15
Access Control Policy Schema

fmeschbe:
See #71.

Proposing:

* `accesscontrolpolicy.schema.json`
* `accesscontrolpolicy.example.1.json`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "xdm",
-  "version": "0.2.1",
+  "version": "0.3.0",
   "description": "Experience Data Models",
   "main": "",
   "config": {
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "aem_user": "packageUser",
     "aem_password": "override me securely",
     "markdown-importer-version": "0.0.4",
-    "schemas": 49
+    "schemas": 50
   },
   "scripts": {
     "clean": "rm -rf docs/reference",
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "aem_user": "packageUser",
     "aem_password": "override me securely",
     "markdown-importer-version": "0.0.4",
-    "schemas": 50
+    "schemas": 51
   },
   "scripts": {
     "clean": "rm -rf docs/reference",
--- a/schemas/common/accesscontrolpolicy.example.1.json
+++ b/schemas/common/accesscontrolpolicy.example.1.json
@@ -1,0 +1,14 @@
+ {
+  "acl": [
+    {
+      "principal": "6A29340459CA8D350A49413A@AdobeOrg:87611788",
+      "privilege": "read",
+      "modifier": "grant"
+    },
+    {
+      "principal": "6A29340459CA8D350A49413A@AdobeOrg:58203452",
+      "privilege": "write",
+      "modifier": "grant"
+    }
+  ]
+}
--- a/schemas/common/accesscontrolpolicy.example.1.json
+++ b/schemas/common/accesscontrolpolicy.example.1.json
@@ -1,14 +1,23 @@
- {
-  "acl": [
+{
+  "repo:acl": [
     {
-      "principal": "6A29340459CA8D350A49413A@AdobeOrg:87611788",
-      "privilege": "read",
-      "modifier": "grant"
+      "repo:principal": "authenticated",
+      "repo:privilege": "read",
+      "repo:modifier": "grant"
     },
     {
-      "principal": "6A29340459CA8D350A49413A@AdobeOrg:58203452",
-      "privilege": "write",
-      "modifier": "grant"
+      "repo:principal": {
+        "xdm:provider": {
+          "@id": "https://ims-na1.adobelogin.com/"
+        },
+        "@id": "C0B648DE57D701277F000101@AdobeID",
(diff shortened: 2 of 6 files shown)


fmeschbe:
@trieloff For review

... though lint.js fails with

```
schemas/common/accesscontrolpolicy.schema.json is not pretty. Fixing.
schemas/common/accesscontrolpolicy.example.1.json is not pretty. Fixing.
```

which is the only code I get.

trieloff:
@fmeschbe run `npm install & npm run lint` locally. The first time `npm run lint` runs, it will fix the indentation and you can just commit the updated files.

package.json
```diff
@@ -7,7 +7,7 @@
     "aem_user": "packageUser",
     "aem_password": "override me securely",
     "markdown-importer-version": "0.0.4",
-    "schemas": 49
+    "schemas": 50
```
trieloff: Also bump the version number.

schemas/common/accesscontrolpolicy.example.1.json
```diff
@@ -0,0 +1,14 @@
+ {
+  "acl": [
```
trieloff: These look like they will need prefixes.

schemas/common/accesscontrolpolicy.schema.json
```diff
@@ -0,0 +1,75 @@
...
+    "acl": {
+      "title": "Access Control List",
+      "description": "List of Access Control Entries defining access to this resource",
+      "type": "array",
+      "items": { "$ref": "#/definitions/ace" }
+    },
+    "ace": {
+      "type": "object",
+      "properties": {
+        "principal": { "$ref": "#/definitions/principal" },
```
trieloff: I'd suggest splitting each of these definitions into separate files, so that you have one schema for each `object`.

schemas/common/accesscontrolpolicy.schema.json
```diff
@@ -0,0 +1,75 @@
...
+  "$id": "https://ns.adobe.com/xdm/common/accesscontrolpolicy",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Access Control Policy",
+  "description": "Schema for the Access Control Policy of a resource",
+  "type": "object",
+  "definitions": {
+    "accesscontrolpolicy": {
+      "type": "object",
+      "properties": {
+        "acl": { "$ref": "#/definitions/acl" }
```
trieloff: Use `xdm` or `repo` as prefix for all properties. If you want to use `repo`, we'll have to add it to the meta-schema files.

schemas/common/accesscontrolpolicy.schema.json
```diff
@@ -0,0 +1,75 @@
...
+          "enum": [ "authenticated", "unauthenticated", "all" ]
+        }
+      ]
+    },
+    "privilege": {
+      "description": "Privilege granted or denied to the Principal",
+      "type": "array",
+      "items": {
+        "type": "string",
+        "enum": [ "ack", "read", "write", "modify", "attach", "delete", "full" ]
```
trieloff: You can use `meta:enum` to document what the values mean.

schemas/common/accesscontrolpolicy.schema.json
```diff
@@ -0,0 +1,75 @@
...
+    "principal": {
+      "anyOf": [
+        {
+          "description": "Reference to the User or Group to which the entry applies",
+          "type": "string",
+          "format": "uri"
+        },
+        {
+          "description": "A special principal to which the entry applies. 'Authenticated' refers to any authenticated user, 'Unauthenticated' refers to an unknown user, 'all' refers to any user regardless of authenticated or not.",
+          "enum": [ "authenticated", "unauthenticated", "all" ]
```
trieloff: `meta:enum` makes nice documentation like this: https://github.com/adobe/xdm/blob/master/docs/reference/assets/asset.schema.md#xdmnotsafe-known-values

schemas/common/accesscontrolpolicy.schema.json
fmeschbe: ack

schemas/common/accesscontrolpolicy.schema.json
fmeschbe: ack

schemas/common/accesscontrolpolicy.schema.json
fmeschbe: Ok. And this principal would probably even be an extension to the [existing principal schema](https://github.com/adobe/xdm/blob/master/schemas/common/principal.schema.json).

trieloff: Makes sense.

schemas/common/accesscontrolpolicy.example.1.json
trieloff: Yes, we have three kinds of properties:

1. core XDM properties (they have a `:` somewhere)
2. extension properties (they look like URLs)
3. special sauce properties (`@id`, `@type`, and maybe `_links` and `_embedded`)

Let's say, I like the consistency of it.

fmeschbe: ack

PR #73, opened 2018-01-16
Include Repository Schemas from Adobe Cloud Platform + HAL

trieloff:
No longer a work in progress. The PR will model the directory and asset representations of ACP, and use them as base classes for XDM asset and XDM content.

This PR subsumes PR #70.

Quick summary of changes:

- introduces HAL for expressing links in `schemas/external/hal`
- introduces `schemas/external/repo` for ACP storage specific schemas
- models ACP storage, i.e. `asset`, `directory`, and `sub-directory` according to ACP spec
- changes existing `content` and `asset` to inherit from ACP `asset`
- apply trickle-down changes to everything that inherits from `asset`, i.e. `image` and `video`
- apply trickle-down changes to all examples

@ksteeter please have a look.


schemas/external/repo/directory.schema.json
```diff
@@ -0,0 +1,63 @@
...
+  ],
+  "definitions": {
+    "directory": {
+      "properties": {
+        "dc:format": {
+          "type": "string",
+          "title": "Format",
+          "const": "application/vnd.adobecloud.directory+json"
+        },
+        "children": {
```
lucianfelix: Shouldn't this belong to HAL _links? Or to _embedded if simple children representation needs to be included?

schemas/external/repo/asset.example.1.json
```diff
@@ -0,0 +1,11 @@
+{
```
lucianfelix: Would be useful to include HAL properties in some examples.

schemas/external/repo/directory.schema.json
trieloff: Not according to [ACP spec](https://git.corp.adobe.com/pages/AdobeCloudPlatform/api-spec/repository_api/hypermedia.html), which I just tried to model here.

schemas/external/repo/asset.example.1.json
trieloff: ✅ f39bc9e

Issue #81, opened 2018-01-19
Model Digital Experience Channels


trieloff:
We need a way to model digital experience channels, so that the core characteristics of each channel can be captured in the OOTB schema, but that evolution of channels and the introduction of new channels is possible.

## What are the schemas that are affected by the issue

- all Experience Cloud schemas that track channel delivery
- all Creative Cloud schemas that track appropriateness of content for a channel

PR #82, opened 2018-01-19
A consistent way to model channels

trieloff:
As per #81, I had a look at the channel definition in `experience-cloud-staging` and found it lacking in a couple of respects. My version improves this in following dimensions:

- new channels can be added easily without having to change the core channel definition
- all channels are addressed consistently, in that they have a *required* `@id` that points to the schema
- channels are described consistently with properties `@type` (for the rough group), `xdm:mode` (push or pull), as well as (optionally) a list of content types, metric types and location types supported by the channel.

The result is very compact and readable:

**old**
```json
{
    "xdm:types": {
      "xdm:socialPlatform": true
    },
    "xdm:socialPlatform": "facebook"
}
```

**new**
```json
{
  "@id": "https://ns.adobe.com/xdm/channels/facebook-feed",
  "@type": "https://ns.adobe.com/xdm/channel-types/social"
}

```

Please note that `@type` is optional and can be omitted.

I have included all types that were listed in the `experience-cloud-staging` branch, although in the future we might want to add `tv`, `video`, `radio`, `podcast`, etc.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,11 +7,7 @@
     "aem_user": "packageUser",
     "aem_password": "override me securely",
     "markdown-importer-version": "0.0.4",
-<<<<<<< HEAD
     "schemas": 79
-=======
-    "schemas": 61
->>>>>>> master
   },
   "scripts": {
     "clean": "rm -rf docs/reference",
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "aem_user": "packageUser",
     "aem_password": "override me securely",
     "markdown-importer-version": "0.0.4",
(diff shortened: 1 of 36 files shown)


kstreeter:
@trieloff, this PR is a little hard to process because it does not show the deltas from the existing definitions.  I think you need to merge those into your branch so they show up in the PR as diffs.

trieloff:
This would not affect anything we already have in `master`, (so there are no deletions), but affect how we work in `experience-cloud-staging`, where we would remove `communicationchannel` in favor of the new, more general `channel` schema.

PR #83, opened 2018-01-22
Introducing an aggregated asset schema

trieloff:
As discussed on the mailing list with @ogoldman and Michael B: this is an asset schema that aggregates all schemas that an asset can take. 

While I can see the value of this for ACP, I'm not sure about its use for XDM itself and would value feedback from @kstreeter and @lrosenthol


schemas/assets/aggregated-asset.schema.json
```diff
@@ -0,0 +1,48 @@
...
+      "$ref": "https://ns.adobe.com/xdm/content/content#/definitions/content"
+    },
+    {
+      "$ref": "https://ns.adobe.com/xdm/assets/copyright-owner#/definitions/copyright"
+    },
+    {
+      "$ref": "https://ns.adobe.com/xdm/assets/asset#/definitions/asset"
+    },
+    {
+      "$ref": "https://ns.adobe.com/xdm/assets/asset#/definitions/asset"
```
fmeschbe: This looks duplicate ;)

lrosenthol:
With the additional properties flag added, I think this moves us along.  We still need to determine which schemas get listed (and which don't), but thats secondary and we can figure out that out as we go forward.

trieloff:
I might have made the PR, but I'm still unsure of its *use* in the context of XDM. Is this the `asset` schema that we would reference from `directory` or is there another role for it?

trieloff:
@lrosenthol @kstreeter can you comment on my question above: https://github.com/adobe/xdm/pull/83#issuecomment-360716976

Why do we need this?
Where would we use it?

PR #86, opened 2018-01-22
Add _page to ACPAPI Directory resource proposal

tdahlgre:
@mburbidgadobe, @trieloff, @lopperma, others - This WIP change proposal adds support for _page for Directory resource per ACPAPI specification.

See inline notes for additional details, and Platform API specification for pagination support (I'm not posting links or copies of internal references to this pull in public git location).

npm test passes locally with these changes, includes new directory sample and valid/invalid json examples for newly added _page object schema.

First time working in XDM so any feedback, guidelines and proposals welcome to make this good!

Questions:
- should there be git issue created for this (there is internal backlog item in Adobe backlog)
- should there be copy of the same changes in docs/ directories (I see some, mostly duplicate, or is it somehow out-generated)?
-- Tuomas: got answer to this, no need to, it will be automatically generated.


schemas/external/repo/directory-page.example.1.json
```diff
@@ -0,0 +1,7 @@
+{
```
tdahlgre: sample with minimal required properties

schemas/external/repo/directory-page.example.2.json
```diff
@@ -0,0 +1,8 @@
+{
```
tdahlgre: sample with optional properties included

schemas/external/repo/directory-page.invalid.1.json
```diff
@@ -0,0 +1,6 @@
+{
```
tdahlgre: invalid example, required 'count' missing.

schemas/external/repo/directory-page.schema.json
```diff
@@ -0,0 +1,67 @@
+{
```
tdahlgre: page-object as defined in ACPAPI for directories. This is now separate schema file located in external/repo folder, could also be defined in separate folder like 'hal' links are or moved to some other 'common' location as well ... feedback and proposals welcome.

schemas/external/repo/directory-page.schema.json
```diff
@@ -0,0 +1,67 @@
...
+    }
+  },
+  "allOf": [
+    {
+      "$ref": "#/definitions/page"
+    }
+  ],
+  "required": [
+    "orderBy",
+    "order",
```
fmeschbe: The order property is optional and actually is not officially listed in the [Sorting, Paging, and Filtering](https://git.corp.adobe.com/pages/AdobeCloudPlatform/api-spec/#sorting-paging-and-filtering) Spec any more. Rather its functionality has been replaced by the `orderby`  property supporting multiple values with optional `+/-` to indicate sort order.

Not sure, maybe we should remove this property alltogether.

schemas/external/repo/directory-page.schema.json
```diff
@@ -0,0 +1,67 @@
...
+  ],
+  "$id": "http://ns.adobe.com/adobecloud/core/1.0/directory-page",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Directory Page Object",
+  "type": "object",
+  "description": "Directory page object contains paging information for related directory when directory result is paginated.",
+  "definitions": {
+    "page": {
+      "properties": {
+        "orderBy": {
```
fmeschbe: The API spec has this all lowercase as `orderby`.  As we have naming convention to be lowerCamelCase for consistency `orderBy` would be correct.
@ogoldman, @bpruessm what do you think ?

trieloff:
The `docs` will get auto-generated once it hits master. You don't need to include it in your PR, but you can run `npm run markdown` to see what it would look like.

schemas/external/repo/directory-page.schema.json
```diff
@@ -0,0 +1,67 @@
...
+    "Copyright 2018 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "http://ns.adobe.com/adobecloud/core/1.0/directory-page",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Directory Page Object",
+  "type": "object",
+  "description": "Directory page object contains paging information for related directory when directory result is paginated.",
```
fmeschbe: I think we should not called this with any relation to *Directory* as it is also used vor the `versions` resource and IMHO is a generic model for paging collection-style objects.

fmeschbe:
As I suggest this to be a generic model, we might even consider moving it to the `schemas/common` folder, for example ?

schemas/external/repo/directory-page.schema.json
tdahlgre: @fmeschbe I do see 'order' listed in current spec, and I assumed that all the properties that do not have 'if any' are mandatory. I can removed or make this optional, and matching changes should be applied to spec as well.

Here's copy of current spec and page-object definition;
image

schemas/external/repo/directory-page.schema.json
tdahlgre: Will move page as common definition.

schemas/external/repo/directory-page.schema.json
fmeschbe: Thanks. I suggest to make at least optional for now.

fmeschbe:
@trieloff I suggested to move this schema to `schemas/common` as I think it is not restricted to `extensions/repo` only.

Please advise whether this is correct/ok or what would be better ?

Thanks.

schemas/external/page/page.schema.json
```diff
@@ -0,0 +1,67 @@
...
+        }
+      }
+    }
+  },
+  "allOf": [
+    {
+      "$ref": "#/definitions/page"
+    }
+  ],
+  "required": [
```
tdahlgre: order now updated to be optional per recommendation from @fmeschbe.

schemas/external/page/page.schema.json
```diff
@@ -0,0 +1,67 @@
+{
```
tdahlgre: @trieloff - this schema is now in schemas/external/page folder (following standard seen with schemas/external/hal to have folder per common type. @fmeschbe proposed moving this to schemas/commons folder to be even more generic. Please advise if that would be desired location and I'll move it there.

trieloff: I think common sounds right. It's not really an external standard (like HAL that we re-implemented), but very common.

trieloff:
@fmeschbe please take another look. 

I won't be able to merge earlier than Friday, because I'm at a conference tomorrow.

schemas/common/page.schema.json
```diff
@@ -0,0 +1,67 @@
+{
+  "meta:license": [
+    "Copyright 2018 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/common/page",
```
tdahlgre: @fmeschbe, @trieloff now page definition moved to schemas/commons folder and id reference updated to 'xdm/common' like used in other definitions in schemas/commons.

schemas/external/repo/directory.example.2.json
```diff
@@ -0,0 +1,25 @@
...
+    "http://ns.adobe.com/adobecloud/rel/create": {
+      "href": "https://api.example.com/v2/create{+path}{?intermediates}",
+      "templated": true
+    },
+    "http://ns.adobe.com/adobecloud/rel/resolve/id": {
+      "href": "https://api2.example.com/v7/resolve{?asset_id}",
+      "templated": true
+    }
+  },
+  "_page": {
```
tdahlgre: I also added another directory sample which includes _page

schemas/common/page.schema.json
```diff
@@ -0,0 +1,67 @@
...
+      "properties": {
+        "orderBy": {
+          "title": "Order by property",
+          "type": "string",
+          "description": "The property by which this page is ordered.\n"
+        },
+        "order": {
+          "title": "Sort order",
+          "type": "string",
+          "pattern": "^(asc|desc)$",
```
trieloff: Better: `"enum": [ "asc", "desc" ]` and 
`"meta:enum": {"asc": "Ascending order, i.e. the entries with the greatest value according to `orderBy` come first.", "desc": "Descending order"}`

schemas/common/page.schema.json
```diff
@@ -0,0 +1,67 @@
...
+          "description": "Sort order, either \"asc\" or \"desc\".\n"
+        },
+        "start": {
+          "title": "First value",
+          "type": "string",
+          "description": "The first value, in sorted order, of the orderby property on this page.\n"
```
trieloff: Make sure `orderBy` is in camelcase, even in descriptions. Use backticks to make it stand out. (the `\n` at the end is not needed)

trieloff: Question: is `orderBy` only allowed for `string` values? If no, the type of this property might be `anyOf` string, number, boolean (at least)

schemas/common/page.schema.json
```diff
@@ -0,0 +1,67 @@
...
+        },
+        "next": {
+          "title": "Next page start value",
+          "type": "string",
+          "description": "The start value for the next page.\n"
+        },
+        "property": {
+          "title": "Filter properties",
+          "type": "string",
+          "description": "The list of properties by which the result is filtered, if any.\n"
```
trieloff: If it's a list, what is the list separator? What is the escape sequence in case the list separator is used in a property name?

In cases like this, it's a good idea to also include an `examples` array right in the property.

schemas/common/page.schema.json
```diff
@@ -0,0 +1,67 @@
...
+        },
+        "type": {
+          "title": "Filter type value",
+          "type": "string",
+          "description": "The applied type filtering value, if any.\n"
```
trieloff: How is `type` related to `filter`? Is this just a special, inconsistent way to filter for `dc:type` values?

schemas/common/page.schema.json
fmeschbe: `orderBy` is a comma-separated list of property names, optionally prefixed with `+` or `-`. Hence it is a `string` always.

schemas/common/page.schema.json
fmeschbe: Comma-separated list if property names („tokens“). Good question what the character space for „token“ actually is. I would suspect comma is not a valid token character.

schemas/common/page.schema.json
fmeschbe: +1 for `meta:enum`, though this property is likely going away.

schemas/common/page.schema.json
fmeschbe: `type` filters on the media type of the list entry, while `property` filters on the existence of a specific property in a list entry.

schemas/common/page.schema.json
trieloff: `orderBy` is always a `string`, that much is clear. The question is if ordering is only allowed for properties that have string values (which does not seem to make sense). In that case, the `start` and `next` properties should be able to have other values, too.

trieloff: What does `+` and `-` mean? All of that needs to go into the description string.

schemas/common/page.schema.json
trieloff: If I understood you correct, then `property` simply checks for existence of the property. What counts as non-existence?

- `undefined`
- `null`
- `0`
- `false`
- `NaN`

schemas/common/page.schema.json
```diff
@@ -0,0 +1,67 @@
...
+  "description": "Page object contains paging information for related resource when resource result is paginated.",
+  "definitions": {
+    "page": {
+      "properties": {
+        "orderBy": {
+          "title": "Order by property",
+          "type": "string",
+          "description": "The property by which this page is ordered.\n"
+        },
+        "order": {
```
fmeschbe: FYI: this property has now been removed and we should probably also removed it here from the _page object. It serves not useful purpose any more. Thanks.

/cc @tdahlgre, @trieloff

trieloff:
@tdahlgre, this PR still has unaddressed issues and change requests from @fmeschbe. Is there anything I can do to help move it forward?

tdahlgre:
@trieloff, @fmeschbe - I have removed 'order' property and added description to orderBy per description in Platform API spec. Let me know if there is other items to be addressed. I also removed "WIP" from pull request title

trieloff:
Looks good from my end. I've updated the documentation and filled in the missing bits.

@fmeschbe care to remove the red x?

trieloff:
@fmeschbe unless I hear from you by tomorrow, I'll get this ready to be merged.

PR #101, opened 2018-01-23
Merge Experience Cloud Staging

trieloff:
These are the final schemas for #59


schemas/context/experienceevent.example.1.json
```diff
@@ -0,0 +1,91 @@
...
+    "@id": "https://data.adobe.io/datasources/datasource-123",
+    "xdm:name": "DataSourceIntegrationCode-123"
+  },
+  "xdm:timestamp": "2017-09-26T15:52:25+00:00",
+  "xdm:endUserIds": {
+    "xdm:realm": {
+      "@id": "https://data.adobe.io/experiencecloud/audiencemanager",
+      "xdm:name": "Adobe Audience Manager"
+    },
+    "https://ns.adobe.com/experiencecloud/mcid": {
```
kstreeter: In the schema for EndUserId, this is xdm:mcid, but here it is in the EC namespace.  Which was intended here?

trieloff: We can still change it, but my rationale was this:
- MCID, Target, Analytics, etc. are all data sources for XDM, but not part of XDM itself. Therefore, I've given them URIs outside the XDM namespace.

PR #103, opened 2018-01-25
Contributing licensing

cdegroot-adobe:
More clearly outlined the licensing requirements as we have partners asking how they may easily contribute. Please verify that these are the ways that Legal would like us to direct contributors to provide their content.


trieloff:
Nice, that will make it easier to contribute.

PR #106, opened 2018-02-12
Add schema for effective privileges, used in a recent addition to the…

mburbidgadobe:
… Platform API.

Please link to the issue #…


schemas/external/repo/effectiveprivileges.example1.json
```diff
@@ -0,0 +1,4 @@
+{
+    "*": ["read", "write"],
+    "api:ac": ["read"]
```
trieloff: `api` is an unknown namespace. Use `repo` instead.

trieloff: Or introduce the namespace (probably replacing `repo`)

schemas/external/repo/effectiveprivileges.schema.json
```diff
@@ -0,0 +1,33 @@
...
+  "$id": "https://ns.adobe.com/xdm/base/effectiveprivileges",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Effective Privileges",
+  "description": "Schema for the Effective Privileges resource",
+  "type": "object",
+  "definitions": {
+    "effectiveprivileges": {
+      "type": "object",
+      "properties": {
+        "*": { "$ref": "#/definitions/privilege" }
```
trieloff: I know `*` is a valid JSON name, but it still feels odd to me. @kstreeter @lrosenthol WDYT?

trieloff:
@mburbidgadobe please also take care of the failing build. `npm run lint` and `npm test` need to succeed.

schemas/external/repo/effectiveprivileges.schema.json
mburbidgadobe: "*" was suggested by Felix. It does accurately described the property. Alternative might be "default".

mburbidgadobe:
I think all tests are succeeding now.

schemas/external/repo/effectiveprivileges.schema.json
trieloff: Ok, then we'll roll with it.

Issue #107, opened 2018-02-13
Rollback metrics related schemas to platform's original definition.


jwen-adobe:
## What are the schemas that are affected by the issue
All metrics related schemas in data folder

## What are examples of products that are impacted by the issue
platform xed conversion

trieloff:
How does this relate to and address the issues raised in #84 ?

jwen-adobe:
@cluby @kstreete @cmathis
The original definition was proposed by analytics and reviewed internally at platform. Basically, it needs to satisfy the requirements below. With all the latest changes, the impact needs to be re-evaluated from analytics point of view. Besides, 1). the structure change creates schema for each metric, so if analytics has eVar1 to eVar1000, are we going to create 1000 schemas? 2). We want core metrics not using URI but shorter fields grouped by different categories.

Analytics requirements:
=================
-Explicit fields for each metric that the user can directly populate or query.  No arrays.
-Structure that lets us group related metrics by “vertical” like page-views and link-clicks under web, sends under direct-marketing, etc
-For the above structure I would like it to be similar to the way we structure the additional data/dimensions in ExperienceEvent so we are consistent and intuitive
-The ability to set a float/double value for the metric
-An optional unique identifier for the metric/event to support deduplication of the metric/event downstream while still keeping the rest of the ExperienceEvent
-Ability to have core, Adobe solution specific, partners, and customer extensions using the same Metric data type for any metric type field

PR #108, opened 2018-02-13
Rollback metrics related schemas for platform xed conversion

jwen-adobe:
This PR links to the issue #107 

@trieloff @kstreete @cmathis

Need to rollback the metrics related schemas back to original definition from platform as none of those changes are reviewed by platform PMs/Analytics folks and should be merged as is. The old definition also works well with xed conversion tooling.


kstreeter:
@jwen-adobe we shouldn't just try to roll this back.  If there are elements that are breaking our tooling, let's propose fixes.  If there are concerns or differing opinions about the semantics of these definitions, we should file an issue and make a counter proposal.

jwen-adobe:
@kstreeter @cmathis @cluby @cdegroot-adobe @cfraser
I agree the rollback approach is too simple for the reconciliation and tooling should not drive that. We should bring Analytics/PMs together to finalize the metrics definitions and format Lars proposed. The tooling can be adapted once the metrics definition/format are set.

cmathis:
I think the only change that was needed in this case was to add back the defined categorized metric fields back but keep around the patternProperties approach as well. The data platform would treat this as a Metrics object/schema that has some dedicated fields for some of the metric types and then the patternProperties becomes a field of type "map" that can be used to pass in and store these other meta:keys options.

Issue #113, opened 2018-02-14
Update schema enduserids


jwen-adobe:
## What are the schemas that are affected by the issue
enduserids
identity
namespace

## What are examples of products that are impacted by the issue
platform xdm tooling

trieloff:
Can you clarify how the schema should be updated?

jwen-adobe:
See PR here: https://github.com/adobe/xdm/pull/114

1. introduce two new schemas identity and namespace
2. remove the patternProperties and call out the fields explicitly like below:

"properties": {
        "https://ns.adobe.com/experience/mcId": {
          "title": "Adobe Marketing Cloud Id",
          "$ref": "https://ns.adobe.com/xdm/context/identity",
          "description": "A unique identifier from Adobe Marketing Cloud."
        },
        "https://ns.adobe.com/experience/analytics": {
          "title": "Adobe Analytics Cloud End User Ids",
          "$ref": "https://ns.adobe.com/xdm/context/identity",
          "description": "Adobe Analytics Cloud extension to End User Ids."
        }
      }

jwen-adobe:
This is fixed.

PR #114, opened 2018-02-14
Update enduserids

jwen-adobe:
This PR links to the issue #113 

@trieloff @kstreete @cmathis

Update enduserids to use uri as field names. 

@trieloff Do we allow the schemas under schema folder reference schemas under extensions folder? I got validation errors when doing that. It seems it works only the other way around which is by design?


cmathis:
Looks good to me. 

patternProperties was not part of the original design of this schema. Not sure if/how that would be used in the data platform. Would suggest we remove it until a use-case needs it, but if we need to leave it around for some unknown use-case then so be it. 

@trieloff - What more needs to be done to merge this in.

schemas/context/enduserids.example.1.json
```diff
@@ -1,18 +1,16 @@
 {
-  "xdm:realm": {
-    "@id": "https://data.adobe.io/experiencecloud/audiencemanager",
-    "xdm:name": "Adobe Audience Manager"
+  "https://ns.adobe.com/experience/mcId": {
```
trieloff: We should use all-lowercase URIs for consistency.

schemas/context/enduserids.schema.json
```diff
@@ -9,42 +9,30 @@
   "$schema": "http://json-schema.org/draft-06/schema#",
   "title": "End User IDs",
   "type": "object",
-  "description": "Condensed, normalized encapsulation of all end user identifiers. NOTE: At least one of the fields is required.",
-  "minProperties": 1,
```
trieloff: If at least one of the properties is required, why do we remove the formalized requirement?

schemas/context/enduserids.schema.json
```diff
@@ -9,42 +9,30 @@
...
+  "description": "Condensed, normalized encapsulation of all end user identifiers. NOTE: At least one of the fields is required.\n",
   "properties": {
-    "xdm:realm": {
```
trieloff: Why do we remove `realm`? Is it being replaced with something else?

schemas/context/enduserids.schema.json
```diff
@@ -9,42 +9,30 @@
...
-      "title": "Realm",
-      "$ref": "https://ns.adobe.com/xdm/data/datasource",
-      "description": "The optional realm associated with the identity stitching strategy."
+    "https://ns.adobe.com/experience/mcId": {
```
trieloff: Should be a lowercase URI

schemas/context/enduserids.schema.json
```diff
@@ -9,42 +9,30 @@
...
+      "title": "Adobe Marketing Cloud Id",
```
trieloff: ID should be capitalized.

schemas/context/enduserids.schema.json
```diff
@@ -9,42 +9,30 @@
...
+      "$ref": "https://ns.adobe.com/xdm/context/identity",
+      "description": "A unique identifier from Adobe Marketing Cloud."
```
trieloff: Clarify what "from" means: Generated by? Understood by? All of the above?

schemas/context/enduserids.schema.json
```diff
@@ -9,42 +9,30 @@
...
+      "$ref": "https://ns.adobe.com/xdm/context/identity",
+      "description": "Adobe Analytics Cloud extension to End User Ids."
     }
   },
   "meta:keys": {
-    "https://ns.adobe.com/experiencecloud/target": "Adobe Target",
-    "https://ns.adobe.com/experiencecloud/campaign": "Adobe Campaign",
-    "https://ns.adobe.com/experiencecloud/analytics": "Adobe Analytics",
-    "https://ns.adobe.com/experiencecloud/mcid": "Marketing Cloud Identity Core Service"
+    "https://ns.adobe.com/experience/target": "Adobe Target",
```
trieloff: What's the rationale for using `experience` over the full product name?

schemas/context/enduserids.schema.json
```diff
@@ -9,42 +9,30 @@
...
+    "https://ns.adobe.com/experience/campaign": "Adobe Campaign",
+    "https://ns.adobe.com/experience/analytics": "Adobe Analytics",
+    "https://ns.adobe.com/experience/mcid": "Marketing Cloud Identity Core Service"
```
trieloff: Inconsistency with the `mcId` above.

schemas/context/enduserids.schema.json
```diff
@@ -9,42 +9,30 @@
...
-        "xdm:confidence": {
-          "type": "number",
-          "minimum": 0,
-          "maximum": 1
-        }
-      },
-      "required": [
-        "xdm:id"
-      ]
+      "$ref": "https://ns.adobe.com/xdm/context/identity"
```
trieloff: I think it's a good idea to refactor this into it's own schema.

schemas/context/enduserids.schema.json
```diff
@@ -9,42 +9,30 @@
...
       "type": "object",
-      "properties": {
-        "xdm:datasource": {
-          "$ref": "https://ns.adobe.com/xdm/data/datasource"
-        },
-        "xdm:id": {
-          "type": "string",
-          "description": "The ID that is being used by the service specified by the outer property name or `xdm:datasource`\n\n. **Note:** opposed to most other IDs in XDM, this is not a URI, as it cannot be guaranteed that all identity services are using URIs as IDs."
-        },
-        "xdm:confidence": {
```
trieloff: It looks like this did not make it into the refactored schema.

schemas/context/identity.schema.json
```diff
@@ -0,0 +1,38 @@
...
+    "Copyright 2017 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/context/identity",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Identity",
+  "type": "object",
+  "description": "Identity from identity services which is the fact of being who or what a person or thing is.\n",
```
trieloff: This description does not make any sense. The `identity` here is neither describing who or what a person is (does not cover things at all, or only tangentially), but a unique value that has been associated with a person, device, or user agent by an identity service.

If we use the name "identity service" we need to explain how it is related to the namespace in this schema.

schemas/context/identity.schema.json
```diff
@@ -0,0 +1,38 @@
...
+  "definitions": {
+    "identity": {
+      "properties": {
+        "@id": {
```
trieloff: You know, I'm a stickler for `uri`s in all places. But here I wonder if it is really practical, given that we cannot control what an external service is using for IDs. In some cases, it might be numbers, in some UUIDs, in some random strings. I think `string` is the safest bet, which means we should also rename the property to just `xdm:id`.

schemas/context/identity.schema.json
```diff
@@ -0,0 +1,38 @@
...
+          "title": "Identifier",
+          "type": "string",
+          "format": "uri",
+          "description": "Identity of the consumer in the related namespace."
```
trieloff: Make it clear that the identifier has been given by the "identity service" (which is itself identified by the "namespace" – if it get that right.)

schemas/context/identity.schema.json
```diff
@@ -0,0 +1,38 @@
...
+        },
+        "xdm:namespace": {
+          "title": "Namespace",
+          "$ref": "https://ns.adobe.com/xdm/context/namespace",
+          "description": "The namespace associated with the id attribute and matched up with the AAM data source integration code."
```
trieloff: ID should be capitalized or backticked. What's this reference to AAM? (continued below)

schemas/context/identity.schema.json
```diff
@@ -0,0 +1,38 @@
...
+        },
+        "xdm:xid": {
+          "title": "Adobe Experience Identifier",
```
trieloff: Should the name include Adobe? i.e. is "Adobe" the authority that governs this identifier, or is it XDM. The description below makes it look like it's only XDM.

schemas/context/identity.schema.json
```diff
@@ -0,0 +1,38 @@
...
+          "type": "string",
+          "description": "The optional, universally unique Adobe identity of the consumer. Generated using the Identity and the related namespace id."
```
trieloff: Generated how? I assume through concatenation with a delimiter. What's the delimiter? This can benefit from an embedded example.

schemas/context/namespace.example.1.json
```diff
@@ -0,0 +1,4 @@
+{
+  "xdm:id": 12345,
```
trieloff: Why is this just a number and not a URI?

schemas/context/namespace.schema.json
```diff
@@ -0,0 +1,37 @@
...
+      "Copyright 2017 Adobe Systems Incorporated. All rights reserved.",
+      "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+      "you may not use this file except in compliance with the License. You may obtain a copy",
+      "of the License at https://creativecommons.org/licenses/by/4.0/"
+   ],
+   "$id":"https://ns.adobe.com/xdm/context/namespace",
+   "$schema":"http://json-schema.org/draft-06/schema#",
+   "title":"Namespace",
+   "type":"object",
+   "description":"The Datasource acts as a namespace or unique identifier associated with a collection of data. Each EndUserID is associated with a given Datasource. A user can create and obtain information about a Datasource during the solution onboarding.\n",
```
trieloff: Inconsistency between "Namespace" in the title and "Datasource" in the description.

schemas/context/namespace.schema.json
```diff
@@ -0,0 +1,37 @@
...
+   "definitions":{
+      "namespace":{
+         "properties":{
+            "xdm:id":{
+               "title":"Identifier",
+               "type":"integer",
```
trieloff: Numeric IDs are a bad idea. Give the poor Namespace a proper URI, which might end with an integer for now.

schemas/context/namespace.schema.json
```diff
@@ -0,0 +1,37 @@
...
+               "description":"The numeric ID associated with this Datasource. This would be provided by the individual or system that created the Datasource.\n"
+            },
+            "xdm:code":{
```
trieloff: I'm completely at a loss why we need both and what the difference is. Above it seems to make some reference to Adobe Audience Manager, which seems to be an implementation detail.

schemas/context/enduserids.schema.json
cmathis: Not sure why that was remove. @jwen-adobe, is there a reason for this? 

Yes, at least on property is required.

schemas/context/enduserids.example.1.json
cmathis: Sounds good. As I mentioned in another comment these URI's within the data platform have to be converted into a nested structure where the "mcId" becomes a property of the parent "experience" object (ns.adobe.com.experience.mcId). We can just add logic to our translation of the schemas to convert that final field name into camelCase.

schemas/context/namespace.schema.json
cmathis: Good point. This is a left over description of when this schema was created from the Datasource schema. Both the Namespace and Datasource schema are identical except they have different use cases. One is used to represent the Namespace of a given Identity (aka user identifier). The other is used to represent the Datasource of a source of data.

schemas/context/namespace.schema.json
cmathis: That's fine, but should go to the system that generates and provides those id's and have them produce the full URI?

schemas/context/namespace.schema.json
trieloff: I think then it makes sense to unify this notion of a "governor of data", as we encounter it in various places:
- where data points are coming from
- generator of IDs
- resolver of IDs
- decider of segment membership
- etc.

I like the term "data source" better than "namespace" because it is more specific and less ambiguous, but it might not adequately capture all meanings. "Realm" has a poetic quality to it, but is even less clear than namespace.

schemas/context/namespace.schema.json
trieloff: For the data platform, I'd assume the system that manages namespaces and the system that generates IDs/URIs is identical. In ingestion, you could even go so far to ignore data that belongs to an unknown namespace, i.e. that has a namespace URI that is outside of your known URI schema.

schemas/context/namespace.schema.json
trieloff: Paging @cdegroot-adobe ;)

trieloff: I think we should be clear which properties are core and which are implementation detail (and can be delegated to an extension). Giving the datasource an identifier (ideally an uniform identifier of the resource) makes a lot of sense. Forcing in an additional identifier that is only relevant to a soon-to-be legacy system will only lead to confusion.

You can count my confusion as a small taste of what's to come when we expose this to users with less understanding of our systems than I have.

schemas/context/enduserids.schema.json
cdegroot-adobe: Realm is currently under review internally, the team has asked for it to be removed now and when they have completed the architecture they will make an amended proposal.

schemas/context/enduserids.schema.json
cdegroot-adobe: Why?

schemas/context/identity.schema.json
cdegroot-adobe: How it is created is an implementation detail. What we will need to do here is add a way of distinguishing what system is providing this representation.

schemas/context/enduserids.schema.json
trieloff: From https://github.com/adobe/xdm/blob/master/CONTRIBUTING.md#coding-styleguides
> Acronyms in camelCase like ID, API, JSON are also capitalized in camelCase, such as assetID

In this case: *ID* stands for Identifier, not for Sigmund Freud's concept of the "Id".

schemas/context/namespace.example.1.json
cdegroot-adobe: Largely for efficiency reasons. User Identities will be processed in the trillions per day and require super low latency.

schemas/context/namespace.schema.json
cdegroot-adobe: There is a recurring pattern used here to make the id spaces unique in systems that are loosely coupled with super high efficiency. While the pattern is the same, the id spaces being made unique are very different. End Users, Segments, Data producers, Merging/stitching functions. The concept of "governor of data" is interesting. Unifying the naming has proven elusive.

schemas/context/enduserids.schema.json
cfraser: Not to be pedantic, but Id is an abbreviation and not an acronym so it should not be cased as such. It should be `Id` not `ID`. This is what was agreed upon in the XDM style guide. 

When did we decide to deviate for what was agreed upon previously?

schemas/context/identity.schema.json
trieloff: The description strongly implies that there are semantics to the internal structure of the `xid`, but never specifies what these semantics or the structure are. Right now, all of these would be valid values:
* ``
* `0`
* `Batman`

schemas/context/namespace.schema.json
trieloff: That's why I strongly distinguish between stuff that's covered by XDM, which is using URIs as `@id` and stuff that's ruled by implementation details, which can be any string in any `xdm:id` attribute.

The goal of the "Namespace/Data Source/Realm/Governor" is to identify which system is providing the specific implementation. Keeping track of a list of systems where data is originating is the least we can do.

schemas/context/namespace.schema.json
cdegroot-adobe: The efficiency and scale of this means that integer processing vs processing a URI is important.

schemas/context/namespace.schema.json
cdegroot-adobe: The systems we are using here are certainly not "soon-to-be legacy", they are being re-organized to provide a single consistent service for customers/partners/adobe etc. Nothing is going away.
What we are doing here does need to be expanded to cater for other "governors or data" services beyond Adobe, and that would certainly be a URI. But the values that need to be processed for a namespace and identity itself must be defined an a way that also meets performance goals for implementations. The way Identity works is not for the faint of heart.

Another aspect of this duality that we need to outline more is the ingest/egress vs enriched processing. Systems generating ExperienceEvents or Profiles will very often not be able to easily generate the XIDs, so will provide only the Id and namespace. Systems that are XID aware  will use and generate XIDs and be able to use them for super low latency processing between each other, especially when identity graphs are in play. XDM is also for this XID-aware system to system interoperability, not just ingress/egress. If the ingress/egress system does populate the XIDs then they can expect that those XIDs will be retained in responses.

schemas/context/enduserids.schema.json
cdegroot-adobe: I find the use of fully capitalized "ID" to be outside the industry norm. What was the reasoning for the change. Note this is not an Acronym, so is not covered by that change.

schemas/context/namespace.schema.json
trieloff: Hashing a string into an integer is a really fast operation (especially considering that you are in the context of JSON processing). 

But keeping a list of integers centrally managed and unique over decades is almost impossible.

schemas/context/namespace.schema.json
trieloff: In my mind, both `xdm:id` and `xdm:code` can be combined into a simple `@id` that follows a pattern like `https://data.adobe.io/{namespace|datasource|…}/<code>/<numeric_id>`. This gives you the ease of processing it in systems that can read only the code or the numeric ID today, but won't lock you in in the future when you decide to use only codes, only numbers, or a completely different identification system.

cdegroot-adobe: As a representation yes. How much that is helping human readability is open in my mind. Considering the primary processing systems in our industry are SQL, R, Scala etc this is not that effective. It would be problematic to natively use it, thus complex processing is needed Unicode parsing, secure URL parsing etc. Scale this to trillions of parsings a day with sub ms SLAs and it is not realistic.

schemas/context/namespace.schema.json
cdegroot-adobe: "(especially considering that you are in the context of JSON processing)" The schema is represented in JSON, the data is sometimes in some cases, but mostly the industry and customer implementations keep the data is a binary format, db logs, parquet and process it from that format without ever going to JSON.

If a different "governor of data" wants to use a different option to generate the integers they are free to do so. That is an implementation detail.

What we are proposing here is implemented at scale and in production for many years.

schemas/context/namespace.schema.json
kstreeter: This actually isn't true.  The identifier is not a composite, like <id + code> or <code + id>.  Id is the identifier, and managed by the system.  Code is an alias, and managed by the application (in the context of a tenant).  There is an expected data flow where a user includes only the code, and the system (knowing the tenant context) looks up the ID.  The ID (which is global across tenants) is then used in the rest of the flow.

schemas/context/namespace.schema.json
kstreeter: We should define the IDs as URIs.  Any ID (numerical or string) can be made a URI.  We should not put any restrictions on the form of the URI though, so that implementors have maximum flexibility.

schemas/context/enduserids.schema.json
kstreeter: Use of "ID" is called out specifically in contributing, and used on the other models (like asset).  I suggest for the purposes of this PR we follow the guidelines.  If we want to change them we should open an issue and have a separate discussion.

schemas/context/namespace.schema.json
trieloff: @cdegroot-adobe the representation that XDM specifies and that we discuss is JSON. Any system processing it, should represent data in a way that is convenient and efficient, by any means. 

In this case here, where the URIs is not restricted in any way and will not be interpreted by your system, hashing it can be an efficient way to turn it into a fixed-length number.

schemas/context/namespace.schema.json
trieloff: So, do we need both? It sounds like `id` is globally unique, but `code` only works when you know the tenant, which we haven't even modeled.

trieloff: Add to that that `id` is required, but `code` is optional. For the flow that @kstreeter describes it sounds like `id` is optional, but `code` is required.

kstreeter: You need one, or the other, or both.  ID is not necessarily globally unique: it is scoped to the namespace.

The code is actually scoped to the namespace, too.  My description above was a little incomplete there: the code is scoped to a namespace, which may internally be owned by a tenant but that is probably an implementation detail at this level.

schemas/context/enduserids.schema.json
trieloff: @cfraser I love me a good pedantic remark: https://github.com/adobe/xdm/pull/131

As @kstreeter pointed out, we use `ID` elsewhere, so let's keep it consistent.

cfraser: @trieloff Ha! Touché, however, since we've agreed on pascal case `ID` doesn't conform, right?

I don't want to make this a semi-colon, no semi-colons debate but...

schemas/context/enduserids.schema.json
jwen-adobe: @cmathis @trieloff The original requirement was at least one field is required, but we can not call out which field..Other ways we can do for this case?

schemas/context/enduserids.example.1.json
jwen-adobe: @cmathis @trieloff  how to convert to camelCase if the text are all lowercase without '-' used??

schemas/context/namespace.schema.json
cmathis: I think we all agree with this, and is what we were trying to do. The Identity (which contains the id and namespace of that ID) is always under a specific namespaced field/implementation - "https://ns.adobe.com/experience/analytics", "https://ns.adobe.com/experience/mcID", etc.

schemas/context/namespace.schema.json
cmathis: Reading through this, I'm not 100% sure what the resolution is. @trieloff  - What open items do you still see?

schemas/context/enduserids.schema.json
cmathis: @jwen-adobe - Isn't that what the "minProperties" is saying. You need to have at least one of the fields populated?

schemas/context/enduserids.example.1.json
cmathis: @jwen-adobe - Good point Joe. Didn't think that one through. If the final value in the URI is lowercase then we have no way to get back to camelCase. A decision is needed to know how critical it is if all final storage field names need to be camelCase.

schemas/context/enduserids.schema.json
jwen-adobe: @cmathis Yes we can use the minProperties here. Was thinking of oneOf and do not want to use it.

schemas/context/identity.schema.json
kstreeter: Agreed, we should make this more clear

schemas/context/identity.schema.json
kstreeter: I can agree with that.  @cdegroot-adobe had a similar comment.

schemas/context/identity.schema.json
kstreeter: My feeling is that this should be defined simply as a global identifier that corresponds to the namespace+id.  This reflects the nature of the system in the correct way:  you have one service that maintains a match between service-provided IDs and user-provided IDs ("codes"), in the context of a namespace.  You have an (optional) second service that collects IDs from the first service and folds them into a global IDs space.

schemas/context/identity.schema.json
kstreeter: we should define it as a URI: an opaque but global identifier (see my comment above)

schemas/context/enduserids.schema.json
cdegroot-adobe: From as "generated by".

schemas/context/identity.schema.json
kstreeter: @jwen-adobe, we can remove the reference to Adobe here.  It's not specific to us, as anyone could build a system that assigns IDs globally across namespaces.  So something like:
 "when present, this value represents a cross-namespace identifier that is unique across all namespace-scoped identifiers in all namespaces."

jwen-adobe: @kstreeter Changes made.

schemas/context/namespace.schema.json
trieloff: Action item for @jwen-adobe: 
- [ ] unify this as "Namespace" for now. Most of the ID-stuff discussed here belongs in another thread in the same PR.

schemas/context/namespace.schema.json
trieloff: Action item for @jwen-adobe:
- [ ] rename to `@id`
- [ ] make it a `uri`

schemas/context/namespace.schema.json
trieloff: Resolution:
- [ ] update description: explain that the `code` is a shortcut to the full `@id` and that either `code` or `@id` can be used
- [ ] update schema with an `oneOf/required` construct, so that validation enforces presence of either `code` or `@id`
- [ ] provide examples and invalid examples for each

/cc @cmathis @jwen-adobe @kstreeter

trieloff:
First:
- [ ] tick the checkboxes above

Final cleanup:
- [ ] merge master into this branch
- [ ] resolve the conflicts
- [ ] run `npm run lint`
- [ ] commit the prettified results
- [ ] fix the tests

Then we should be ready to merge.

schemas/context/namespace.schema.json
jwen-adobe: @trieloff @cmathis done.

schemas/context/namespace.schema.json
jwen-adobe: @trieloff @cmathis done.

schemas/context/namespace.schema.json
jwen-adobe: @trieloff @cmathis @kstreeter Done. I did not use oneOf here as comment from Kevin stated both can be used as well. I replaced "required" with "minProperties" though.

schemas/context/identity.schema.json
trieloff: Resolved

PR #116, opened 2018-02-15
Change xdm:timestamp to integer

jwen-adobe:
This PR links to the issue #117 

@trieloff @kstreete @cmathis

Change timestamp field back to integer as this has been discussed and agreed from platform.

### DIFF
--- a/schemas/context/experienceevent.schema.json
+++ b/schemas/context/experienceevent.schema.json
@@ -26,9 +26,10 @@
         },
         "xdm:timestamp": {
           "title": "Timestamp",
-          "type": "string",
-          "format": "date-time",
-          "description": "The time when the first event of the interaction occurred."
+          "type": "integer",
+          "description": "The timestamp when the first event of the touchpoint occurred. Milliseconds since midnight of January 1, 1970.",
+          "minimum": 1,
+          "maximum": 9223372036854770000
         },
         "xdm:endUserIds": {
           "title": "End User IDs",
--- a/schemas/context/experienceevent.example.1.json
+++ b/schemas/context/experienceevent.example.1.json
@@ -4,7 +4,7 @@
     "@id": "https://data.adobe.io/datasources/datasource-123",
     "xdm:name": "DataSourceIntegrationCode-123"
   },
-  "xdm:timestamp": "2017-09-26T15:52:25+00:00",
+  "xdm:timestamp": 978336000000,
   "xdm:endUserIds": {
     "xdm:realm": {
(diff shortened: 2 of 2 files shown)


cmathis:
Looks good. Yes, this should be an int64 timestamp.

schemas/context/experienceevent.schema.json
```diff
@@ -26,9 +26,10 @@
         },
         "xdm:timestamp": {
           "title": "Timestamp",
-          "type": "string",
-          "format": "date-time",
-          "description": "The time when the first event of the interaction occurred."
+          "type": "integer",
```
trieloff: As pointed out in #117, I think this change is bad idea in multiple ways:

1. it introduces an inconsistency within XDM, because we now have some places where time is indicated through `date-time` and some where it is an `integer`
2. it makes XDM inconsistent with the rest of the JSON ecosystem that is using `string`-encoded dates all over the place.

The only reason I can see why we would do that is because it is more convenient for one single implementation, but that does not justify the price of this inconsistency for everyone else.

schemas/context/experienceevent.schema.json
```diff
@@ -26,9 +26,10 @@
...
+          "description": "The timestamp when the first event of the touchpoint occurred. Milliseconds since midnight of January 1, 1970.",
```
trieloff: The description refers to "touchpoint" which is not used elsewhere in this schema. 

What's the time zone of reference for January 1st, 1970?

cmathis: Should be "Experience Event". 

The timestamp is nothing more then a unix timestamp in the form of millisecond which has a clear understanding amongst those that use it.  We can likely convert this to a string with the date-time format if that means the same thing.

trieloff: Yes, they are semantically equivalent, with a couple of advantages:
- `date-time` is built in in JSON Schema and used throughout XDM
- UNIX time has ambiguities around leap seconds

trieloff:
From @cmathis 
> We can likely convert this to a string with the date-time format if that means the same thing.

Let's do this then. I'll close the PR for now.

schemas/context/experienceevent.schema.json
jwen-adobe: @trieloff will the date-time support millisecond by using 
"time-hour ":" time-minute ":" time-second [time-secfrac]" mentioned here? https://tools.ietf.org/html/rfc3339

jwen-adobe:
@cmathis @kstreeter @cdegroot-adobe @@cfraser. FYI, this is not merged.

Issue #122, opened 2018-02-16
Add properties field in organization schema's identifier and classifier


jwen-adobe:
## What are the schemas that are affected by the issue
organization

## What are examples of products that are impacted by the issue
platform xdm definition

jwen-adobe:
Close for now without fix.

Issue #66, opened 2018-01-12
Questions on Organizations (Experience Cloud)


trieloff:
Right now, we have following external IDs in this schema:

- `duns`: Dun & Bradstreet
- `sicCode`: [SIC Code](https://en.wikipedia.org/wiki/Standard_Industrial_Classification)

Given that neither is the only provider of organizational identification or classification (in particular SIC, which is being replaced with NAICS since **1997**), we could model it in a more extensible way:

```json
  "xdm:identifiers": [
    {
      "@type": "http://dnb.com",
      "xdm:id": "1262"
    }
  ],
  "xdm:classifiers": [
    {
      "@type": "http://www.companieshouse.gov.uk/infoAndGuide/sic/sic2007.shtml",
      "xdm:class": "1241"
    }
  ]
```

That way, new classifiers and identifiers can be added easily, while we only have to deal with a couple of known `@type` values.

trieloff:
@kstreeter @jwen-adobe

kstreeter:
I'm not sure I see the advantage of making these arrays.  We have actually had this conversation in several contexts.  In general, no processor is going to iterate over an array.  They will look for the specific field they care about, if it is present.  This is better represented as a specific field, just as it is now.  You extend it by adding new fields.

Note that we should not use fields with an @ in models that are intended to be stored in the data platform.  It will complicate our "data safe" encoding.

trieloff:
@kstreeter good point on the "accessibility" of arrays. It's indeed a bit cumbersome. Thinking about this a bit more, it might be a good scenario for our extensions. Because the D&N reference is essentially a vendor extension, I think it should live on the same level as references to other company databases like [Orbis](https://www.bvdinfo.com/en-us/our-products/company-information/international-products/orbis), [Datafox](http://www.datafox.com), [Angellist](https://angel.co), or [Crunchbase](https://www.crunchbase.com) (these are just the ones I have used in the past).

It would look like this:

```json
"xdm:identifiers": {
  "http://dnb.com": "1262",
  "https://angel.co/": "https://angel.co/omniture"
}
```

Accessing the D&B number would be as simple as `data.xdm:identifiers.['http://dnb.com']`, i.e. not much more difficult than `data.xdm:dnbCode`.

kstreeter:
Yes I think this is the right direction.  It's just a property, and qualified like other properties are: with a prefix or as a full IRI.

[...]

PR #123, opened 2018-02-16
Add properties for dun and sicCode in organization

jwen-adobe:
This PR links to the issue #122 

@trieloff @kstreete @cmathis

This PR adds properties back for dun and sicCode in organization inside identifier and classifier.

### DIFF
--- a/schemas/common/organization.schema.json
+++ b/schemas/common/organization.schema.json
@@ -66,7 +66,7 @@
           "title": "Organization Identifier",
           "type": "object",
           "properties": {
-            "https://ns.adobe.com/external/dnb": {
+            "xdm:duns": {
                 "title": "Dun & Bradstreet number",
                 "type": "string",
                 "description": "The Dun & Bradstreet DUNS number for identifying an organization or business person."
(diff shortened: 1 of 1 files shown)


cmathis:
@trieloff  and @kstreete  - This seems like another area we are using patternProperties without a plan for how we would use that in the data platform. Right now we would just ignore it rather then define some generic field of type map. If we did convert it to a field of type map, I'm not sure what the field name would be. If we need it there for another use-case we will just plan to ignore it for now.

trieloff:
As discussed in #66, there are many, many ways to identify and classify organizations. Introducing a property will give a hard preferential treatment to only two vendors/standards (one vendor that we have not yet engaged enough with to have them claim their own namespace URL, and one standard that has been superseded by another standard).

This is a place where extensibility will become an issue fast. If you don't plan to support it in your implementation, I'm fine with that, but please don't make the standard fit your implementation. With `meta:keys` you do have a hint of what properties can be expected.

jwen-adobe:
@cmathis @kstreeter @cdegroot-adobe @@cfraser. FYI, this is not merged.

Issue #124, opened 2018-02-16
Add back more properties fields for optinout


jwen-adobe:
## What are the schemas that are affected by the issue
optinout

## What are examples of products that are impacted by the issue
platform xdm definition

trieloff:
Made obsolete by #139

PR #125, opened 2018-02-16
Add more properties to optinout

jwen-adobe:
This PR links to the issue #124 

@trieloff @kstreeter @cmathis

Add more properties fields back to optinout.


kstreeter:
I see two ways to resolve this, depending on how "JSON-LDish" we want the solution to be.

If we want to be "JSON-LD heavy", we should define both a namespace and a prefix for each channel type.  We can then use this prefix for defining all extension points that are channel based, including metrics and optins.  (As noted in discussion on metrics, having full URI everywhere in our standard schema is not really a workable option).

The downside of this is that we have a bunch of new namespaces, which we had previously decided we want to avoid.  If we want to avoid new namespaces, we can have a "JSON-LD light" solution and just put everything under the XDM namespace and prefix.  This would mean making sure that names did not conflict, but since this is all within the standard/core, that should not be a problem.  3rd-party (customer, vendor) extensions could use a full URI in the namespace of the 3rd-party.

kstreeter:
@trieloff can you please advise on this?  My preference is that we define both a namespace, and a prefix, for all channels.  We will then use the prefix to define both metrics and optins.  Are there objections to this?

trieloff:
> My preference is that we define both a namespace, and a prefix, for all channels. We will then use the prefix to define both metrics and optins.

Let's go with this approach. I'll file a new issue and close this PR.

kstreeter:
Hey @trieloff , @jwen-adobe I am re-opening this PR so we can at least make the change to move the keys from "meta:keys" into "properties", as we have done with metrics.  Let's keep https://github.com/adobe/xdm/issues/139 open, as I still feel it would be good to have short names for the channels.  But in the short term just getting the full URIs into "properties" will unblock our tooling.

Issue #126, opened 2018-02-16
Add uniqueId to measure and replace properties in metrics with shortened name


jwen-adobe:
## What are the schemas that are affected by the issue
-measure
-metrics

## What are examples of products that are impacted by the issue
platform xdm definition

jwen-adobe:
This is fixed.

PR #127, opened 2018-02-16
Add uniqueId to measure and replace metrics fields with shorter name

jwen-adobe:
This PR links to the issue #126 

@trieloff @kstreete @cmathis

-add uniqueId to measure schema (Analytics use case needs this)


cmathis:
@trieloff - The reason for this is because we have Adobe solutions using this metrics schema today and have already built code to populate existing non-nested fields (URI fields here become a long nested structure within the data platform).

@jwen-adobe  - The XDM UI today is not able to handle schemas that inline other objects. Will need to have these top level fields (xdm:directMarketing, xdm:commerce, etc.) then $ref another schema file. I think you had it this way in the PR earlier this week.

jwen-adobe:
@cmathis added new schema files for directmarketing/commerce/adverstising/web.

schemas/data/measure.schema.json
```diff
@@ -19,6 +19,11 @@
           "format": "uri",
           "description": "Links to the `Metric` that this measure refers to. The `@type` can be omitted in contexts where it is obvious what metric is being referred to, for instance in the `Metrics` object that is keyed with `Metric` URIs."
         },
+        "xdm:uniqueId": {
```
trieloff: I think this field is duplicating the [`@type`](https://github.com/adobe/xdm/blob/fac76175274eac81b0e47ee2c39f65b680710784/schemas/data/measure.schema.json#L17), because they both point to the Metric that this measure belongs to.

`uniqueId` has following disadvantages, though:
- it's not spelled `uniqueID` as the `CONTRIBUTING.md` guidelines want
- it's just a string, not a URI, which means it cannot be resolved
- it's unclear who governs the uniqueness of these values

schemas/data/metrics.schema.json
```diff
@@ -16,65 +16,29 @@
     }
   },
   "properties": {
-    "https://ns.adobe.com/xdm/data/metrics/commerce/abandons": {
```
trieloff: I discussed with (and agree with) @kstreeter that the full URIs here are inconvenient and that we would rather prefer short names. @cmathis: the "storage" argument is not so powerful as the "query language/consumability" argument, btw.

But: I think there are easier ways of doing this than the proposed approach. Let's consider following points:

1. yes, metrics can be grouped, but naming patterns like URIs or prefixes are better than nested objects, because ultimately, the metrics object is just a list of measures.
2. We could just use namespaces instead of full URIs for the known properties: `https://ns.adobe.com/xdm/data/metrics/commerce/abandons` becomes `commerce:abandons` (disadvantage: we have to define more namespaces)
3. We could just use one namespace, and patterns there, such as: `metrics:commerce-abandons` (disadvantage: nesting is less clear)
4. In any case we have to decide if the short name is a replacement or a synonym for the full URI, e.g. if both would be valid. Both approaches come with pros and cons.
5. In any case, `metrics` will be an area where extensibility of the platform will be tested. People will invent new metrics that are not part of XDM Core quickly.

cmathis: We can likely make a smaller namespace such as "metrics" work, but it will change the structure of a schema that has been around for six months and has solutions using it today.  Not sure you are aware of the issue with URI's, but the use of colons and slashes within a field name are not possible within the data storage tech. As a result these longer URI's have to be converted into a nested structure which REALLY changes what we had before. The closer we are to the original schema we are using today the better. The data platform has never had to deal with these long json-ld style schemas which need special translations in and out of the underlying storage tech. As of today, these schemas need to support both the ingestion, storage, and consumption format of the data.   

The "uniqueID" associated with each metric is to support a use-case where the user wants to  associate a value with a given measurement that can be used to identify duplicates later. That value could be a transaction ID, or any value that can be used during query time to remove duplicate records which is a common issue. Without this uniqueID, the query engines have no easy way to filter these duplicates out. We no doubt need a better description on this field.

trieloff:
> The "uniqueID" associated with each metric is to support a use-case where the user wants to associate a value with a given measurement that can be used to identify duplicates later. That value could be a transaction ID, or any value that can be used during query time to remove duplicate records which is a common issue. Without this uniqueID, the query engines have no easy way to filter these duplicates out. We no doubt need a better description on this field.

Let's take a stab at this first, it seems easily resolvable.

cmathis:
When you say take a stab at this, do you mean a better description, or do you have a better idea for how we capture that requirement? I'm open for whatever.

jwen-adobe:
@trieloff @cmathis I rolled back metrics schema definition to use the URIs after discussing with the team. 

One qq, is the metric property field below added intentionally? It does not belong to any metric categories and was defined to be used by order schema before.

"https://ns.adobe.com/xdm/data/discount": {
      "$ref": "https://ns.adobe.com/xdm/data/measure"
    },

trieloff:
@jwen-adobe no, that was an oversight. I went through the metrics somewhat mechanically.

trieloff:
@cmathis: better description of the intent.

schemas/data/measure.schema.json
```diff
@@ -20,6 +20,12 @@
           "description":
             "Links to the `Metric` that this measure refers to. The `@type` can be omitted in contexts where it is obvious what metric is being referred to, for instance in the `Metrics` object that is keyed with `Metric` URIs."
         },
+        "@id": {
+          "title": "Unique Identifier",
+          "type": "string",
+          "format": "uri",
+          "description": "Unique identifier of the metric."
```
trieloff: Shouldn't this be: "Unique identifier of the *measure*? We have the metric identified using `@type` above.

If yes: then let's also turn this into a `xdm:id` or `xdm:uniqueMeasure`, make it a plain `string` and use a description along the lines of: 
> In cases of data collection using lossy communication channels, such as mobile apps or websites with offline functionality, where transmission of measures cannot be ensured, this property contains a client-generated, unique ID of the measure taken. 
> It is best practice to make this sufficiently long to ensure enough entropy. Additionally, if information such as time stamp, device ID, IP or MAC address, or other potentially user-identifying values are incorporated in the generation of the `xdm:uniqueMeasure`, the result should be hashed, so that no PII is encoded in the value, as the goal is not to identify user or device, but the specific measure in time.

It is probably also a good idea to set a minimum length of 20 characters.

If not: then just close the PR because it's a duplicate.

jwen-adobe: @trieloff updated.

trieloff:
@jwen-adobe can you fix the linting error? Otherwise it looks good to merge.

PR #131, opened 2018-02-20
Abbreviations are no acronyms,

trieloff:
technically, but should be treated the same.

Please link to the issue #…


cmathis:
I don't have a strong opinion on this. If all acronyms and abbreviations need to be capitalized then so be it.

Issue #133, opened 2018-02-20
Add new schemas from DMa local repo: segment, propensity


jwen-adobe:
## What are the schemas that are affected by the issue
new schemas
-segment
-propensity

## What are examples of products that are impacted by the issue
platform xdm definition

jwen-adobe:
Close this. @cmathis @cdegroot-adobe

PR #110, opened 2018-02-14
Add new schemas from platform

jwen-adobe:
This PR links to issue #109 

@trieloff @kstreete @cmathis @cdegroot-adobe  cdegroot@adobe.com

Add following new schemas
-identity
-namespace
-times
-openinghours
-segment
-store
-propensity

PR #135, opened 2018-02-20
WIP - Add new schemas from DMa platform: segment and propensity

jwen-adobe:
This PR links to the issue #133 

Add new schemas below, see old comments from Lars here: https://github.com/adobe/xdm/pull/110
-segment
-propensity


cmathis:
These are two other schemas that I have had very little experience with in the past. I know @cdegroot-adobe has been driving multiple discussions with the Profile Service and others around this segment schema. 
 
@trieloff - This is another area we should start off with what ever questions or suggestion you may have. Unlike the ExperienceEvent and Profile schemas I don't know how much these are being used today.

schemas/data/propensity.schema.json
```diff
@@ -0,0 +1,65 @@
...
+    "description": "The propensity schema represents the likelihood of a certain action happening in a certain window of time. The propensity schema can host various predictive results e.g churn. This is a general-purpose schema that is designed to contain the results of any propensity scoring algorithm.",
+    "definitions": {
+        "propensity": {
+            "properties": {
+                "xdm:score": {
+                    "title": "Score",
+                    "type": "number",
+                    "description": "The score is a real number between 0 and 1 (both inclusive). A higher value indicates a higher likelihood of the action occurring."
+                },
+                "xdm:action": {
```
kstreeter: It's a little weird that this property is called "action", but the value is expected to map to a metric.  It certainly makes sense that this maps to a metric, but shouldn't we then just call the property metric?

schemas/data/propensity.schema.json
```diff
@@ -0,0 +1,65 @@
...
+                    "title": "Reason Code",
+                    "type": "string",
+                    "description": "The code that represents the algorithmic selected reason that the action is likely."
+                },
+                "xdm:reasonText": {
+                    "title": "Reason Text",
+                    "type": "string",
+                    "description": "The description text that represents the algorithmic selected reason that the action is predicted to occur."
+                },
+                "xdm:timeWindow": {
```
kstreeter: How is this defined?  Is this a duration or a span?

schemas/data/propensity.schema.json
```diff
@@ -0,0 +1,65 @@
...
+                    "meta:enum": {
+                        "seconds": "Seconds",
+                        "minutes": "Minutes",
+                        "hours": "Hours",
+                        "days": "Days"
+                    }
+                },
+                "xdm:algorithmId": {
+                    "title": "Algorithm Id",
+                    "type": "string",
```
kstreeter: should be a URI

schemas/audiences/segment.schema.json
```diff
@@ -0,0 +1,61 @@
...
+  "$id": "https://ns.adobe.com/xdm/audiences/segment",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Segment Assignment",
+  "type": "object",
+  "description":
+    "",
+  "definitions": {
+    "segment": {
+      "properties": {
+        "@id": {
```
trieloff: As this is not the full segment or audience, but only the assignment, it might be better to make this a `@type` (or introduce a `@type` in addition to the `@id`)

schemas/audiences/segment.schema.json
```diff
@@ -0,0 +1,61 @@
...
+          "title": "Identity",
+          "type": "string",
+          "format": "uri",
+          "description": "Unique identity of the segment."
```
trieloff: … assignment

schemas/audiences/segment.schema.json
```diff
@@ -0,0 +1,61 @@
...
+        },
+        "xdm:endUserIDs": {
+          "title": "End User IDs",
+          "$ref": "https://ns.adobe.com/xdm/context/enduserids",
+          "description":
+            "Condensed, normalized encapsulation of an end user identifier that the segment is assigned to."
```
trieloff: Ok, here is my interpretation: a person (`Profile`) belongs to a somehow defined group of people (`Audience`). This declaration of membership is made in this schema, as a "Segment Assignment".

So, what's the role of `EndUserIDs`?

Some applications that perform an assignment perform the assignment based on `EndUserIDs`, so it makes sense to track this.

If my interpretation is correct, we need to rephrase the description to something like:

> The list of `EndUserIDs` that the segment assignment is based on.

schemas/audiences/segment.schema.json
```diff
@@ -0,0 +1,61 @@
...
+        },
+        "xdm:name": {
```
trieloff: I'd remove that property, as we can look it up in the segment/audience itself.

schemas/audiences/segment.schema.json
```diff
@@ -0,0 +1,61 @@
...
+          "title": "Name",
+          "type": "string",
+          "description": "The common name associated with the segment."
+        },
+        "xdm:assignedOn": {
```
trieloff: I think we should use the properties we are using elsewhere
- `xdm:repositoryLastModifiedBy`
- `repo:lastModifiedDate`

schemas/audiences/segment.schema.json
```diff
@@ -0,0 +1,61 @@
...
+          "format": "date-time",
+          "description": "The date the segment was assigned to the profile."
+        },
+        "xdm:validUntil": {
+          "title": "Valid Until",
+          "type": "string",
+          "format": "date-time",
+          "description": "The date the segment assignment is valid until."
+        },
+        "xdm:assignedBy": {
```
trieloff: I think we should use the properties we are using elsewhere
- `xdm:repositoryLastModifiedBy`
- `repo:lastModifiedDate`

If the assignment has been made by a program, let's link to "Namespace"

schemas/audiences/segment.schema.json
```diff
@@ -0,0 +1,61 @@
...
+  ],
+  "$id": "https://ns.adobe.com/xdm/audiences/segment",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Segment Assignment",
+  "type": "object",
+  "description":
+    "",
+  "definitions": {
+    "segment": {
+      "properties": {
```
trieloff: I think a `xdm:confidence` property is missing here to capture the confidence of the segment assignment, as outlined in #110

schemas/data/propensity.schema.json
```diff
@@ -0,0 +1,71 @@
...
+  "$id": "https://ns.adobe.com/xdm/data/propensity",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Propensity",
+  "type": "object",
+  "description":
+    "The propensity schema represents the likelihood of a certain action happening in a certain window of time. The propensity schema can host various predictive results e.g churn. This is a general-purpose schema that is designed to contain the results of any propensity scoring algorithm.",
+  "definitions": {
+    "propensity": {
+      "properties": {
+        "xdm:score": {
```
trieloff: Elsewhere we are using `xdm:confidence` (in `EndUserIDs`) and we should be using it in Segment Assignment. I'm fine using the name "Score" in the title or description, but we should keep all examples of modeling confidence intervals the same.

schemas/data/propensity.schema.json
```diff
@@ -0,0 +1,71 @@
...
+          "title": "Score",
+          "type": "number",
+          "description":
+            "The score is a real number between 0 and 1 (both inclusive). A higher value indicates a higher likelihood of the action occurring."
```
trieloff: Add `min` and `max` to enforce this.

schemas/data/propensity.schema.json
```diff
@@ -0,0 +1,71 @@
...
+        },
+        "xdm:action": {
```
trieloff: From @kstreeter (before reformatting moved stuff around)
> It's a little weird that this property is called "action", but the value is expected to map to a metric. It certainly makes sense that this maps to a metric, but shouldn't we then just call the property metric?

trieloff: I can see at least two ways of modeling propensity:

1. the likelihood of an event occurring (within a certain time frame)
2. the likelihood of observing a certain metric reaching or exceeding a certain value

In both cases it makes sense to have a stronger pointer (URI) here. In the second case, we also need to define what the threshold is and what kind of threshold (above, below) it is.

trieloff: Probably more a @cdegroot-adobe then @jwen-adobe question.

schemas/data/propensity.schema.json
```diff
@@ -0,0 +1,71 @@
...
+          "title": "Action",
+          "type": "string",
+          "description":
+            "The action that the propensity is testing the likelihood of occurring. Selected from the metric types of Experience Event where applicable."
+        },
+        "xdm:reasonCode": {
+          "title": "Reason Code",
+          "type": "string",
```
trieloff: As long as this is just a string, it's not resolvable. Is that intentional?

schemas/data/propensity.schema.json
```diff
@@ -0,0 +1,71 @@
...
+          "description":
+            "The code that represents the algorithmic selected reason that the action is likely."
+        },
+        "xdm:reasonText": {
+          "title": "Reason Text",
+          "type": "string",
+          "description":
+            "The description text that represents the algorithmic selected reason that the action is predicted to occur."
+        },
+        "xdm:timeWindow": {
```
trieloff: From @kstreeter 
> How is this defined? Is this a duration or a span?

trieloff: I'd add: let's merge this with `xdm:unitOfTime` and just call it something like `xdm:until`, make it a `date-time` and call it the "time until the user will be observed to have performed the action or a measure corresponding to the defined threshold has been reached"

schemas/data/propensity.schema.json
```diff
@@ -0,0 +1,71 @@
...
+          "meta:enum": {
+            "seconds": "Seconds",
+            "minutes": "Minutes",
+            "hours": "Hours",
+            "days": "Days"
+          }
+        },
+        "xdm:algorithmId": {
+          "title": "Algorithm Id",
+          "type": "string",
```
trieloff: From @kstreeter 
> should be a URI

trieloff: +1

schemas/data/propensity.schema.json
trieloff: Although if I understood @cdegroot-adobe's comment in #110 correct, we should establish the use case before we introduce this property at all.

jwen-adobe:
Close this. @cmathis @cdegroot-adobe

Issue #137, opened 2018-02-21
Update profile with latest changes from DMa xdm schemas


jwen-adobe:
## What are the schemas that are affected by the issue
profile

## What are examples of products that are impacted by the issue
platform xdm definition

PR #138, opened 2018-02-21
Update profile to the latest from DMa

jwen-adobe:
This PR links to the issue #137 

Updated profile to latest from DMa. As profile also referenced identity and namespace, these are also included here, but will be updated based on the changes from https://github.com/adobe/xdm/pull/114


trieloff:
I'd suggest we review this once #114 has made it in or remove the changes that are also covered in #114, so we don't have the same discussion in two places (maybe with different outcomes)

trieloff:
@jwen-adobe please remove the files that overlap with #114, which has been merged and is now causing a conflict.

jwen-adobe:
@trieloff I will close this PR and create a new one which is clean.

Issue #142, opened 2018-02-27
experienceevent web field


cmathis:
@trieloff and @kstreeter 

In the experienceevent schema today we have this xdm:web field that is referencing a channel/web schema that does not have the information that this was originally meant to store. This xdm:web field is intended to store web related information about the page meta-data (name, isErrorPage, isHomepage, referrer, etc.) and link meta-data (linkName, URL, etc.). The channel/webpage and context/weblink are still around, but they are not being used. The original context/web that we pushed over included those, but somewhere this was removed and the reference was change to channel/web which seems like a different purpose. 

Wondering what we should do in this case. Do we add the webpage and weblink references to the channel/web or bring back the context/web as before with those included. 

cc @jwen-adobe

jwen-adobe:
A PR is created for creating another web schema under context folder while keep the channels/web as is.  Another option is replace the web with webpage and weblink in the experienceevent schema.
PR: https://github.com/adobe/xdm/pull/143

trieloff:
Notes from conversation with @kstreeter 
- [ ] channel independent representation will make it harder to produce, but easier to consume (we prioritize consumability)
- [ ] we need to look at page name as a way to identify channel locations in places where URLs are absent (mobile) or weak (legacy web apps)
- [ ] we need good documentation on best practices for creating experience events, so that crucial properties aren't forgotten
- [ ] we need more (and channel-specific) examples for experience events

trieloff:
I will try to create an alternative PR for this.

cmathis:
@trieloff - Are you saying an alternative PR to the existing #143?  

If this the case please let me know what we can do to speed this along. Teams are expecting this expereinceevent schema to be stable right after Summit. We have another small change we may need to make, but I'm waiting to see what we end up with here before I do anything else. 

cc @kstreeter @cdegroot-adobe

trieloff:
Please give feedback on #186

Issue #100, opened 2018-01-23
Experience Event: should commerce be transaction?


trieloff:
Should we rename the commerce property to transaction?

cmathis:
"commerce" is what has been used by data producers for a few months now.  Not sure we can handle a change at this time.

trieloff:
I get that a change would be inconvenient, but my question is a about the scope of the property: are we planning to restrict this only to commerce (buying goods) or all kinds of transactions (managing subscriptions, memberships, etc.)

If it's the former, we should stay with `commerce`. If it's the latter, we need to decide if it makes sense to have either both a `commerce` and a `transaction` property, and where the line between the two would be, or to expand the scope and name of the `commerce` property accordingly.

For data producers that are posting data according to a pre-1.0 version of XDM, you might want to consider adapters to re-write the data, but until we've reached 1.0, there can and will be changes. (Unless we agree to change the rules)

cdegroot-adobe:
```transaction``` is a subset of ```commerce```, the intent of this area is more then a "transaction" so I am much more comfortable with ```commerce```.

trieloff:
Ok.

PR #143, opened 2018-02-27
Add web schema in context folder for experience event

jwen-adobe:
This PR links to the issue #142 

@trieloff @kstreeter @cmathis

This PR add web schema under context folder which will be used by experienceevent. Currently, the experienceevent is using the web under channel folder which is not the same as before the merge.


cmathis:
I added some details in the issue here - https://github.com/adobe/xdm/issues/142

kstreeter:
Yes, looks like this was dropped in the refactor.  I agree we should put back.  My only question is naming: "Web" as channel name makes sense, but this schema feels like an intermediate container and just calling it "Web" seems nebulous.  Can we call it "WebInfo" or "PageInfo" or something to differentiate from web as a channel?

jwen-adobe:
@kstreeter Renamed to webinfo.

trieloff:
I wonder if we can find a better way to encapsulate the meaning of these properties in a way that is less connected to the web. The way I see this, we have following interesting properties:

- `xdm:page`: describes the location within a channel where the ExperienceEvent was observed
- `xdm:link`: describes the referrer, i.e. the location in the same or another channel

None of this is really web-related or web-only, and the `link` property is semantically much closer to the `marketing` properties we have right now: campaign code, etc.

How much of an issue would it be to split this along the lines of current location/referring location instead of the channel-split we have right now?

kstreeter:
@trieloff, I believe that this does signal that the EE "hit" was a web interaction.  So you are right that the page/link info is not web-only, but they are encapsulated in this way for web interactions.  We had considered placing this into channels/Web, but that is a channel definition, not a container for aspects of the web interaction.
A possible approach here would be to call it "webinfo", but include a "channel" property that links this back to the web channel as a concept.  We could encapsulate this pattern into a "channelinfo" base schema.

cmathis:
The original intention of the "web", "commerce", and "marketing" fields what to provide the additional meta-data that related to the "web",  "commerce", or "marketing" metrics that were set in the record being passed in. Given that the commerce and and marketing ended up as context schemas, it seems like this web would be a similar context schema. It's there to provide additional details about that metric.web.pageView or metric.web.link that occurred.

kstreeter:
@cmathis, yes I have had this discussion with @cdegroot-adobe as well.  There are two orthogonal concepts: channels, which are methods for touching customers (email, etc), and there are vertical domains or activities (commerce, marketing, etc).  "web" just happens to exist in both. We shouldn't combine the two concepts, we need to allow both to continue to exist independently.

I feel like this is a good argument for reintroducing context/web.

trieloff:
> There are two orthogonal concepts: channels, which are methods for touching customers (email, etc), and there are vertical domains or activities (commerce, marketing, etc).

Fully agree. Channels constrain how content is delivered and metrics are collected, and represent a technical constraint. 

Vertical activities can be performed independently from the channel, as long as it adheres to the constraints of the channel (hence the existence of *-commerce).

> "web" just happens to exist in both.
> I feel like this is a good argument for reintroducing context/web.

I agree, we should reintroduce the context/web concept, but we should not call it web, in order to avoid confusion with the channel and instead capture the essence of the metrics that get captured.

Here's how I see what we currently have:

| Container   | Properties     | What it captures                                                                    |
| ----------- | -------------- | ----------------------------------------------------------------------------------- |
| `commerce`  | `order`        | What people are buying, i.e. transactions, conversions, or more generally *conduct* |
| `marketing` | `trackingCode` | How people got to the experience, i.e. referral, history, or *induction*            |
| `web`       | `page`         | The (virtual) place within the channel where the experience happened, i.e. location |
| `web`       | `link`         | How people got to the experience, i.e. referral again.                              |

While I have mostly conceded defeat in renaming `commerce` (#100), I think it makes a lot of sense to regroup `marketing` and `web` so that they are not overly generic (marketing can be a lot of things) and easily confused with a channel.

This could look like this:

| Container  | Properties     | What it captures                                                                    |
| ---------- | -------------- | ----------------------------------------------------------------------------------- |
| `commerce` | `order`        | What people are buying, i.e. transactions, conversions, or more generally *conduct* |
| `referral` | `trackingCode` | How people got to the experience, i.e. referral, history, or *induction*            |
| `referral` | `link`         | How people got to the experience, i.e. referral again.                              |
| `location` | `page`         | The (virtual) place within the channel where the experience happened, i.e. location |

I'm happy to adjust the PR according to this proposal.

cmathis:
As part of this do we change the metric field "https://ns.adobe.com/xdm/data/metrics/**web**/page-views" to be "https://ns.adobe.com/xdm/data/**metrics**/referral/page-views" or is metrics always referring to the channel concept?

cmathis:
@cdegroot-adobe - This is your call here. Looks like the change would be to convert the existing "web" and "marketing" schemas (and probably field names in EE) into "referral" and "location".

cdegroot-adobe:
@trieloff if we want to go down the road of a channel neutral representation, I think that we should consider a context of "experience", with properties like name, url, category (for web name would be equivalent to pageName in Analytics). Marketing as a container would contain things like trackingCodes, campaigns etc.

trieloff:
>  Marketing as a container would contain things like trackingCodes, campaigns etc.

100% agree.

> I think that we should consider a context of "experience", with properties like name, url, category (for web name would be equivalent to pageName in Analytics)

Agree partially. `name`, `url`, `category` should be part of the container. In my mind, the container is describing the addressable place (in the virtual world) where the interaction happened. I typically call this the location and consider it to be part of the channel and dependent on the channel (e.g. a mobile app will have a different way of addressing screens in an app than a website would address pages in or areas on a page)

I've seen the term "experience" used in two ways before:

1. as a specific set of content that is assigned to specific locations and can be consumed by one or multiple audiences. For instance, when you see a hero image on the home page, a small note on the product page, an overlay in the shopping cart, and a line in the confirmation mail asking you to sign up for a special promotion for Canadian customers, that would be an experience. 
2. as a synonym for "campaign", i.e. a collection of who, where, what decisions (audience, location/channel, content) that all serve the same why (metrics/goals).

I haven't seen it used as a synonym for what I'd call "location", although I'd agree that "location" probably could be refined to avoid collisions with places in the real world (that we already have modeled)

cdegroot-adobe:
@trieloff @kstreeter 
One test I have is to consider any change from the perspective of a cross channel query. If the query has to have a test of(if {some source semantic} then {look here to read a common thing}). If that is ever true then I think the decision is not leading us to an omni channel direction. Thus I have been concerned with the web context from the start. Moving it into channel is making it worse, not better.

```
I've seen the term "experience" used in two ways before:

1. as a specific set of content that is assigned to specific locations and can be consumed by one or multiple audiences. For instance, when you see a hero image on the home page, a small note on the product page, an overlay in the shopping cart, and a line in the confirmation mail asking you to sign up for a special promotion for Canadian customers, that would be an experience.
2. as a synonym for "campaign", i.e. a collection of who, where, what decisions (audience, location/channel, content) that all serve the same why (metrics/goals).
```

I do not think either of these is acceptable as guidance so we need to really define it as we think is right longer term.

My proposal is therefore to have a single common representation of details about the experience irrespective of source/channel.

```
experience.name = web pageName, video scene/chapter/segment name, mobile app experience name, call center route name.
```
This to me is quite fundamental to making ExperienceEvents source independent. If you look at the asset XDMs they do try unify common properties where they can, like height/width. I see this as the same.

I am not so concerned about the name, more the usage pattern.

kstreeter:
Very much agree with this @cdegroot-adobe.  I think @trieloff 's proposal is compatible with that, as it places focus on the role of the information in a channel-neutral and activity-neutral way.

Are there places where you feel we have made things too channel-centric?  We should fix those as a part of this change, too.

cdegroot-adobe:
@kstreeter @trieloff 
On, I am good with this PR, lets move it in. 

We are left with the open question about where to place data like 'page name' in a general location.

kstreeter:
@cdegroot-adobe, @trieloff after spending some time discussing with folks I now feel like keeping the "web" container is the way to go.  The thinking is related to your comment Chris regarding where to put things like "page name".

While organizing things around neutral semantics like "location" or "referrer" is reasonable, it has the problem that whenever we want to extend our schemas into a new area (for example, mobile apps) we have to find buckets for all the elements in that area, possibly creating new ones, etc.  So extending into a new environment/area will always involve changing multiple schemas.

Alternatively, we can organize things based on the environment/area, and add everything related to that in one place.

I am not claiming one of these organizations is strictly better than the other, they are just different ways to organize the information and are both valid.  So we probably need to pick a methodology based on most common use.  The feedback I have gotten is that most folks think of that data based on the area (like "web" or "mobile") versus looking at things like "what are all the ways I can have a referrer over all areas"?  

For this reason, my vote is to keep the existing organization.

schemas/context/webinfo.example.1.json
```diff
@@ -0,0 +1,15 @@
+{
+  "xdm:page": {
```
lrosenthol: xdm;page is *NOT* a web page.  We've been over this one before...call it xdm;webpage if you wish.

schemas/context/webinfo.schema.json
```diff
@@ -0,0 +1,35 @@
...
+  "$id": "https://ns.adobe.com/xdm/context/webinfo",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Web Information",
+  "type": "object",
+  "description":
+    "Information about the web page and/or link of the ExperienceEvent.",
+  "definitions": {
+    "webinfo": {
+      "properties": {
+        "xdm:page": {
```
lrosenthol: xdm;page is NOT a web page. We've been over this one before...call it xdm;webpage if you wish.

schemas/context/webinfo.schema.json
```diff
@@ -0,0 +1,35 @@
...
+          "title": "Web Page",
+          "$ref": "https://ns.adobe.com/xdm/channels/webpage",
+          "description": "The web page information of the ExperienceEvent."
+        },
+        "xdm:link": {
```
lrosenthol: xdm:link is *not* just a web link. It could be a link to many other things...

cmathis:
Have we come to any conclusions here? I think Kevin's last comment explains the situation well. We use top level fields/sub-schemas in experienceevent to group data based on a common type or environment. This web environment/context is just one of those. 

This PR has the web related information in a schema named xdm/context/webinfo to differentiate it from xdm/channel/web. Happy to convert this back to xdm/context/web or whatever we need to do finish this up. This xdm:web field in experienceevent is the last to be migrated over.

cdegroot-adobe:
@lrosenthol can you please comment on the response of @cmathis. Are you saying that all fields in the "xdm:*" naming needs to be either unique or mean exactly the same thing? Irrespective of the json hierarchy?

cdegroot-adobe:
@kstreeter @trieloff  I would like to factor out the "xdm:name" field of channels/webpage.schema.json into a channel neutral location so it is the same for all channels. I think this will help omni channel queries.

kstreeter:
I think we have been a little loose about keeping the strict "JSON-LD interpretation" of property names, but in general I don't think its a bad idea to try to keep individual fields as semantically meaningful as possible.  I agree with @lrosenthol here that in this area, "pages", we would do ourselves a service and have clear and distinct semantics between web pages and pages in other document types.

cmathis:
Let's rename the fields to xdm:webpage and xdm:weblink.

Another problem I see with this PR is that the webpage.schema is in the "channel" namespace/folder and it should be under "context" next to weblink.schema. I do not have access to update this branch so I will have Joe make that change.

jwen-adobe:
Changes made. while moving webpage.schema, I also made the field name changes mentioned in the PR here: https://github.com/adobe/xdm/pull/174. I will close the PR 174 as it is not needed anymore.

schemas/context/webpage.schema.json
```diff
@@ -50,17 +50,17 @@
...
           "description": "The referrer of the web page.",
           "properties": {
-            "URL": {
+            "xdm:URL": {
               "title": "URL",
               "type": "string",
               "description": "The referrer URL."
             },
-            "domain": {
+            "xdm:domain": {
```
lrosenthol: Why do you need to separate out the domain if you have the full URL?

lrosenthol:
I don't understand the difference between a webpage and a weblink - and the descriptions don't help me here.   Can we put in some better explanations for folks that aren't familiar with this space?

trieloff:
As far as I understand: web page is the page where the interaction has been observed. Web link is how the visitor got there.

lrosenthol:
then perhaps the latter should be something like referrerLink or sourceLink or something??

schemas/context/webpage.schema.json
kstreeter: @jwen-adobe, can you answer this for @lrosenthol?  I am also unclear why this is needed.

jwen-adobe: This field was added from Analytics. I do not know Chris Luby's public repo id, will ask him through email to see if we can remove this. @kstreeter

jwen-adobe: @kstreeter @lrosenthol xdm:domain is removed for now after communicating with analytics.

lrosenthol:
There appears to still be some inconsistencies with webLink & webPage - the names, the descriptions, etc.  However, it's also difficult in this PR to really see what we're working with.

trieloff:
@lrosenthol please take a look at #186 – maybe that's a clearer example that we can use as a starting point.

lrosenthol:
@trieloff I don't see how #186 relates to this one, sorry.

schemas/context/experienceevent.schema.json
```diff
@@ -84,7 +84,7 @@
         },
         "xdm:web": {
```
lrosenthol: this should also be webinfo to match the new type.

schemas/context/webinfo.schema.json
```diff
@@ -0,0 +1,35 @@
+{
+  "meta:license": [
+    "Copyright 2017 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/context/webinfo",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Web Information",
```
lrosenthol: Web Page Information?

schemas/context/webinfo.schema.json
```diff
@@ -0,0 +1,35 @@
...
+  "type": "object",
+  "description":
+    "Information about the web page and/or link of the ExperienceEvent.",
```
lrosenthol: Why is this only relevant to an experienceevent?  Shouldn't this be a general purpose webinfo object?

schemas/context/webinfo.schema.json
```diff
@@ -0,0 +1,35 @@
...
+  "definitions": {
+    "webinfo": {
+      "properties": {
+        "xdm:webPage": {
+          "title": "Web Page",
+          "$ref": "https://ns.adobe.com/xdm/context/webpage",
+          "description": "The web page information of the ExperienceEvent."
```
lrosenthol: Replace with something like:
This is the information about the page (URL) itself.

schemas/context/webinfo.schema.json
```diff
@@ -0,0 +1,35 @@
...
+        },
+        "xdm:webLink": {
+          "title": "Web Link",
+          "$ref": "https://ns.adobe.com/xdm/context/weblink",
+          "description": "The web link information of the ExperienceEvent."
```
lrosenthol: Replace with something like:
This is the information about the referrer link (URL) for the page.

schemas/context/webpage.schema.json
```diff
@@ -5,7 +5,7 @@
     "you may not use this file except in compliance with the License. You may obtain a copy",
     "of the License at https://creativecommons.org/licenses/by/4.0/"
   ],
-  "$id": "https://ns.adobe.com/xdm/channels/webpage",
+  "$id": "https://ns.adobe.com/xdm/context/webpage",
   "$schema": "http://json-schema.org/draft-06/schema#",
   "title": "Web Page",
   "type": "object",
```
lrosenthol: For Description:
Replace with something like:
This is the information about the page (URL) itself.

For Section:
What is a section?  There is no relevant term in HTML or the Web - please use a term that relates, or at least describe the relationship in the description.

For name:
Is this the title of the page, as defined by the title tag?  If so, say so, if not, explain what it is.

For server:
A hostname has nothing to do with a server.  What is this? It's not used in the example.

Error page:
What is an "error page"?  Can you refer to some specification that describes what one is?

Home Page:
What is a "home page"?  Can you refer to some specification that describes what one is?

schemas/context/webpage.schema.json
```diff
@@ -50,17 +50,12 @@
...
-            "domain": {
-              "title": "Domain",
-              "type": "string",
-              "description": "The domain of the referrer."
-            },
-            "type": {
+            "@type": {
               "title": "Type",
               "type": "string",
               "description": "The referrer type."
```
lrosenthol: What are the possible values here?  Do they come from an existing standard?  If so, please cite.

kstreeter:
Hi @lrosenthol I just posted an update that hopefully addresses your comments.  I have revised some of the field names and improved descriptions, as well as updated the examples.  Let me know what you think!

cc/ @jwen-adobe to make sure I didn't break anything.

schemas/context/weblink.schema.json
```diff
@@ -9,7 +9,8 @@
   "$schema": "http://json-schema.org/draft-06/schema#",
   "title": "Web Link",
   "type": "object",
-  "description": "Information about the web link of the ExperienceEvent.",
+  "description":
+    "Details about the web link (URL) that corresponds to the web page where the interaction occurred, as recorded by an `ExperienceEvent`",
```
trieloff: This description is very hazy to me, and easily confused with the description for `webPage`. It would probably make sense to move both into a separate `*.description.md`

The way I see it, we have two things that are relevant here:
- the web page where an interaction is happening
- the hyperlink that has lead to the page where the interaction is recorded

The description as it stands, says that `webpage` is the page where the interaction happened. And `weblink` is about this, too.

schemas/context/weblink.schema.json
```diff
@@ -27,12 +28,13 @@
         "xdm:URL": {
           "title": "URL",
           "type": "string",
-          "description": "The link URL."
+          "description": "The actual link/URL used for this web interaction"
```
trieloff: This is adding further confusion, because while a link *points to* a URL, and *exists at* another URL, it *is not* a URL itself. If we want to model this, we need to make clear which URL this is:
- the one where the interaction is recorded
- the one where the interaction was triggered
- the one that was being linked to

schemas/context/webpage.schema.json
```diff
@@ -0,0 +1,103 @@
+{
+  "meta:license": [
+    "Copyright 2017 Adobe Systems Incorporated. All rights reserved.",
```
trieloff: 2018

schemas/context/webinfo.schema.json
```diff
@@ -0,0 +1,37 @@
+{
+  "meta:license": [
+    "Copyright 2017 Adobe Systems Incorporated. All rights reserved.",
```
trieloff: 2018

schemas/context/webpage.schema.json
```diff
@@ -0,0 +1,103 @@
...
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/context/webpage",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Web Page",
+  "type": "object",
+  "description":
+    "Details about the web page where a web interaction occurred, as recorded by an `ExperienceEvent`.",
```
trieloff: Please move this into a separate `webpage.description.md`

schemas/context/webpage.schema.json
```diff
@@ -0,0 +1,103 @@
...
+          "title": "Site Section",
+          "type": "string",
+          "description":
+            "The normative name of the site section where this web page resides, which may be used to classify or categorize the interaction"
+        },
+        "xdm:name": {
+          "title": "Name",
+          "type": "string",
+          "description":
+            "The normative name of the web page. This name is not necessarily the page title or directy associate with page content, but is used to organize a site's pages for classification purposes"
```
trieloff: typo: directly.

schemas/context/webpage.schema.json
```diff
@@ -0,0 +1,103 @@
...
+        },
+        "xdm:URL": {
+          "title": "URL",
+          "type": "string",
+          "format": "uri",
+          "description":
+            "The normative or usual URL of the web page.  This may or may not be the actual URL used to reach the page, which would be recorded using `WebLink`."
+        },
+        "xdm:server": {
```
trieloff: I think the name `server` is a clear misnomer here, when we are talking about a hostname.

schemas/context/webpage.schema.json
```diff
@@ -0,0 +1,103 @@
...
+          "title": "Server",
+          "type": "string",
+          "format": "hostname",
+          "description":
+            "The normative or usual server that hosts the web page.  This may or may not be the host or server that actually served the page interaction, but is used for classification purposes."
```
trieloff: Again, I'd put the focus not on which server is doing the hosting, but under which hostname the experience can be accessed.

The fact that there is a real (technical) and virtual (used for classification purposes) hostname is fine.

schemas/context/webpage.schema.json
```diff
@@ -0,0 +1,103 @@
...
+        },
+        "xdm:isErrorPage": {
```
trieloff: Shorter: `xdm:error`

schemas/context/webpage.schema.json
```diff
@@ -0,0 +1,103 @@
...
+          "title": "Is Error Page",
+          "type": "boolean",
+          "description":
+            "Flag that indicate if the page is error page or not.  Error here is defined by the application, and may nor may not correspond to a page served with an HTTP error code.  This flag is used to broadly categorize web interactions."
+        },
+        "xdm:isHomePage": {
```
trieloff: Shorter: `xdm:home`

schemas/context/webpage.schema.json
```diff
@@ -0,0 +1,103 @@
...
+          "title": "Is Home Page",
+          "type": "boolean",
+          "description":
+            "Flag that indicate if the page is the site home page or not.  The definition of home page is determined by the application, but is commonly used to designate a top level landing page or common site entry point.  This flag is used to broadly categorize web interactions."
+        },
+        "xdm:referrer": {
```
trieloff: Making `xdm:referrer` part of `webpage` is very counterintuitive. I think `webpage` should model the page where something is happening. How someone got there is a different container (and `weblink` looks very much like the correct container)

schemas/context/webpage.schema.json
```diff
@@ -0,0 +1,103 @@
...
+            "xdm:URL": {
+              "title": "URL",
+              "type": "string",
+              "description": "The referrer URL."
+            },
+            "xdm:referrerType": {
+              "title": "Referrer Type",
+              "type": "string",
+              "description": "The referrer type.",
+              "enum": [
```
trieloff: Let's use `snake_case` for `enum` keys.

schemas/context/weblink.schema.json
cdegroot-adobe: @trieloff - No, you understand this incorrectly. There are two kinds of web events, a PageView(a full new loading of a web page) and a Page Event which is an action happening inside a page that has already had a PageView. The latter does not increment a page count, but will be used in single page hits to indicate that the user really interacted with the page and thus take it out of a bounce count.
@kstreeter FYI

schemas/context/webinfo.example.1.json
```diff
@@ -0,0 +1,19 @@
...
+    "xdm:isErrorPage": false,
+    "xdm:isHomePage": true,
+    "xdm:referrer": {
+      "xdm:URL": "https://www.some-adserver.com",
+      "xdm:referrerType": "external"
+    }
+  },
+  "xdm:webLink": {
+    "xdm:type": "other",
+    "xdm:URL": "https://www.example.com/products/store/?view=1",
```
cdegroot-adobe: @kstreeter I recommend that we change this example to clearly state the link nature of an intra page action.
"xdm:URL": "https://www.example.com/products?view=stores",
"xdm:name": "product stores expanded"

This will better show the SPA nature of the link event.

schemas/context/weblink.schema.json
cdegroot-adobe: The Web Link is normally used for the single page app sort of case. So analogous to s.tl() in AppMeasurement. It is best to think of the first class nature of this being a Page Event. If this were a pure PageView then we may not have this sort of referrer information, if we do then good it can appear here. But the primary case is where there is a page loaded, and a local link (anchor) event occurs that indicates the end user drilling into some aspect of the page. Thus a "link-click" event would always have its primary data for the link being executed in this "Web Link"

schemas/context/weblink.schema.json
cdegroot-adobe: Its primary usage is "the one where the interaction was triggered". It is actually recorded after the link is fully executed if memory serves.

schemas/context/webpage.schema.json
cdegroot-adobe: No, this is the actual host name that served the root html page. It is used for debugging purposes mostly.

cdegroot-adobe: Actually thinking about it a bit, this variable is used in two different ways that are incompatible. It is used to hold the actual name of the server (hostname) and in other cases can be the domain. We cannot have this going forward. 
The domain should be extracted from the URL parameter

schemas/context/webpage.schema.json
cdegroot-adobe: I like isErrorPage as it is more descriptive. "error" maybe considered to have the definition of the error.

schemas/context/webpage.schema.json
cdegroot-adobe: I like "isHomePage" for the same reason I like "isErrorPage". do you have a stylistic norm to propose that we may be missing?

schemas/context/webpage.schema.json
cdegroot-adobe: +1

schemas/context/webpage.schema.json
cdegroot-adobe: I agree with your point about the incongruence of referrer in webpage. Lets put it in either the webLink or maybe more preferably in a peer to webpage and weblink.

schemas/context/webpage.schema.json
trieloff: I think having a `hostname` parameter that is able to override the hostname in the URL is sensible. e.g. you might want to treat traffic from `www-1.adobe.com` as if it originated from `www.adobe.com`. The point that I'm trying to make is that all of this is hostname arithmetics, not server arithmetics, because the traffic can come from the same or different servers (machines).

schemas/context/webpage.schema.json
trieloff: https://github.com/adobe/xdm/blob/master/CONTRIBUTING.md#design-for-consumability

> the principle of minimal verbosity: make keep things as short as possible, but not shorter

I would bargain for `isError`/`isHome` – and we can check if there are other `boolean` properties in our schemas that can be made to follow the same pattern.

schemas/context/webinfo.example.1.json
trieloff: @cdegroot-adobe  There's no harm in adding `webinfo.example.2.json`

schemas/context/weblink.schema.json
trieloff: Ah. That makes things a bit clearer. Let me replay that, so that we can transfer the emerging clarity into the schema:

- we have two types of interactions: an entire page is loaded vs. an interaction happens on a page that does not trigger a page load
- for each type of interaction we want to capture a couple of properties including where (URL, page name, link name)

The names we use are:

|What happened|Adobe Analytics|This PR|My Proposal|
|----------------|----------------|--------|------------|
|A page was loaded| Page View | `webinfo`| `webpageview` |
|An interaction on a page was recorded| Page Event | `weblink` | `webinteraction` |

@lrosenthol please take a look: I don't want to add further confusion

@cdegroot-adobe if that looks alright, let's rename the two buckets, and give them proper descriptions (too much is buried here in these review threads)

trieloff:
@kstreeter I think @cdegroot-adobe's comment cleared up a lot of the confusion, now we just need to reflect it in the model.

One thing that occurred to me is that it would be helpful to have an easy way to distinguish which type of `ExperienceEvent` we are dealing with: is it a full page view, is it an interaction on a page, is it a click on an email, etc. This has come up before (#144) and I think it might be worth revisiting, because inspecting the experience event for the presence of child properties might be a bit error-prone.

schemas/context/webreferrer.schema.json
```diff
@@ -28,15 +28,19 @@
...
-            "typedBookmarked": "Typed/bookmarked referrers are identified when visitors type your site’s URL directly into their browser, or if they access the site by selecting bookmarks.",
+            "search_engine": "Search engine referrers are identified when visitors use a search engine to access the site.",
+            "typed_bookmarked": "Typed/bookmarked referrers are identified when visitors type your site’s URL directly into their browser, or if they access the site by selecting bookmarks.",
             "email": "A referring domain is considered as an email referring domain type when visitors click an emailed message link containing the protocol imap:// or mail:// and arrive at the site.",
-            "social": "The referrer is identified as a social network."
+            "social": "The referrer is identified as a social network.",
+            "unknown": "The referrer is unidentifiable (unknown). `No JavaScript` could be the cause.",
+            "usenet": "The URL of a referrer started with news://. As such, the referrer link was posted on a Usenet newsgroup rather than a web page."
```
lrosenthol: usenet should instead be NNTP or news.  Usenet is/was the name of a whole network, not just one specific usecase.

schemas/context/webpage.schema.json
trieloff: `xdm:errorPage` it is then.

schemas/context/webpage.schema.json
trieloff: `xdm:homePage` it is.

trieloff:
@trieloff to fix:
- [ ] `webpageview`/`webinteraction ` https://github.com/adobe/xdm/pull/143#discussion_r180039989
- [ ] `xdm:homePage` https://github.com/adobe/xdm/pull/143#discussion_r180042047
- [ ] `xdm:errorPage` https://github.com/adobe/xdm/pull/143#discussion_r180042047

I will then approve this PR.

@kstreeter does the final review and merge.

cdegroot-adobe:
@trieloff  @kstreeter - Yes those made sense. I have incorporated the two changes and updated all examples. We should be good to merge now.

Issue #149, opened 2018-03-01
Change $ref to auditable by using auditable#/definition/auditlog


jwen-adobe:
## What are the schemas that are affected by the issue
address
person
emailaddress
phonenumber

## What are examples of products that are impacted by the issue
platform xed converter

PR #150, opened 2018-03-01
Change $ref to auditable by using auditable#/definitions/auditlog

jwen-adobe:
This links to the issue #149 

@trieloff @kstreeter @cmathis

The open source tooling works better with the proposed change when handling allOf merging for nested json structures.


cmathis:
This appears to be small change to way we reference the auditable schema. 

What are the rules for when you reference the base schema vs a definition in that schema?

jwen-adobe:
I do not think there is a rule in our case related to the $ref format at this point. The reason I make this change is to adapt the schema to a open source tooling while handing the nested structures contains same $ref. It seems the tooling works well with one format only (which maybe for a reason).

PR #151, opened 2018-03-01
Require property types

cdegroot-adobe:
It was found that with enumeration support that types can be variable. This is something that can cause great difficulties in data interoperability and code generation from JSON Schema. So to close this hole I am proposing that all properties have concrete types.

Example of what we do not want to allow is :
Even more than that: the enum constraint can go without accompanying type constraint and contain values of different type, e.g.
```json
{
"$id": "https://ns.adobe.com/xdm/sample1",
"definitions": {
"False": {
"enum": [0, false, "false"]
}
},
"$ref": "#/definitions/False"
}
```


kstreeter:
+1 on this, I think this has been the practice but is not in the docs.  Should we also include a recommendation to include min and max values or (or max length for string) where that is possible?

cdegroot-adobe:
@kstreeter in response to the min/max for numbers and max for strings yes, I agree, I intend to add that to an update on datatypes shortly

trieloff:
+1 to the original PR.

I'll open a separate for `string` handling and an improvement issue for enforcing this practice in the `meta.schema.json`

PR #152, opened 2018-03-02
Site/app implementation details

cdegroot-adobe:
Tracking the implementation details for how a request is made to an Experience Service from the client is important for support, debugging and usage analysis. This PR provides a way of providing this information in a structured way.

1. Created a new general purpose Implementation schema that records the API name and Version.
2. Defined the Adobe Extension that sets the location for implementation details for Analytics, Target and Identity Service. Will be used in future with ExperienceEvent.


cdegroot-adobe:
Need to correct syntax...

PR #154, opened 2018-03-02
Add schema descriptors for defining schema-level relationships

kstreeter:
This PR introduces a set of schemas that describe relationships and other schema-level metadata for a set XDM schemas.

The basic concept is that of a "descriptor", which encapsulates some information about schema (and possibly more specifically about a single property in the schema).  The descriptor has a type, which both describes the meaning of the descriptor, and signals the additional properties that might be present on the descriptor.  The type is defined using @type semantics, so it is extensible.

The descriptor types defined in this initial checkin:

- inter-schema relationships (1:1, 1:m, m:n)
- identities
- primary keys
- instantiable/uninstantiable
- update/merge policies


schemas/common/identitydescriptor.schema.json
```diff
@@ -0,0 +1,52 @@
...
+  "description":
+    "A descriptor that allows a schema property to be marked as an identity, even when it does not conform to the Identity schema.  This descriptor allows any value to be mapped into the Identity structure.",
+  "definitions": {
+    "identitydescriptor": {
+      "properties": {
+        "xdm:namespace": {
+          "title": "Namespace",
+          "type": "string",
+          "description":
+            "URI of the identity namespace this identity value belongs to"
```
trieloff: If it's a URI, then the property is missing a `format: uri`.

schemas/common/itemselector.schema.json
```diff
@@ -0,0 +1,72 @@
...
+  "meta:license": [
+    "Copyright 2017 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/common/itemselector",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Item Selector",
+  "meta:extensible": true,
```
trieloff: Should it be extensible? It looks like a low-level plumbing schema where we will be unlikely to be able to handle extensions in any implementation.

schemas/common/itemselector.schema.json
```diff
@@ -0,0 +1,72 @@
...
+                "When present, indicates the item with this @type value should be selected."
+            }
+          },
+          "required": ["xdm:type"]
+        },
+        {
+          "properties": {
+            "xdm:schema": {
+              "title": "Schema",
+              "type": "string",
```
trieloff: Should be `format: uri`

schemas/common/itemselector.schema.json
```diff
@@ -0,0 +1,72 @@
...
+          },
+          "required": ["xdm:id"]
+        },
+        {
+          "properties": {
+            "xdm:type": {
+              "title": "Type",
+              "type": "string",
+              "description":
+                "When present, indicates the item with this @type value should be selected."
```
trieloff: Should have a `format: uri`

schemas/common/schemadescriptor.schema.json
```diff
@@ -0,0 +1,75 @@
...
+  "meta:license": [
+    "Copyright 2017 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/common/schemadescriptor",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Schema Descriptor",
+  "meta:extensible": true,
```
trieloff: Do we need extensibility here?

trieloff:
@kstreeter I think I get what we're trying to achieve here: express relationships between schemas and handling of instances of data conforming to a schema that is independent from the JSON representation.

Apart from the stylistic questions mentioned in the review, I have a few more fundamental questions:

- why do we handle these relationships on the instance level (through schemas) instead of handling them on the schema level (through meta schemas)
- why do we use simple property name references instead of the more common JSON Path (that's already used in JSON Schema
- Can we have a bit more elaborate description on using each of these properties in `Contributing.md`

schemas/common/schemadescriptor.schema.json
kstreeter: Hi @trieloff, yes I believe that we do want to support extensibility of schemas descriptors.  The model here would be to create a new descriptor type (via @type), and extend this schema to include any additional properties needed by the new descriptor type.

kstreeter:
Thank you for your comments @trieloff, I will prepare an update that fixes some of the issues you noted.  In response to your questions:

> why do we handle these relationships on the instance level (through schemas) instead of handling them on the schema level (through meta schemas)

This is a good observation.  I think we need both.  It would be good to be able to tag a schema with something like "meta:descriptors" and list relevant descriptors.  I will add this to the PR.  But we also need to be able to refer to the descriptor schema directly, because we will be using it in our APIs that allow users to interactively create relationships across schemas.

>why do we use simple property name references instead of the more common JSON Path (that's already used in JSON Schema

> I need to think about this, but it seems reasonable.  We would still need an item descriptor, as I don't think JSON Pointer covers the matching semantics this implies.

> Can we have a bit more elaborate description on using each of these properties in Contributing.md

Definitely, I will prepare that.

schemas/common/updatepolicydescriptor.schema.json
```diff
@@ -0,0 +1,51 @@
...
+          "type": "string",
+          "description":
+            "The update policy to apply to entities of this schema",
+          "enum": [
+            "xdm:updateMerge",
+            "xdm:updateReplace",
+            "xdm:updateTimeSeries"
+          ],
+          "meta:enum": {
+            "xdm:updateMerge": "Merge:",
```
dmitry-shkolnik: Do we need a colon inside the value?

schemas/common/identitydescriptor.schema.json
```diff
@@ -0,0 +1,52 @@
...
+        },
+        "xdm:property": {
+          "title": "Namespace",
```
dmitry-shkolnik: title seems to be copied from a previous property

schemas/common/relationshipdescriptor.schema.json
```diff
@@ -0,0 +1,53 @@
...
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/common/relationshipdescriptor",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Relationship Descriptor",
+  "meta:extensible": true,
+  "meta:extends": [
+    "https://ns.adobe.com/xdm/common/schemadescriptor#/definitions/descriptor"
+  ],
```
dmitry-shkolnik: I'm not sure we need an inheritance graph across descriptors. They have almost nothing in common. I would prefer to copy `source...` part there, and not to use `source` prefix in the identity descriptor.

trieloff:
> But we also need to be able to refer to the descriptor schema directly, because we will be using it in our APIs that allow users to interactively create relationships across schemas.

Ok, I can see how you could even compile an update Schema out of a base schema and a number of descriptor files.

schemas/common/relationshipdescriptor.schema.json
kstreeter: A descriptor at minimum needs to say what schema it is a descriptor for.  Why would we have a different way to do this for each descriptor type?

schemas/common/identitydescriptor.schema.json
kstreeter: good catch, will fix

schemas/common/updatepolicydescriptor.schema.json
kstreeter: I went back/forth on whether these need to be extensible (and therefore a URI) or not.  I would be good with just making them strings.

schemas/common/relationshipdescriptor.schema.json
dmitry-shkolnik: But the relationship descriptor belongs to two schemas, expecially in many-to-many case. But that's not that important.

schemas/common/updatepolicydescriptor.schema.json
dmitry-shkolnik: No, I meant the value "Merge:" (should be "Merge"), not the property name.

kstreeter: Got it, will fix thanks Dmitry

kstreeter:
@trieloff, @dmitry-shkolnik I think I have all the changes in based on feedback that I received.  We've got several teams starting to implement descriptors and relationships, so I'd like to get this in soon so folks have the latest.  Take a look as soon as you can, thanks!

CONTRIBUTING.md
```diff
@@ -351,6 +351,226 @@ The third schema is `third.schema.json`, it extends both `second`, and transitiv
...
+```json
+{
+  "@id": "https://example.com/descriptors/1",
+  "@type": "xdm:descriptorOneToMany",
+  "xdm:sourceSchema": "https://ns.adobe.com/xdm/example/parent",
+  "xdm:sourcePropery": "@id",
+  "xdm:destSchema": "https://ns.adobe.com/xdm/example/child",
+  "xdm:destProperty": "xdm:parent"
+}
+```
```
dmitry-shkolnik: This example shows a classical one-to-many relationship, where an instance of "many" side has a property (foreign key) with a value being a primary key of an instance of "one" side. We need to differentiate it with the case where want to have on "one" side an array of references to "many" side. Should we have two separate cases for "oneToMany" and "manyToOne"?

kstreeter: Hi @dmitry-shkolnik, I'm not sure I completely follow.  Are you describing a case where 1:M and "M:1" aren't really just reciprocal,  but need to have distinct semantics?  Can you give a specific example of that?

dmitry-shkolnik: I am talking about the case, where it is more convenient to have an array of pointers to the children in the parent than pointer to parent in each children. In that case, when the relationship  between two instances is created, we'd like to add @id of a child to array in parent, instead of setting `parent` property in the child. 

This is similar to aggregation (when an object has an array of children), but if we decide to store children separately, we cannot express that with the descriptors.

As an examples: each instance of the offer schema (which we are working on) has an array of representations (how the offer may look like on different channels /  placements). Right now representations are fully embedded into offer. But if we decide to make them a separate first-class citizens with their own id, we''ll have to get rid of array in offer and add parent property to the representation, which is an old-fashion rdbms way of modeling.

kstreeter: Yes, that is definitely a case we haven't considered.  In general, the semantics of the relationship in that case don't change: you still have a 1:M relationship from parent to child.  But the relationship descriptor (as proposed) can't be applied to the denormalized form in your example.  I agree it would be good to be able to handle that, I'll need to think about how to approach it.

CONTRIBUTING.md
```diff
@@ -351,6 +351,226 @@ The third schema is `third.schema.json`, it extends both `second`, and transitiv
 }
 ```
 
+### Schema Descriptors
+
+Schema descriptors are an extensible mechanism for providing additional metadata about an XDM schema. For example, schema descriptors can be used to define relationships between schemas or to annotate schema properties with additional metadata. Schema descriptors may be used when certain properties of a schema are not static (which could usually be described in the schema directly) but may vary from usage to usage.
+
+Details on using and defining schema descriptors may be found in [SCHEMA_DESCRIPTORS.md](./SCHEMA_DESCRIPTORS.md)
```
trieloff: We might want to move that into the `docs` folder (and give it a lowercase name)

CONTRIBUTING.md
```diff
@@ -351,6 +351,226 @@ The third schema is `third.schema.json`, it extends both `second`, and transitiv
...
+
+Schema descriptors are extensible, and new descriptors may be creating by defining a new URI value and using it in
```
trieloff: may be created.

CONTRIBUTING.md
```diff
@@ -351,6 +351,226 @@ The third schema is `third.schema.json`, it extends both `second`, and transitiv
...
+* `xdm:updateTimeSeries`: the data is time series data, and the new object should be logged/collected without changing any existing data
+
+Update policies are defined using the UpdatePolicyDescriptor schema.
+
+#### Other Supported Schema Descriptors
+
+A number of additional schema descriptors are defined by XDM:
+
+* `xdm:descriptorIdentity`: allows a property in a schema to be used as an identity, even if it does not conform to the Identity schema.
+* `xdm:descriptorPrimaryKey`: allows a property other than `@id` to be flagged as the primary key for a schema
```
trieloff: It's unclear what the difference to the `descriptorIdentity` above is. Or it's unclear what the `descriptorIdentity` is in the first place.

CONTRIBUTING.md
```diff
@@ -351,6 +351,226 @@ The third schema is `third.schema.json`, it extends both `second`, and transitiv
...
+* `xdm:descriptorInstantiable`: allows a schema to be flagged as 'instantiable', which may be used to differentiate schemas that define primary business objects versus supporting schemas intended to be embedded in another schema.
```
trieloff: This seems to be simply the negation of `meta:abstract`, so I'm not sure we need it.

CONTRIBUTING.md
```diff
@@ -351,6 +351,226 @@ The third schema is `third.schema.json`, it extends both `second`, and transitiv
...
+
+```json
+{
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "$id": "https://ns.adobe.com/xdm/example/child",
+  "title": "Child",
+  "type": "object",
+  "properties": {
+    "@id": { "type": "string" },
+    "xdm:parent": { "type": "string" }
```
trieloff: I would include `format: uri`, as the `xdm:parent` property can have only values that would be valid `@id`s in `parent.schema.json`

CONTRIBUTING.md
```diff
@@ -351,6 +351,226 @@ The third schema is `third.schema.json`, it extends both `second`, and transitiv
...
+  }
+}
```
trieloff: What seems to be missing from the child schema is an equivalent descriptor that makes it clear that `xdm:parent` is not just a string (and not just a URL), but a URL pointing to an instance of `child.schema.json`, as identified by the `@id` property.

This inversion is making it hard from two perspectives:

1) as a reader of the schema, I have no indication when reading `child.schema.json` that `xdm:parent` is somewhat "special"
2) when generating documentation, I will have to load all known schemas into memory, and then do a look up for every single property to decide if there might be a schema descriptor that is pointing to the schema at hand.

CONTRIBUTING.md
```diff
@@ -351,6 +351,226 @@ The third schema is `third.schema.json`, it extends both `second`, and transitiv
...
+  "xdm:namespace": "https://id-server.adobe.com/1234",
+  "xdm:property": "code"
+}
+```
+
+The descriptor signals the namespace the ID is managed under (in this case, a fictitious service at id-server.adobe.com), and also signals that the value is a "code", meaning it is the externally managed handle for some ID managed by the namespace.
+
+##### Example Primary Key Descriptor
+
+We have a schema that describes a sales order taken from an external sales managemenbt system. As this schema is directly transcribed from the external system's data schema, it does not follow the XDM best practice of using @id as the primary key:
```
trieloff: typo: management.
backticks around `@id`

CONTRIBUTING.md
trieloff: After reading `SCHEMA_DESCRIPTORS.md` in full, some additional comments:

- definitely move it into `docs` and make it an official part of the XDM documentation
- remove duplicative content from `CONTRIBUTING.md`, just point out that there are schema descriptors, that they can be defined within the schema and outside it, recommend to have schema descriptors for core XDM schemas inside the schema (not standalone)
- For each schema descriptor, create a `description.md` file with examples and usage instructions, so that we can have discussions about the semantics and direction of schemas like the one-to-many descriptor there (because they don't apply to the framework in general, but to the specific descriptor)
- shorten the examples in `SCHEMA_DESCRIPTORS.md` (and remove them entirely from `CONTRIBUTING.md`), as they can be handled better and with more detail in the reference documentation.

schemas/common/itemselector.schema.json
```diff
@@ -0,0 +1,75 @@
+{
+  "meta:license": [
+    "Copyright 2017 Adobe Systems Incorporated. All rights reserved.",
```
trieloff: 2018

schemas/common/relationshipdescriptor.schema.json
```diff
@@ -0,0 +1,54 @@
+{
+  "meta:license": [
+    "Copyright 2017 Adobe Systems Incorporated. All rights reserved.",
```
trieloff: 2018

schemas/common/updatepolicydescriptor.schema.json
```diff
@@ -0,0 +1,51 @@
+{
+  "meta:license": [
+    "Copyright 2017 Adobe Systems Incorporated. All rights reserved.",
```
trieloff: 2018

trieloff:
@asteed I'd like to get your comments, too. You will be implementing a lot of this for our tooling.

kstreeter:
@trieloff, I think I had a cut/paste error that left some old documentation in CONTRIBUTING.  I have fixed this, and moved things around as you have suggested.  Please let me know if this resolves some of your concerns.

lrosenthol:
@kstreeter What's the status on this?  Do we want to wrap this up??

schemas/common/descriptors/relationshipdescriptor.schema.json
```diff
@@ -0,0 +1,54 @@
...
+        },
+        "xdm:destinationItem": {
+          "title": "Destination Item",
+          "description":
+            "When present, the selector used to match a specific item in the array pointed to by destProperty",
+          "$ref":
+            "https://ns.adobe.com/xdm/common/descriptors/itemselector#/definitions/selector"
+        }
+      },
+      "required": ["xdm:destinationSchema"]
```
jaganperi: It'll be good for the source schema related properties to also be part of relationshipDescriptor schema.json rather than in schemadescriptor.schema.json

Is there a reason for splitting source schema/destination schema properties across schemadescriptor.schema.json and relationshipDescriptor.schema.json?

cfraser: SchemaDescriptor is the base for RelationshipDescriptor. As such it is inheriting the source properties from that schema definition. We try to encourage reuse as much as possible in XDM so this is the right approach.

CC: @kstreeter

kstreeter:
The contents of this PR have either been merged in via other PRs or superceded by later PRs, so I am going to close.

PR #156, opened 2018-03-05
Added Adobe Target Experience Event extensions and End User ID (Plat 14128) 

cdegroot-adobe:
Added Adobe Target Experience Event extensions and End User ID (Plat 14128) 
This adds application specific fields that are only for Adobe Targets use except for the tntid which is a general identity. Updated relevant examples


cmathis:
Is there a reason you needed to create a separate pagedetails schema that you reference and not just add those two fields to the main extension schema. 

https://ns.adobe.com/experience/target/pagedetails/pageid

https://ns.adobe.com/experience/target/pagedetails/pagescore

Also, because "pageID" is part of a URI, it needs to be lower case - "pageid". Seems like the only way to use an upper case is if that field is associated with a prefix. If you need it to be camel case in another view (XED for example), then you will need to use a separator like page_id or something.

jwen-adobe:
@trieloff @kstreeter

cdegroot-adobe:
@cmathis  I merged in the pageDetails and added the extends metadata. Can you please see if this is now that way you would expect.

Thanks C.

cmathis:
@cdegroot-adobe  - Reminder to move this schema under the /extensions location and not under /schema/extensions

cdegroot-adobe:
@jwen-adobe I have moved these to the extensions folder and updated the @id. Can you please review to make sure it looks good. Thanks C.

schemas/context/enduserids.example.1.json
```diff
@@ -2,8 +2,8 @@
   "https://ns.adobe.com/experience/mcid": {
     "@id": "https://ns.adobe.com/entities/identity/mcid123",
     "xdm:namespace": {
-      "@id": "https://ns.adobe.com/entities/namespace/11111",
-      "xdm:code": "AA11111"
+      "@id": "https://ns.adobe.com/entities/namespace/AVID",
```
trieloff: Is this a real customer name? Would it make sense to replace it with a non-protected name?

schemas/context/enduserids.example.1.json
```diff
@@ -2,8 +2,8 @@
...
+      "xdm:code": "4"
```
trieloff: Is a longer value more realistic?

schemas/context/enduserids.schema.json
```diff
@@ -22,6 +22,11 @@
       "title": "Adobe Analytics Cloud End User Ids",
       "$ref": "https://ns.adobe.com/xdm/context/identity",
       "description": "Adobe Analytics Cloud extension to End User Ids."
+    },
+    "https://ns.adobe.com/experience/tntid": {
```
trieloff: Why do we use `tntid` instead of `target`? `tnt` either refers to an explosive or a product name that we stopped using more than five years ago.

cdegroot-adobe: This is in a extension for a application and that is the name of the field that is currently in use and will remain in use until it fully migrates over to mcid. So in the context of that extension it is 100% accurate, though as you rightly point out it has suffered from naming evolution.

schemas/context/enduserids.example.1.json
cdegroot-adobe: I had a good review with the Identity team right after this which gave me good information to help make this sort of example more illustrative. I will update accordingly.

schemas/context/enduserids.example.1.json
cdegroot-adobe: AVID is the registered name in our "namespace issuing authority" and stands for "Analytics Visitor Identity" It is used like this on almost all report suites that have migrated to use the Visitor Identity Service. So it is accurate.

trieloff: Ok.

schemas/context/enduserids.example.1.json
trieloff: Ok, then that's the last remaining comment that needs to be addressed. Please also resolve the conflict.

schemas/context/enduserids.schema.json
kstreeter: if that's the common name used by the application we should keep it.  As @cdegroot-adobe it is appropriately namespaced.

PR #158, opened 2018-03-07
How to restrict string properties

trieloff:
As discussed in #151


kstreeter:
This looks like a good addition to me.

Issue #159, opened 2018-03-07
Bug fix of analytics-extensions


jwen-adobe:
## What are the schemas that are affected by the issue
analytics-extensions

## What are examples of products that are impacted by the issue
platform xdm definition

PR #160, opened 2018-03-07
Fix the $id by adding adobe to be consistent with the folder path

jwen-adobe:
This PR links to the issue #159 

@trieloff @kstreeter @cmathis

Fix the $id of the schema which is missing "adobe".

### DIFF
--- a/extensions/adobe/experience/analytics/analytics-extensions.schema.json
+++ b/extensions/adobe/experience/analytics/analytics-extensions.schema.json
@@ -1,5 +1,5 @@
 {
-  "$id": "https://ns.adobe.com/xdm-extensions/experience/analytics/analytics-extensions",
+  "$id": "https://ns.adobe.com/xdm-extensions/adobe/experience/analytics/analytics-extensions",
   "$schema": "http://json-schema.org/draft-06/schema#",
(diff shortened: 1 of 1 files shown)


cmathis:
The id's do not need to follow the same physical directly structure do they?

@kstreeter  - Given these are considered "Experience Cloud Schemas" as [described in the document here](https://wiki.corp.adobe.com/display/XDM/XDM+Contribution+Process+for+Experience+Cloud?focusedCommentId=1448823232#comment-1448823232) they should have an id that starts with https://ns.adobe.com/experience, but as you can see from the other extension schemas they do not do this. There is some inconsistency on what the @id should actually be for these.

jwen-adobe:
It does not need to be, but we will need to change the mapping in someway that the schema files can be found as the $ref is using the $id. The current converter has following mappings to search standard and extension schema files:

https://ns.adobe/com/xdm --> schemas folder
https://ns.adobe.com/xdm-extensions --> extensions folder
etc..
 
so if "adobe" is not part of $id, the schema file can not be found from $ref then unless we change the mapping.

cmathis:
As of now it appears that the $id follows the folder path so let's continue with that. This change to have the adobe name in the path seems right.

Issue #163, opened 2018-03-08
Change the URIs for analytics-extension event fields


jwen-adobe:
## What are the schemas that are affected by the issue
analytics-extension

## What are examples of products that are impacted by the issue
DMa platform xdm definition

PR #164, opened 2018-03-09
Change the URI for the event fields

jwen-adobe:
This PR links to the issue #163 

@trieloff @kstreeter @cmathis

This PR changes the analytics-extensions event fields' URIs, so the analytics events fields have the same parent namespace like standard metrics (requested by analytics).

### DIFF
--- a/extensions/adobe/experience/analytics/analytics-extensions.schema.json
+++ b/extensions/adobe/experience/analytics/analytics-extensions.schema.json
@@ -2670,5002 +2670,5002 @@
       "type": "string",
       "description": "The original delimiter used to generate the list."
     },
-    "https://ns.adobe.com/experience/analytics/event1to100/event1": {
+    "https://ns.adobe.com/xdm/data/metrics/_experience/analytics/event1to100/event1": {
       "title": "event1",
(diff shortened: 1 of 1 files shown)


kstreeter:
hey @jwen-adobe, why the underscore in "_experience"?

jwen-adobe:
@kstreeter if we use experience rather than _experience, the namespace will be not in sync with other extension fields like "_experience.analytics.customdimensions.evars". For example by use experience, the event field will be "experience.analytics.event1to100.event1" under _metrics object.

cmathis:
This was only way we could get those metric extensions to line up with existing standard metric fields when the extension is applied.

This was the goal:
_metric
    commerce
         purchase
   _experience
        analytics
            event1

Without this it ended up being
_metric
    commerce
         purchase
_experience
     analytics
         event1

This get's into the issue of being able to add extensions to a sub-schema of the main schema you are extending. If you can think of a better way to do this would be fine.

cmathis:
Hold off on merging this PR until we can figure out the better way to do this as being discussed in email.

jwen-adobe:
Close this as adding namespaces to existing standard field's uris is not allowed.

Issue #157, opened 2018-03-07
Require Property Types in meta.schema.json #151


trieloff:
#151 is adding this as a written guideline, but we can and should enforce this in the meta schema.

## What are the schemas that are affected by the issue

- all schemas

## What are examples of products that are impacted by the issue

- all products

cdegroot-adobe:
@trieloff  - do we still need this issue at this stage? I think we have done what we can.

trieloff:
Yes, we still need this.

PR #172, opened 2018-03-13
Added meta:xdmType to hyper schema

jwen-adobe:
This PR links to the issue #157 

@trieloff @kstreeter @cmathis

This is the first draft of the meta:xdmType added to hyper schema, please review and make recommendation as needed. 

It is very simple at this moment as it can be quite complicated if want to cover all the dbs and programming languages, also we do not have use cases for those yet.

### DIFF
--- a/meta.schema.json
+++ b/meta.schema.json
@@ -71,6 +71,33 @@
         }
       ]
     },
+    "meta:xdmType": {
+      "string": {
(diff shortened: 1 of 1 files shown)


jwen-adobe:
Close this as it will be in document only rather than hyper schema for now.

PR #178, opened 2018-03-19
[trivial] fixing broken anchor-link

actinium15:


trieloff:
All the docs are machine generated, so this should go into adobe/jsonschema2md

Issue #170, opened 2018-03-13
Sort out channel, locations, content


trieloff:
As discussed with @cdegroot-adobe 

- [ ] introduce a new concept: Channel-Location (a "box" in the virtual world where content can go)
- [ ] rename Location Context to Place Context (to avoid confusion with the above)
- [ ] make ExperienceEvent refer to Channel-Location by ID (optionally by name)
- [ ] make ExperienceEvent refer to Content by ID (the thing that goes into the box)

cdegroot-adobe:
@trieloff we are not starting to get traction in the areas of 1,2,4 of the open items. I think 3 is out of scope and not really related.

We need to move away from the term 'channel' here as it is conflating an already highly complex term.

I will keep this issue up to date as the AEM and AdCloud teams drill into these areas.

trieloff:
I think the #245 PR will force us to sort this out. We need to have congruence between the planning side (#245) and the tracking side (Experience Event, #170) – and later on, the reporting side.

PR #186, opened 2018-03-20
WIP: Example of consistent modeling of channelLocation and referral

trieloff:
This relates to #143 #142 #170

Right now, it's only the example, so that we can get clarity of what it should look like. Once we have agreement, I can update the schema accordingly.

It would also pose an opportunity for streamlining "search", which has been getting special treatment so far.

@kstreeter let me know what you think.


schemas/context/experienceevent.example.1.json
```diff
@@ -85,7 +85,31 @@
       "schema:longitude": 139.73237
     }
   },
-  "xdm:marketing": {
-    "xdm:trackingCode": "marketingcampaign111"
+  "xdm:referral": {
+    "xdm:trackingCode": "KKQWX",
```
cdegroot-adobe: Moving the `xdm:trackingCode` to the proposed  `xdm:referral` suggests that the current page/email/mobile state is not able to set the trackingCode.

trieloff: Interesting point. I would see the `trackingCode` as something that the referring experience sets, opposed to properties like `currentCampaign`, `selectedOffer`, `relevantExperience`, etc. that can be resolved in the current context. 

We can keep it in `xdm:marketing` if it makes more sense.

trieloff: I've moved it back in 1a49af8 @cdegroot-adobe

PR #196, opened 2018-03-21
Changing DL for contacting Adobe XDM Core WG

ktukker:
Changing DL for contacting Adobe XDM Core WG

Please link to the issue #…


lrosenthol:
@ktukker - can you do this against master so we don't accidently screw up the merge?

PR #197, opened 2018-03-21
Improved ExperienceEvent example and corrected datasource schema xdm:name

cdegroot-adobe:
1. Improve experienceevent.example.1.json:
- Added metric
- Updated endUserIDs
- Updated datasource name->code
- improved the order pricing

2. Updated datasource.schema.json
- Renamed xdm:name to xdm:code to better reflect the meaning of the parameter.


cdegroot-adobe:
@trieloff this is to generally update the experience event example and do one small correction to data source that seems to be an error the crept in at some point.

Hopefully this is an easy merge.

kstreeter:
Hey @cdegroot-adobe, I am seeing several additional changes in this PR that don't appear to be around the fixes you mentioned.  Can you check if all these are intended?

trieloff:
The PR itself seems to be fine, but I'm not sure how `xdm:name` to `xdm:code` clarifies things. Last time we discussed this, the property values were not guaranteed to be unique and unambiguous (that's why we have a proper `@id`) and were acting more like a human-readable name, hence `name`.

cdegroot-adobe:
@trieloff "code" has a specific meaning that is different to "id" and "name". A "code" is an alternate reference to an entity that has an "id", and may have a "name". The "code" allows the entity to be referenced more readily in some other domain. Examples of codes are "trackingCode" which is a way of providing a short http get parameter safe alternate name for a Campaign. 
In the case of datasource and namespace they are human friendly references that are not as freeform as a "name". There can be duplicate codes for the same id in some cases, and thus for the same "name" as a general concept. In future we may add a "name" which is fully human readable (white space etc).

cdegroot-adobe:
@kstreeter The only change I recall that was outside the "improvement" of the example was the "name"->"code". did you see others?

kstreeter:
@cdegroot-adobe , looks like we just needed to update latest from master.  I am good with this, will merge.

schemas/data/datasource.schema.json
```diff
@@ -21,10 +21,11 @@
...
             "The unique ID of this data source. This would be provided by the individual or system that created the Datasource.\n"
         },
-        "xdm:name": {
-          "title": "Name",
+        "xdm:code": {
+          "title": "Code",
           "type": "string",
-          "description": "Human readable name of the data source. "
+          "description":
+            "The code is a shortcut to the full @id and at least one of the code or @id can be used. Sometimes, this code refered to as the data source integration code."
```
trieloff: @kstreeter please take care of the typo and make sure that @cdegroot-adobe's explanation of what a `code` is is reflected in the documentation. 

Also, @cdegroot-adobe's explanation made it sound like `code` would exist in addition to `name`, the PR you merged replaces it.

Issue #195, opened 2018-03-20
Change "http://" namespace urls to use "https://"


dgeorge:
In the comments for https://github.com/adobe/xdm/issues/188, @trieloff posted a complete list of all namespace URLs referenced in XDM schemas.

The list includes the following XDM URLs, which begin with  "http://" instead of "https://"

```
http://ns.adobe.com/xdm/IMSOrg
http://ns.adobe.com/xdm/IMSUser
http://ns.adobe.com/xdm/aem/instance
http://ns.adobe.com/xdm/aem/user
http://ns.adobe.com/xdm/asset
http://ns.adobe.com/xdm/assets/asset
http://ns.adobe.com/xdm/assets/asset#
http://ns.adobe.com/xdm/event/created
http://ns.adobe.com/xdm/event/published
http://ns.adobe.com/xdm/event/unpublished
http://ns.adobe.com/xdm/event/updated
http://ns.adobe.com/xdm/page
http://ns.adobe.com/xdm/services
```

I lack all of the context behind those URL choices.  On the surface, though, it would seem that they should be changed to use "https://".  Also, the following already-existing URLs may be redundant with entries in the above list:

```
https://ns.adobe.com/xdm-extensions/ims/organization
https://ns.adobe.com/xdm-extensions/ims/user OR https://ns.adobe.com/xdm/base/imsuser#
https://ns.adobe.com/xdm/aem/user
https://ns.adobe.com/xdm/assets/asset
https://ns.adobe.com/xdm/assets/asset#/definitions/asset
https://ns.adobe.com/xdm/common/event/created
https://ns.adobe.com/xdm/common/event/published
https://ns.adobe.com/xdm/common/event/updated
https://ns.adobe.com/xdm/common/page OR https://ns.adobe.com/xdm/content/page-component
```

trieloff:
@dgeorge thanks for having a look at this.

dgeorge:
@asteed If you'd prefer to refactor this into multiple smaller issues, that's totally fine with me.

trieloff:
Yes, it's needed for 0.9, as it is a breaking change.

asteed:
The following is a list of changes made to the items raised by @dgeorge 

| Original | Updated |
| --- | --- |
| `http://ns.adobe.com/xdm/IMSOrg` | `https://ns.adobe.com/xdm-extensions/ims/organization` |
[...]

PR #200, opened 2018-03-23
Rename xdm namespaces from 'http' to 'https'

asteed:
Addresses #195
Rename `http://ns.adobe.com/xdm => https://ns.adobe.com/xdm`


kstreeter:
Thanks for making this change @asteed, I don't see any issues

PR #202, opened 2018-03-26
Adding generic and Campaign specific extensions to Profile and Address schema.

aarjain:


schemas/context/profile.schema.json
```diff
@@ -65,6 +65,11 @@
           "$ref": "https://ns.adobe.com/xdm/context/phonenumber",
           "description": "Mobile phone number.\n"
         },
+        "xdm:faxPhone": {
```
lrosenthol: This should not be a top level element in xdm - it should be inside something else, say Address or Phone.  Looks like there are a bunch of these in this particular schema.  @asteed can you resole?

schemas/context/profile.schema.json
```diff
@@ -86,6 +91,24 @@
           "description":
             "The unit ID within the organization owning the profile. This ID can be used to reference the organization details maintained in another dataset."
         },
+        "xdm:preferredLanguage": {
```
lrosenthol: there is no information about what format this takes?  BPC-47?  Other?

schemas/context/profile.schema.json
```diff
@@ -86,6 +91,24 @@
...
+          "title": "Preferred Language",
+          "type": "string",
+          "description":
+            "Preferred Language"
+        },
+        "xdm:thumbnail": {
```
lrosenthol: thumbnail is something completely different.  This needs a name like "profileLink" to correct describe it.  (also it doesn't belong in the top level namespace)

schemas/context/profile.schema.json
```diff
@@ -86,6 +91,24 @@
...
+          "title": "Thumbnail",
+          "type": "string",
+          "description":
+            "Link to profile's picture"
+        },
+        "xdm:emailFormat": {
```
lrosenthol: format in what format??  Is this a regex?  What is it??

schemas/context/profile.schema.json
aarjain: Added format information (ISO-639-1 language codes) in latest commit.

schemas/context/profile.schema.json
aarjain: Changed the name to "profilePictureLink" in latest commit.

schemas/context/profile.schema.json
aarjain: Added format details (enum) in latest commit.

schemas/context/profile.schema.json
```diff
@@ -94,20 +94,23 @@
         "xdm:preferredLanguage": {
           "title": "Preferred Language",
           "type": "string",
-          "description":
-            "Preferred Language"
+          "description": "Preferred language based on ISO-639-1 format"
```
trieloff: A couple of questions:
- why not BCP 47, which is used in all other paces in XDM
- if you are using a standard, link to it using `[ISO-639-1](http://blabla)`
- provide examples, most people don't know from the top of their head what a language code looks like
- provide a pattern (if you are using BCP 47, there's one in `dc:language` to steal

schemas/context/profile.schema.json
```diff
@@ -94,20 +94,23 @@
         "xdm:preferredLanguage": {
           "title": "Preferred Language",
           "type": "string",
```
trieloff: Do we allow keeping a language preference order like any browser or operating system? My preference order would be: `["de_DE", "de", "en_US", "en"]`, i.e. "real" German first, then any form of German, then US English, then any form of English?

schemas/context/profile.schema.json
```diff
@@ -94,20 +94,23 @@
...
           "title": "Email Format",
           "type": "string",
-          "description":
-            "Format of email relevant to the user"
+          "description": "Email format relevant to the user. This can be HTML/Text/Unknown",
+          "default": "unknown",
+          "meta:enum": {
+            "html": "HTML",
+            "text": "Text",
+            "unknown": "Unknown"
```
trieloff: Do we really need to save an "unknown" value? How does this compare against the value just being `null`?

schemas/context/profile.schema.json
```diff
@@ -94,20 +94,23 @@
...
-          "description":
-            "Link to profile's picture"
+          "description": "Link to profile's picture"
         },
         "xdm:emailFormat": {
           "title": "Email Format",
           "type": "string",
-          "description":
-            "Format of email relevant to the user"
+          "description": "Email format relevant to the user. This can be HTML/Text/Unknown",
```
trieloff: Is this a user preference or an inferred setting? "Relevant to the user" sounds inferred, but together with the other stuff in this PR, it looks like a preference.

schemas/context/profile.schema.json
```diff
@@ -94,20 +94,23 @@
...
+          "default": "unknown",
+          "meta:enum": {
+            "html": "HTML",
```
trieloff: Mention rich text in description.

schemas/context/profile.schema.json
```diff
@@ -94,20 +94,23 @@
...
+            "text": "Text",
```
trieloff: Mention plain text in description.

schemas/context/profile.schema.json
lrosenthol: Only -1 codes?  That's very old school.  Are you sure that is what is being used?  Why arent' we using the modern BCP-47 approach to language specification?

schemas/context/profile.schema.json
lrosenthol: I would actually call it plaintext instead of just text

schemas/context/profile.schema.json
lrosenthol: +100 - agreed on all points here.

schemas/context/profile.schema.json
aarjain: Added a schema for language, making it flexible for applications to specify language codes of their choice.

schemas/context/language.schema.json
```diff
@@ -0,0 +1,42 @@
+{
```
lrosenthol: This something that should be part of common - since if we are going to standardize this, it should be across the board for all schemas.

trieloff:
@aarjain can you resolve the conflicts? I can then take another look.

schemas/context/language.schema.json
```diff
@@ -0,0 +1,42 @@
...
+    "The system of communication used by a particular person, country or community.",
+  "definitions": {
+    "language": {
+      "properties": {
+        "xdm:format": {
+          "title": "Format Identifier",
+          "type": "string",
+          "description":
+            "Standardized nomenclature used to classify the language",
+          "meta:enum": {
```
trieloff: I'm not a fan of this at all. We should use BCP-47 across the board, except in cases where we have to maintain compatibility with XMP, which predates BCP-47.

schemas/context/profile.schema.json
trieloff: @lrosenthol Adding `faxPhone` is consistent with the other properties, so I would not reject it because of this.

But: I agree that it is an area that will lead to channel-creep (is `whatsAppPhone` next?). If you agree, let's file an issue to have channel-encoded contact coordinates and opt-in/out instead.

schemas/context/language.schema.json
aarjain: @trieloff , @lrosenthol : We follow ISO-639-1 codes in Campaign for language specification and hence need this functionality. Please comment on how can we incorporate this in the best way.

trieloff: Within XDM: use BCP-47
In the Adobe Campaign extensions of XDM: use whatever you want (although my recommendation would be BCP-47, as it makes things easier)
In the Adobe Campaign implementation: use whatever you want and map it to BCP-47.

extensions/adobe/experience/campaign/profile.schema.json
```diff
@@ -0,0 +1,31 @@
...
+    "https://ns.adobe.com/xdm-extensions/adobe/experience/campaign/experienceevent",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "ExperienceEvent Extension for Adobe Campaign",
+  "type": "object",
+  "meta:extends": ["https://ns.adobe.com/xdm/context/experienceevent"],
+  "description": "All ExperienceEvent properties specific to Adobe Campaign.",
+  "definitions": {
+    "profile": {
+      "properties": {
+        "https://ns.adobe.com/experience/campaign/profile/timeZone": {
```
saurabhere: @cdegroot-adobe @trieloff Should this be part of the core Profile?
@aarjain `timezone` and a better description will help.

schemas/context/language.schema.json
saurabhere: While I was pushing for supporting ISO-636-1 earlier as well, I agree with this point now.
@aarjain We should use a mapper to send data in the BCP-47 format. We finally want to send data in a "standard" format.

extensions/adobe/experience/campaign/profile.schema.json
aarjain: Added changes to incorporate timeZone into core profile XDM.

schemas/context/language.schema.json
aarjain: Made modifications to use BCP 47 as the standard for specifying language codes.

schemas/context/profile.schema.json
aarjain: @trieloff @lrosenthol : Can we have a resolution of the above xdm field from your side?

extensions/adobe/experience/campaign/profile.schema.json
aarjain: @cdegroot-adobe  : Please check the addition of "timeZone" in standard profile in the new commit and share your comments on the same.

schemas/context/profile.schema.json
saurabhere: @trieloff @lrosenthol @aarjain I have created https://github.com/adobe/xdm/issues/269 for this. We can take this discussion there and let this be unblocked for now. WDYT?

extensions/adobe/experience/campaign/address.schema.json
```diff
@@ -0,0 +1,37 @@
...
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id":
+    "https://ns.adobe.com/xdm-extensions/adobe/experience/campaign/experienceevent",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "ExperienceEvent Extension for Adobe Campaign",
+  "type": "object",
+  "meta:extends": ["https://ns.adobe.com/xdm/context/experienceevent"],
+  "description": "All ExperienceEvent properties specific to Adobe Campaign.",
```
saurabhere: @aarjain Please fix the description.

extensions/adobe/experience/campaign/address.schema.json
```diff
@@ -0,0 +1,37 @@
...
+  "definitions": {
+    "address": {
+      "properties": {
+        "https://ns.adobe.com/experience/campaign/profile/addressErrorCount": {
```
saurabhere: should be `https://ns.adobe.com/experience/campaign/address/addressErrorCount`

extensions/adobe/experience/campaign/address.schema.json
```diff
@@ -0,0 +1,37 @@
...
+          "title": "Address Error Count",
+          "type": "integer",
+          "description":
+            "Number of consecutive errors when sending to this address."
+        },
+        "https://ns.adobe.com/experience/campaign/profile/addressQuality": {
```
saurabhere: should be `https://ns.adobe.com/experience/campaign/address/addressQuality`

extensions/adobe/experience/campaign/address.schema.json
```diff
@@ -0,0 +1,37 @@
...
+        "https://ns.adobe.com/experience/campaign/address/addressErrorCount": {
+          "title": "Address Error Count",
+          "type": "integer",
+          "description":
+            "Number of consecutive errors when sending to this address."
+        },
+        "https://ns.adobe.com/experience/campaign/address/addressQuality": {
+          "title": "Address Quality",
+          "type": "integer",
+          "description": "Address quality rating."
```
trieloff: Needs a better description. Just repeating the title is not enough.

extensions/adobe/experience/campaign/address.schema.json
```diff
@@ -0,0 +1,37 @@
...
+      "properties": {
+        "https://ns.adobe.com/experience/campaign/address/addressErrorCount": {
+          "title": "Address Error Count",
+          "type": "integer",
+          "description":
+            "Number of consecutive errors when sending to this address."
+        },
+        "https://ns.adobe.com/experience/campaign/address/addressQuality": {
+          "title": "Address Quality",
+          "type": "integer",
```
trieloff: You should consider making this a `number` between `0` and `1` – we use this in multiple places. If it is an integer, please specify what the values mean. Is more better, is less better, etc.

schemas/context/profile.schema.json
trieloff: Sounds good, @saurabhere.

@lrosenthol – do you agree that we can merge even with #269 unresolved (which is an XDM-wide concern)

schemas/context/profile.schema.json
trieloff: This has been resolved, we are using BCP-47 now.

schemas/context/profile.schema.json
```diff
@@ -86,6 +91,35 @@
           "description":
             "The unit ID within the organization owning the profile. This ID can be used to reference the organization details maintained in another dataset."
         },
+        "xdm:preferredLanguage": {
+          "title": "Preferred Language",
+          "type": "string",
+          "pattern":
+            "^(((([A-Za-z]{2,3}(-([A-Za-z]{3}(-[A-Za-z]{3}){0,2}))?)|[A-Za-z]{4}|[A-Za-z]{5,8})(-([A-Za-z]{4}))?(-([A-Za-z]{2}|[0-9]{3}))?(-([A-Za-z0-9]{5,8}|[0-9][A-Za-z0-9]{3}))*(-([0-9A-WY-Za-wy-z](-[A-Za-z0-9]{2,8})+))*(-(x(-[A-Za-z0-9]{1,8})+))?)|(x(-[A-Za-z0-9]{1,8})+)|((en-GB-oed|i-ami|i-bnn|i-default|i-enochian|i-hak|i-klingon|i-lux|i-mingo|i-navajo|i-pwn|i-tao|i-tay|i-tsu|sgn-BE-FR|sgn-BE-NL|sgn-CH-DE)|(art-lojban|cel-gaulish|no-bok|no-nyn|zh-guoyu|zh-hakka|zh-min|zh-min-nan|zh-xiang)))$",
+          "description":
+            "Describes the preferred system of communication used by the profile. Language codes are expressed in BCP 47 format, for example `en-GB`."
```
trieloff: Use `examples` for examples. There's a dedicated property in JSON Schema

schemas/context/profile.schema.json
```diff
@@ -86,6 +91,35 @@
...
+        },
+        "xdm:timeZone": {
+          "title": "Time Zone",
+          "type": "string",
+          "description":
+            "Describes which time zone the profile is present in, most frequently/the time zone preferred by the profile."
```
trieloff: I missed this before. What's the format for time zones? If we don't have a defined format, I'd suggest using the IETF `tz` database: https://www.ietf.org/timezones/tzdb-2016i/tz-link.htm

We can probably build a `pattern` regex and *should* provide a couple of `examples`.

extensions/adobe/experience/campaign/address.schema.json
aarjain: Added enum values for the same

aarjain: @trieloff @lrosenthol : Incorporated review comments. Lets try to progress into merging this.

trieloff:
@aarjain issues #277 and #284 might have an impact on your PR. Please use `https://ns.adobe.com/experience/campaign/*` instead of `https://ns.adobe.com/experience/campaign/*/*` for all property names, i.e. treat `https://ns.adobe.com/experience/campaign/` as your namespace base (you get only one) and bear responsibility of having unambiguous names below that.

extensions/adobe/experience/campaign/address.schema.json
```diff
@@ -0,0 +1,45 @@
...
+          "description": "Address quality rating.",
+          "meta:enum": {
```
trieloff: If these are the only values you support, make the type a proper `enum` instead of an `integer`. Also: we typically use strings as enum keys instead of numbers, which is something we would do when moving this into the standard.

schemas/context/profile.schema.json
lrosenthol: yes @trieloff I am fine since we have the bigger issue identified separately and we agree not to "creep" in any more

aarjain:
@trieloff @lrosenthol: Incorporated URI changes according to  #277 and #284.
Made changes to make enum keys as strings instead of integer.
Lets move forward into merging this.

extensions/adobe/experience/analytics/experienceevent.schema.json
```diff
@@ -1640,405 +1639,405 @@
...
-        "https://ns.adobe.com/experience/analytics/custom-dimensions/list-props/prop74":{
-           "title":"prop74",
-           "description":"Custom property variable74",
-           "$ref":"https://ns.adobe.com/experience/analytics/listdetails"
-        },
-        "https://ns.adobe.com/experience/analytics/custom-dimensions/list-props/prop75":{
-           "title":"prop75",
-           "description":"Custom property variable75",
-           "$ref":"https://ns.adobe.com/experience/analytics/listdetails"
+        "https://ns.adobe.com/experience/analytics/custom-dimensions/list-props/prop1": {
```
lrosenthol: This whole custom property value thing bothers me!  It's not in line with the XDM model

lrosenthol:
Looking better, thanks @aarjain - but the custom property stuff bothers me.

aarjain:
@lrosenthol : The custom property commit was a result of incorrect merge of master. Reverted the same.
So we can drop that. Would you say we are good to merge this one?

lrosenthol:
yes, if we pull that out - then I am good to go.  @trieloff ?

aarjain:
@lrosenthol : Already reverted that. Thanks
@trieloff : Please share your review.

trieloff:
@lrosenthol good from my side. Please change your review when we are good to merge.

aarjain:
Great @lrosenthol , @trieloff .
Let's go ahead with merging this. Thanks

Issue #206, opened 2018-03-27
Add Campaign Extensions


saurabhere:
## What are the schemas that are affected by the issue
experienceEvent and profile schemas will need to be extended to accomodate for campaign's use-case

## What are examples of products that are impacted by the issue
platform xdm definition
adobe campaign

PR #209, opened 2018-03-29
Adding Campaign's extension of Experience Events.

saurabhere:
Fix Issue #206 
This extension is needed for campaign's experienceEvents to be represented in XDM.

Please review: @trieloff @kstreeter @cmathis @cdegroot-adobe


extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,259 @@
...
+          ],
+          "description": "Outbound IP address used to deliver the message."
+        },
+        "https://ns.adobe.com/experience/campaign/delivery/id": {
+          "title": "DeliveryID",
+          "type": "integer",
+          "description": "The campaign activity originating this message."
+        },
+        "https://ns.adobe.com/experience/campaign/delivery/channel": {
+          "title": "Communication Channel",
```
saurabhere: Should this be added to the core experienceEvent?

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,259 @@
...
+          "title": "Occurrence",
+          "type": "integer",
+          "description":
+            "URL occurrence index in the content (first is 1). When the same URL is available several times in the same content, its is necessary to track which occurrence of the URL has been clicked on.",
+          "default": 1,
+          "minimum": 1,
+          "maximum": 32767
+        },
+        "https://ns.adobe.com/experience/campaign/link/tags": {
+          "title": "Tags",
```
saurabhere: This looks too unstructured. Campaign's requirement is to have the following metadata properties to the URLs,
* Tracking URL Label
* Tracking URL Category
Should I add these to the core/extended weblink.schema ? Suggestions?

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,259 @@
...
+          "type": "integer",
+          "description":
+            "Size in bytes of the message.\n* For email, the `size` reflects the all MIME envelop of the message, including the encoded attachments if any.\n* For SMS, the `size` reflects the number of bytes necessary to deliver the message.\nIf that number is over 160 the message will be delivered over more than one SMS.",
+          "minimum": 0,
+          "maximum": 2147483647
+        },
+        "https://ns.adobe.com/experience/campaign/message/reason": {
+          "title": "Reason",
+          "type": "integer",
+          "meta:enum": {
```
lrosenthol: Why use an integer instead of a string?

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,259 @@
...
+            "The reason why the message could/would not be delivered.",
+          "default": "Undefined"
+        },
+        "https://ns.adobe.com/experience/campaign/message/reasonMessage": {
+          "title": "Reason Message",
+          "type": "string",
+          "description":
+            "Reason of message delivery failure when the status is `bounce` or `nonDeliverable`."
+        },
+        "https://ns.adobe.com/experience/campaign/message/ip": {
```
lrosenthol: shouldn't this be outboundIP, to be more descriptive?

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,259 @@
...
+          "title": "DeliveryID",
+          "type": "integer",
+          "description": "The campaign activity originating this message."
+        },
+        "https://ns.adobe.com/experience/campaign/delivery/channel": {
+          "title": "Communication Channel",
+          "type": "string",
+          "meta:enum": {
+            "https://ns.adobe.com/xdm/channels/email": "Email",
+            "https://ns.adobe.com/xdm/channels/sms": "Mobile (SMS)",
```
lrosenthol: enum values should *not* have spaces (or parens)

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,259 @@
...
+            "https://ns.adobe.com/xdm/channels/phone": "Phone",
+            "https://ns.adobe.com/xdm/channels/direct-mail": "Direct Mail",
```
lrosenthol: enum values should *not* have spaces

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,259 @@
...
+            "https://ns.adobe.com/xdm/channels/fax": "Fax",
+            "https://ns.adobe.com/xdm/channels/apns":
```
lrosenthol: enum values should *not* have spaces (or parens)

Also, is APNS different than ANS?

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,259 @@
...
+              "Apple Push Notification Service",
+            "https://ns.adobe.com/xdm/channels/gcm": "Google Cloud Messaging",
```
lrosenthol: enum values should *not* have spaces (or parens)

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,259 @@
...
+          "type": "boolean",
+          "description":
+            "Specify whether campaign activity is of type A/B testing.",
+          "default": false
+        },
+        "https://ns.adobe.com/experience/campaign/delivery/messagesClass": {
+          "title": "Message Class",
+          "type": "integer",
+          "meta:enum": {
+            "oneTime": "Marketing (One Time)",
```
lrosenthol: you say the type is an integer, but the enums here are strings.  Please rectify

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,259 @@
...
+          },
+          "description":
+            "Marketing classification of the messages sent through that delivery.\n",
+          "default": 0
+        },
+        "https://ns.adobe.com/experience/campaign/delivery/template": {
+          "title": "Template",
+          "type": "integer",
+          "description":
+            "The delivery template used to initialize this delivery.\n\nThe type the template used in Adobe Campaign can be identified using the `messageClass`.\n\n* `messageClass` = `oneTime`: the template is standard delivery template.\n* `messageClass` = `continuous`: the template is a recurring delivery.\n* `messageClass` = `transactional`: the template is a transactionnal message template.\n"
```
lrosenthol: you say the type is an integer, but the enums here are strings. Please rectify

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,259 @@
...
+          "type": "string",
+          "description": "Unique Identifier of the Link"
+        },
+        "https://ns.adobe.com/experience/campaign/link/description": {
+          "title": "Tracking URL Description",
+          "type": "string",
+          "description": "A human-friendly description of the Link."
+        },
+        "https://ns.adobe.com/experience/campaign/link/url": {
+          "title": "URL",
```
lrosenthol: If the *id* field is a URL, what is this URL and how do they differ?

extensions/adobe/experience/campaign/experienceevent.schema.json
lrosenthol: more importantly - why isn't this using the standard/core Tags model?

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,259 @@
...
+          "title": "Offers",
+          "type": "array",
+          "description": "Offers displayed in the message.",
+          "items": {
+            "$ref":
+              "https://ns.adobe.com/xdm-extensions/adobe/experience/campaign/offer"
+          }
+        },
+        "https://ns.adobe.com/experience/campaign/message/offerClicked": {
+          "title": "Offer Clicked",
```
lrosenthol: aren't you missing the type here?

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,259 @@
...
+          "$ref":
+            "https://ns.adobe.com/xdm-extensions/adobe/experience/campaign/offer",
+          "description":
+            "The offer and its location in the message for a click event related to an offer."
+        },
+        "https://ns.adobe.com/experience/campaign/message/externalId": {
+          "title": "External ID",
+          "type": "string",
+          "description":
+            "An ID generated by Voyager for a message to check later if the incoming experienceEvent was generated from this ID."
```
lrosenthol: I assume Voyager is the code name of something internal - please remove.

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,259 @@
...
+        },
+        "https://ns.adobe.com/experience/campaign/businessContext": {
+          "title": "Business Context",
+          "type": "string",
+          "description":
+            "Internal field for identifying the various actions/intents from Voyager."
```
lrosenthol: I assume Voyager is the code name of something internal - please remove.

extensions/adobe/experience/campaign/offer.schema.json
```diff
@@ -0,0 +1,62 @@
...
+  "type": "object",
+  "description":
+    "The Offer Placement represents the instance of an offer displayed in a message.",
+  "definitions": {
+    "offer": {
+      "properties": {
+        "https://ns.adobe.com/experience/campaign/offer/id": {
+          "title": "Offer ID",
+          "type": "integer",
+          "description": "Unique identifier of the offer.",
```
lrosenthol: Unique in what context?

extensions/adobe/experience/campaign/offer.schema.json
```diff
@@ -0,0 +1,62 @@
...
+        "https://ns.adobe.com/experience/campaign/offer/name": {
+          "title": "Offer Name",
+          "type": "string",
+          "description": "Name of the offer."
+        },
+        "https://ns.adobe.com/experience/campaign/offer/activityId": {
+          "title": "Activity ID",
+          "type": "integer",
+          "description":
+            "Unique identifier of the offer activity in the Offer Service.",
```
lrosenthol: Unique in what context?

extensions/adobe/experience/campaign/offer.schema.json
```diff
@@ -0,0 +1,62 @@
...
+        },
+        "https://ns.adobe.com/experience/campaign/offer/activityName": {
+          "title": "Activity Name",
+          "type": "string",
+          "description": "Name of the offer activity in the Offer Service."
+        },
+        "https://ns.adobe.com/experience/campaign/offer/placementId": {
+          "title": "Placement ID",
+          "type": "integer",
+          "description": "Unique identifier of the placement of the offer.",
```
lrosenthol: Unique in what context?

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: Agree.

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: Aren't these enum Labels?

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: APNS is Apple's Push notification service used to send push notifications to apple devices.

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: Correction on the way.

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: this is, infact, an integer templateID. Will change the description.

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: id: This identifies a link to be re-used in all the emails sent in a bulk delivery. For the same URL, it will be different for each delivery.
url: the URL of the link

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: This is a reference to the schema `https://ns.adobe.com/xdm-extensions/adobe/experience/campaign/offer`.
I might be missing something, but I don't think type is required here, can you help?

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: sure.

extensions/adobe/experience/campaign/offer.schema.json
saurabhere: Unique to the campaign product. So essentially not unique. Will just change the description.

extensions/adobe/experience/campaign/experienceevent.schema.json
trieloff: These are labels.

lrosenthol: So then what are the actual values that can be used?

extensions/adobe/experience/campaign/experienceevent.schema.json
lrosenthol: Ah - missed the Apple...

extensions/adobe/experience/campaign/experienceevent.schema.json
lrosenthol: Oh - I see.  Can you put some of this explanation in the descriptions

extensions/adobe/experience/campaign/experienceevent.schema.json
trieloff: As the `type: string` says: any string is valid, but only the values listed in `meta:enum` (the URLs like https://ns.adobe.com/xdm/channels/sms) are understood by Campaign.

lrosenthol: I see - I was reading the enum backwards!

extensions/adobe/experience/campaign/experienceevent.example.click.json
```diff
@@ -0,0 +1,100 @@
+{
+  "xdm:timestamp": 1522321544448,
```
trieloff: Please take a look at the definition of `xdm:timestamp`, it's a `date-time`-formatted `string`

extensions/adobe/experience/campaign/experienceevent.example.click.json
```diff
@@ -0,0 +1,100 @@
...
+  "xdm:environment": {
+    "xdm:type": "browser",
+    "xdm:browserDetails": {
+      "xdm:name": "Chrome"
+    },
+    "xdm:operatingSystem": "MAC OS"
+  },
+  "xdm:device": {
+    "xdm:typeID": "IMac",
+    "xdm:typeIDService": "https://ns.adobe.com/xdm/external/proprietary"
```
trieloff: This example looks unrealistic.

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,260 @@
...
+  "type": "object",
+  "meta:extends": ["https://ns.adobe.com/xdm/context/experienceevent"],
+  "description": "Adobe Campaign extension to ExperienceEvent.",
+  "definitions": {
+    "experienceevent": {
+      "properties": {
+        "https://ns.adobe.com/experience/campaign/message/id": {
+          "title": "Message Identifier",
+          "type": "number",
+          "description": "Adobe Campaign message's broadlog ID."
```
trieloff: What's a "broadlog"?

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,260 @@
...
+        },
+        "https://ns.adobe.com/experience/campaign/message/profile": {
+          "title": "Profile",
+          "$ref": "https://ns.adobe.com/xdm/context/profile",
+          "description":
+            "The recipient of the message. This property is primarily used to link the message to a Profile using the EndCustomerId but it can also be used to freeze some properties of the profile at the time the message was sent."
```
trieloff: Backticks around `EndCustomerID` (note the `ID`)

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,260 @@
...
+          "type": "string",
+          "description":
+            "A campaign activity can have multiple variants (Multilingual or A/B testing) but a message delivered to a recipient always contains one of them.\n> IMPORTANT: In case of multilingual, the mirror page, the recipient can switch from one variant to another (ex: switch of the language)."
+        },
+        "https://ns.adobe.com/experience/campaign/message/seedMember": {
+          "title": "Seed Member",
+          "type": "boolean",
+          "default": false,
+          "description":
+            "The recipient of this message is a member of a seed list for this campaign activity."
```
trieloff: Explain what a seed list is.

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,260 @@
...
+          "type": "boolean",
+          "default": false,
+          "description":
+            "Address moved to quarantine. The quarantine is set to true when the message failed to deliver in a way that indicates that no future messages will be deliverable either to this address."
+        },
+        "https://ns.adobe.com/experience/campaign/message/proofMember": {
+          "title": "Proof Member",
+          "type": "boolean",
+          "description":
+            "The recipient of this message is a member of the proof group for this campaign activity.",
```
trieloff: Explain what a proof group is.

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,260 @@
...
+          "default": false
+        },
+        "https://ns.adobe.com/experience/campaign/message/controlGroupMember": {
+          "title": "Control Group Member",
+          "type": "boolean",
+          "description":
+            "The recipient of this message is a member of a control group for this campaign activity. Messages to control group members are not delivered.",
```
trieloff: Explain what a control group is. The term test group and control group seem to be used differently from other areas of the industry, where control group experiences are delivered (otherwise you would not be able to compare test performance)

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,260 @@
...
+          "type": "boolean",
+          "description":
+            "The recipient of this message is a member of a test group for this campaign activity.",
+          "default": false
+        },
+        "https://ns.adobe.com/experience/campaign/message/size": {
+          "title": "Size (Bytes)",
+          "type": "integer",
+          "description":
+            "Size in bytes of the message.\n* For email, the `size` reflects the all MIME envelop of the message, including the encoded attachments if any.\n* For SMS, the `size` reflects the number of bytes necessary to deliver the message.\nIf that number is over 160 the message will be delivered over more than one SMS.",
```
trieloff: typo: envelope.

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,260 @@
...
+          "minimum": 0,
+          "maximum": 2147483647
+        },
+        "https://ns.adobe.com/experience/campaign/message/reason": {
+          "title": "Reason",
+          "type": "string",
+          "meta:enum": {
```
trieloff: In other parts of XDM, we use a convention of `lower_case_snake_case` enum keys. Might be worth adopting.

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,260 @@
...
+            "Undefined": "Not defined",
+            "UnknownUser": "User unknown",
+            "InvalidDomain": "Invalid domain",
+            "Unreachable": "Unreachable",
+            "Disabled": "Account disabled",
+            "MailboxFull": "Mailbox full",
+            "NotConnected": "Not connected",
+            "Refused": "Refused",
+            "ErrorIgnored": "Error ignored",
```
trieloff: What does that mean?

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,260 @@
...
+            "AddressUndefined": "Address not specified",
+            "Blacklisted": "Blacklisted address",
+            "Quarantine": "Quarantined address",
+            "Duplicate": "Duplicate",
+            "TypologyRule": "Excluded by a SQL rule",
```
trieloff: What does that mean?

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,260 @@
...
+            "BusinessRanking": "Excluded after arbitration",
+            "Cancelled": "Delivery canceled",
+            "Quality": "Bad-quality address",
+            "Unchecked": "Unqualified address",
+            "OfferMissing": "Not eligible for the offers",
+            "OverDeliveryLimit": "Target limited in size",
```
trieloff: How does this compare to `TooLong`?

extensions/adobe/experience/campaign/experienceevent.schema.json
trieloff: Makes sense.

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,260 @@
...
+          "meta:enum": {
+            "https://ns.adobe.com/xdm/channels/email": "Email",
+            "https://ns.adobe.com/xdm/channels/sms": "Mobile (SMS)",
+            "https://ns.adobe.com/xdm/channels/phone": "Phone",
+            "https://ns.adobe.com/xdm/channels/direct-mail": "Direct Mail",
+            "https://ns.adobe.com/xdm/channels/fax": "Fax",
+            "https://ns.adobe.com/xdm/channels/apns":
+              "Apple Push Notification Service",
+            "https://ns.adobe.com/xdm/channels/gcm": "Google Cloud Messaging",
+            "agency": "Agency",
```
trieloff: Introducing this as a proper channel (URL, schema) would make sense.

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,260 @@
...
+            "other": "Other"
```
trieloff: I would not list `other` here. The field is a `string` anyway, so you can put other stuff if you want to or keep it `undefined`

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,260 @@
...
+          },
+          "description": "The marketing channel used to execute this activity."
+        },
+        "https://ns.adobe.com/experience/campaign/campaign/id": {
+          "title": "CampaignID",
+          "type": "integer",
+          "description":
+            "The marketing campaign to which activity originating this message belongs to.."
```
trieloff: Too many `.`

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,260 @@
...
+        },
+        "https://ns.adobe.com/experience/campaign/delivery/from": {
+          "title": "From",
+          "type": "string",
+          "description":
+            "Address used as a from/sender/caller address. Depending on the communication channel, the `address` has a different format.\n\n  * `email`: an email address.\n  * `textMessage` or `phone`: a shortcode or phone number."
```
trieloff: It might make sense to make this more specific:
- using `oneOf`
- by having schemas for each delivery type

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,260 @@
...
+        },
+        "https://ns.adobe.com/experience/campaign/delivery/testEnabled": {
+          "title": "Test Enabled",
+          "type": "boolean",
+          "description":
+            "Specify whether campaign activity is of type A/B testing.",
+          "default": false
+        },
+        "https://ns.adobe.com/experience/campaign/delivery/messagesClass": {
```
trieloff: message*s* like in the property name or message like in the title?

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,260 @@
...
+          "title": "Message Class",
+          "type": "string",
+          "meta:enum": {
```
trieloff: I'd use `snake_case` keys.

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,260 @@
...
+            "oneTime": "Marketing (One Time)",
+            "continuous": "Recurring",
+            "messageCenter": "Transactional (Event-based)"
+          },
+          "description":
+            "Marketing classification of the messages sent through that delivery.\n",
```
trieloff: Needs a better description. Right now, the description makes me think this is about the marketing classification (i.e. is it promotional stuff that I can ignore, or is it the transaction confirmation that I want to keep). The keys and descriptions in `meta:enum` make it sound like it is mostly about the frequency of delivery, not about the nature.

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,260 @@
...
+            "Specify whether campaign activity is of type A/B testing.",
+          "default": false
+        },
+        "https://ns.adobe.com/experience/campaign/delivery/messagesClass": {
+          "title": "Message Class",
+          "type": "string",
+          "meta:enum": {
+            "oneTime": "Marketing (One Time)",
+            "continuous": "Recurring",
+            "messageCenter": "Transactional (Event-based)"
```
trieloff: Looks like a product name or feature name to me. This is dangerous, because feature names change faster than underlying implementations. I'd pick something more neutral and boring.

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,260 @@
...
+          },
+          "description":
+            "Marketing classification of the messages sent through that delivery.\n",
+          "default": 0
+        },
+        "https://ns.adobe.com/experience/campaign/delivery/templateID": {
+          "title": "Template",
+          "type": "integer",
+          "description":
+            "The delivery template's ID used to initialize this delivery.\n\nThe type the template used in Adobe Campaign can be identified using the `messageClass`.\n\n* `messageClass` = `oneTime`: the template is standard delivery template.\n* `messageClass` = `continuous`: the template is a recurring delivery.\n* `messageClass` = `transactional`: the template is a transactionnal message template.\n"
```
trieloff: The description is inconsistent with the keys defined above.

extensions/adobe/experience/campaign/offer.schema.json
```diff
@@ -0,0 +1,62 @@
...
+    "Copyright 2018 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm-extensions/adobe/experience/campaign/offer",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Offer",
+  "type": "object",
+  "description":
```
trieloff: Nice. But what is an offer?

extensions/adobe/experience/campaign/offer.schema.json
```diff
@@ -0,0 +1,62 @@
...
+          "description": "Identifier of the offer.",
+          "minimum": 0,
+          "maximum": 2147483647
+        },
+        "https://ns.adobe.com/experience/campaign/offer/name": {
+          "title": "Offer Name",
+          "type": "string",
+          "description": "Name of the offer."
+        },
+        "https://ns.adobe.com/experience/campaign/offer/activityId": {
```
trieloff: You have `offer/id` above and `activityId` here.

extensions/adobe/experience/campaign/offer.schema.json
```diff
@@ -0,0 +1,62 @@
...
+          "title": "Placement Name",
+          "type": "string",
+          "description": "Name of the placement of the offer."
+        }
+      }
+    }
+  },
+  "allOf": [
+    {
+      "$ref": "#/definitions/offerplacement"
```
trieloff: This ref does not exist. It is called `offer` above.

schemas/context/device.schema.json
```diff
@@ -27,7 +27,8 @@
           "description":
             "The namespace of the service that is used to identify the device type.",
           "meta:enum": {
-            "https://ns.adobe.com/xdm/external/deviceatlas": "Device Atlas"
+            "https://ns.adobe.com/xdm/external/deviceatlas": "Device Atlas",
+            "https://ns.adobe.com/xdm/external/proprietary": "Proprietary"
```
trieloff: What does that mean?

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: a unique broadlog refers to every single message sent as part of a campaign activity (which is called a delivery). I'd rather call it a  `message` here.

extensions/adobe/experience/campaign/experienceevent.example.click.json
saurabhere: :+1:

extensions/adobe/experience/campaign/experienceevent.example.click.json
saurabhere: @trieloff This is derived from an actual event that currently Adobe Campaign receives when someone clicks on a web-link inside an email. Can you help me understand what ways this should be improved?

Regarding the `typeIDService`, at Adobe Campaign, we parse the user-agent string and have our own enumeration of values that are set as the value of `device` depending upon some regex rules. This might look bad and we have plans to move to DeviceAtlas. But until this happens, we have data that flows using this. 
I assume there may be similar producers and this "proprietary" enum was created to honor a contract between the producers and consumers that the service is not a standard one but their own internal service. Basically, an enum for "others". Yes, this would not work when the consumer doesn't know the producer or consumes from multiple producers using different "proprietary" service.

I first thought of extending the device schema (say `campaign/device.schema`) to add this "proprietary" enum value but I couldn't find a way to extend an enum set through an extended schema. Also, `context/experienceEvent` has the `device` field. I would want this to be overridden to `campaign\device` in `campaign/experienceEvent`. I don't think we can do that unless we create a fresh field.
As a simple solution, I could just change this to call it `CampaignDeviceService` if that seems best but I didn't want to include Adobe specific things in the core xdm.
Any suggestions?

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: sure

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: :+1

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: :+1:

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: :+1:

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: Control Group members are a random set of recipients which are not selected to be not targeted. Then the behaviour changes in the control group vs the targeted can be used to measure the campaign's effectiveness.

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: :+1:

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: This means that there was an error while sending to the recipient but since the user is whitelisted, we will ignore the error and not mark them as blacklisted. I will update this in the description.
For more: https://docs.campaign.adobe.com/doc/AC/en/DLV_Tracking_deliveries_Quarantine_management.html

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: This _probably_ means "The campaign activity had more than allowed number of target profiles.".

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: Updated description. This is more campaign specific.

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: removed.

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: :+1:

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: Not sure what the question is.

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: :+1:

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: `event_based` it will be!

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: changed!

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: Not sure, but I changed the title to "Identifier of the Template" if that's what you mean.

extensions/adobe/experience/campaign/offer.schema.json
saurabhere: :+1:

schemas/context/device.schema.json
saurabhere: replied above.

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: :+1:

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: :+1:

extensions/adobe/experience/campaign/experienceevent.schema.json
lrosenthol: please include these definitions in the description.

saurabhere:
Gentle nudge to @lrosenthol @trieloff to please re-review with the latest changes.

extensions/adobe/experience/campaign/experienceevent.example.click.json
trieloff: I would just give this a URL that includes the name `Adobe Campaign`. @asteed might have a suggestion that is consistent with our schemas.

Just having a bucket for "others" is not useful, but knowing that "Adobe Campaign" though this is an IMac, adds a lot of value.

extensions/adobe/experience/campaign/experienceevent.schema.json
trieloff: It is "messageClass", as in class of the message - or is it "message*s*Class, which either sounds ungrammatical and inconsistent or hides a subtlety that evades me.

extensions/adobe/experience/campaign/experienceevent.schema.json
trieloff: You had different keys in the description than you defined in `meta:enum` above.

extensions/adobe/experience/campaign/offer.schema.json
trieloff: Please also change the property to `activityID` – see `CONTRIBUTING.md` for the style guide.

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,257 @@
...
+          "title": "Profile",
+          "$ref": "https://ns.adobe.com/xdm/context/profile",
+          "description":
+            "The recipient of the message. This property is primarily used to link the message to a Profile using the `EndCustomerID` but it can also be used to freeze some properties of the profile at the time the message was sent."
+        },
+        "https://ns.adobe.com/experience/campaign/message/variant": {
+          "title": "Content Variant",
+          "type": "string",
+          "description":
+            "A campaign activity can have multiple variants (Multilingual or A/B testing) but a message delivered to a recipient always contains one of them.\n> IMPORTANT: In case of multilingual, the mirror page, the recipient can switch from one variant to another (ex: switch of the language)."
```
trieloff: > In case of multilingual

multilingual what?

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,257 @@
...
+        },
+        "https://ns.adobe.com/experience/campaign/message/seedMember": {
+          "title": "Seed Member",
+          "type": "boolean",
+          "default": false,
+          "description":
+            "The recipient of this message is a member of a seed list for this campaign activity. Seed addresses are used to target recipients who do not match the defined target criteria. Seed members could be your marketing team or other stakeholders who are interested in tracking/measuring the quality of the campaign."
```
trieloff: Maybe:

> Seed addresses are used to include recipients who do not match the defined target criteria.

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,257 @@
...
+        },
+        "https://ns.adobe.com/experience/campaign/message/quarantine": {
+          "title": "Quarantine",
+          "type": "boolean",
+          "default": false,
+          "description":
+            "Address moved to quarantine. The quarantine is set to true when the message failed to deliver in a way that indicates that no future messages will be deliverable either to this address."
```
trieloff: Backticks around `true`

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,257 @@
...
+        },
+        "https://ns.adobe.com/experience/campaign/message/proofMember": {
+          "title": "Proof Member",
+          "type": "boolean",
+          "description":
+            "The recipient of this message is a member of the proof group for this campaign activity. The proof is a special message that lets you test a delivery before sending to the main target. Proof recipients are responsible for approving both the form and content of the message.",
```
trieloff: Do I get this right:
- `proofMember` gets the proof message, which is a pre-delivery
- `seedMember` gets the actual message, at the same time with regular recipients, even when the `seedMember` does not qualify to be a regular recipient?

If yes, include this in the description.

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,257 @@
...
+          "default": false
+        },
+        "https://ns.adobe.com/experience/campaign/message/controlGroupMember": {
+          "title": "Control Group Member",
+          "type": "boolean",
+          "description":
+            "The recipient of this message is a member of a control group for this campaign activity. A control group is a population which will not receive the delivery; it is used to track post-delivery behavior and campaign impact by making a comparison with the behavior of target population, which has received the delivery.",
```
trieloff: Ah. It would make sense to include the terms "hold out" in the description.

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,257 @@
...
+          "default": false
+        },
+        "https://ns.adobe.com/experience/campaign/message/testMember": {
+          "title": "A/B Test Member",
+          "type": "boolean",
+          "description":
+            "The recipient of this message is a member of a test group for this campaign activity. This is primarily used for identifying the initial population of an A/B Test.",
```
trieloff: I think we need to add a bit more here.

The way I'm reading this right now is that there is a group of pre-determined people that will be included for testing and that might see different variants of the message.

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,257 @@
...
+          },
+          "description":
+            "The reason why the message could/would not be delivered.",
+          "default": "Undefined"
+        },
+        "https://ns.adobe.com/experience/campaign/message/reasonMessage": {
+          "title": "Reason Message",
+          "type": "string",
+          "description":
+            "Reason of message delivery failure when the status is `bounce` or `nonDeliverable`."
```
trieloff: Ok, I've been searching for a property `status` and could not find it. Same for the values `bounce` and `nonDeliverable`. The only place where they are mentioned is here. You are either missing a property, or the description is off.

extensions/adobe/experience/campaign/experienceevent.schema.json
trieloff: In particular: `messagesClass` above, `messageClass` here.

extensions/adobe/experience/campaign/experienceevent.schema.json
trieloff: For the campaign extension I think it is ok, but in other places we need to standardize.

extensions/adobe/experience/campaign/experienceevent.schema.json
trieloff: Yes, the `$ref` is giving you a type.

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,257 @@
...
+          }
+        },
+        "https://ns.adobe.com/experience/campaign/message/offerClicked": {
+          "title": "Offer Clicked",
+          "$ref":
+            "https://ns.adobe.com/xdm-extensions/adobe/experience/campaign/offer",
+          "description":
+            "The offer and its location in the message for a click event related to an offer."
+        },
+        "https://ns.adobe.com/experience/campaign/message/externalId": {
```
trieloff: Should be `external/id` for consistency. Or `externalID` for style guide.

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -0,0 +1,257 @@
...
+          "title": "External ID",
+          "type": "string",
+          "description":
+            "An ID generated by the parent activity invoking service to check later if the incoming experienceEvent was generated using this ID."
+        },
+        "https://ns.adobe.com/experience/campaign/businessContext": {
+          "title": "Business Context",
+          "type": "string",
+          "description":
+            "Internal field for identifying the various actions/intents."
```
trieloff: This is almost worse than useless. What are the values that this can take?

extensions/adobe/experience/campaign/offer.schema.json
```diff
@@ -0,0 +1,64 @@
...
+            "Identifier of the offer activity in the Offer Service. Offer Activity is a container which may have multiple offers and a subset of them will be shown to the user based on their profile attributes or rules set by the Marketeer. ",
+          "minimum": 0,
+          "maximum": 2147483647
+        },
+        "https://ns.adobe.com/experience/campaign/offer/activityName": {
+          "title": "Activity Name",
+          "type": "string",
+          "description": "Name of the offer activity in the Offer Service."
+        },
+        "https://ns.adobe.com/experience/campaign/offer/placementId": {
```
trieloff: `placement/id`

trieloff: Also, if you can use `uri` references, they are much better than `integer` references.

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: Oh that's a typo. Nice pair of eyes there.

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: I like it this way.

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: sure

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: Fixing it.

extensions/adobe/experience/campaign/offer.schema.json
saurabhere: :+1: for `placement/id`
not sure if I can use uri refs here with our current implementation. I'd like to keep it this way for now.

extensions/adobe/experience/campaign/experienceevent.schema.json
saurabhere: the original request comes from the internal issue plat-14497

trieloff: In assets, we have a property `layerStyles` that controls how the layer is styled. All that is known about the property is that it is a `string`.

This property clearly is a structured property that either points to another value or encodes the actions or intents. If that's the case, we should either document it, or leave it out entirely.

trieloff:
From my point of view:

- [ ] fix https://github.com/adobe/xdm/pull/209#discussion_r180053125
- [ ] decide what to do with `businessContext`: either you take it out or you document what the allowed value and semantics of the field are.

jbeckert:
As a representative of the offer management core service I strongly urge that campaign does not introduce an offer XDM under https://ns.adobe.com/xdm-extensions/adobe/experience/campaign/offer
@dmitry-shkolnik is going to introduce a comprehensive set of models for Offer Management in general

dmitry-shkolnik:
Our initial set of XDMs for Offer Management is now available for the review at https://github.com/adobe/xdm/pull/245

extensions/adobe/experience/campaign/experienceevent.schema.json
lrosenthol: and it's not xdm:Tags - it is using a namespace, so that should indeed be fine.
👍

extensions/adobe/experience/campaign/experienceevent.schema.json
lrosenthol: 👍

extensions/adobe/experience/campaign/experienceevent.schema.json
lrosenthol: 👍

extensions/adobe/experience/campaign/experienceevent.schema.json
lrosenthol: 👍

extensions/adobe/experience/campaign/experienceevent.example.click.json
```diff
@@ -0,0 +1,100 @@
...
+      }
+    },
+    "variant": "English",
+    "seedMember": false,
+    "quarantine": false,
+    "proofMember": false,
+    "controlGroupMember": false,
+    "testMember": true,
+    "outboundIP": "10.20.30.40",
+    "offerProposed": [
```
jbeckert: @saurabhere you updated the schema to take out the offerProposed property but the examples still have it.

extensions/adobe/experience/campaign/experienceevent.example.click.json
```diff
@@ -0,0 +1,100 @@
...
+      {
+        "id": 1235,
+        "name": "bogo",
+        "activityId": 1357,
+        "activityName": "major-banner",
+        "placementId": 3,
+        "placementName": "sidebar"
+      }
+    ],
+    "offerClicked": {
```
jbeckert: See above, offerClicked is removed (as discussed) but still present in the example

extensions/adobe/experience/campaign/experienceevent.example.email-send.json
```diff
@@ -0,0 +1,68 @@
...
+    },
+    "variant": "English",
+    "seedMember": false,
+    "quarantine": false,
+    "proofMember": false,
+    "controlGroupMember": false,
+    "testMember": true,
+    "size": 421,
+    "outboundIP": "10.20.30.40",
+    "offerProposed": [
```
jbeckert: Same as above. You updated the schema to take out the offerProposed property, as discussed and desired, but the examples still have it.

extensions/adobe/experience/campaign/experienceevent.example.click.json
saurabhere: missed it.

extensions/adobe/experience/campaign/experienceevent.example.click.json
saurabhere: missed it.

extensions/adobe/experience/campaign/experienceevent.example.email-send.json
saurabhere: missed it.

saurabhere:
@trieloff Can this be merged now please?

saurabhere:
thanks for noticing this. fixed.

Issue #187, opened 2018-03-20
Determine candidate schemas to be considered 'stable'


asteed:
Need a mechanism to find and list all schemas which are not yet marked `"meta:status": "stable"` but meet the requirements for `stable` candidacy, including:
* Schema has spent a minimum of 3 months under `stabilizing` status
* No major changes in the past 3 months

An automated way is needed to generate a list of these available candidate schemas

trieloff:
I think #198 is a good first step, but I see a couple of improvements:

- it does not run at all under node 6
- under node 8, the list of candidates is `undefined`
- you can probably use `$.exec` instead of promisifying `ChildProcess.exec`

trieloff:
Thanks for #213 – here are a few more ideas:

1. group schemas by status and suggested status - the schemas that can be promoted should go first
2. have a list (probably in a file) of "known to be trivial" updates. This way we can retroactively mark things as trivial
3. use relative dates (7 days ago)
4. use `backticks` around the path name
5. don't use `code` for the commit message

asteed:
Thanks @trieloff - could you provide a couple clarifications?
1. :ok_hand: 
2. Are these known to be trivial 'updates' indicated simply by specific git revisions? eg: 
```
2e0cffca526751122a3c0720435cb9b52de48a0e
eb2a226a2a2f4c0d7c13fdd8e63f95355619af35
4661b2729388847417c562adc2a585e6b87e0758
994d2c9c6f604efe0041b2c4b41c9e730afa5197
aa3d853a36a0b5b7bf3d34a226243c153b90ee5d
```
3. My experience with relative dates (I did some initial look into it for this task) is that it will require another npm dependency. Do you see any issues with the addition of a dependency for this?
4. :ok_hand:
5. :ok_hand: for removing `<code>` however retaining the `<br>` to support multiline table cells for multi-commits

PR #213, opened 2018-03-30
Requested changes to stabilization candidates script

asteed:
Addresses @trieloff's review feedback in Issue #187


trieloff:
Running it now, we might have to add a warning or progress bar as it can be a bit time-consuming.

Issue #177, opened 2018-03-19
Revert person-name from given/sur(name) back to first/last(name) for industry compatibility.


cdegroot-adobe:
## What are the schemas that are affected by the issue
https://github.com/adobe/xdm/blob/master/schemas/context/person-name.schema.json

## What are examples of products that are impacted by the issue
This will bring in greater alignment with the industries current norms, which are implemented and very hard to change. While it is agreed that Given Name and Surname are superior for representing names, it is not the way current systems work and there is no reasonable automatic way of translating between the two representations. XDM will be harder to adopt unless this change is made.

Examples of the first/last(name) usage:
1. https://github.com/Microsoft/CDM/blob/master/schemaDocuments/core/applicationCommon/foundationCommon/crmCommon/Contact.cdm.json
2. https://developer.salesforce.com/docs/atlas.en-us.api.meta/api/sforce_api_objects_contact.htm
3. https://adobe-apiplatform.github.io/umapi-documentation/en/api/getUser.html
4. Gigia Profile https://developers.gigya.com/display/GD/Profile+REST
5. Facebook, twitter, linkedin, paypal, amazon.

Examples for given/sur(name):
1. International documents/passports
2. Governmental institutions 
3. W3C https://www.w3.org/International/questions/qa-personal-names

cdegroot-adobe:
@kstreeter  fyi, I am putting this back on the radar as an issue to resolve. Thanks C.

trieloff:
You can add schema.org to the list of examples for given name: http://schema.org/Person

cdegroot-adobe:
@trieloff it is actually worse then that. schema.org uses familyName rather than surName

E.g from http://schema.org/Person
familyName | Text | Family name. In the U.S., the last name of an Person. This can be used along with givenName instead of the name property.

kstreeter:
@cdegroot-adobe, @trieloff it definitely seems like there is no widely adopted norm on this, so we need to decide our own path.  The W3C and schema.org seems the most considered to me.  That would be my preference.  We can ensure that the descriptions reference the alternative nomenclatures (first, last, surname).

trieloff:
@kstreeter – agreed.

cdegroot-adobe:
@kstreeter  @trieloff - disagree.
There is a very entrenched norm in our customers and the industry and that is to use firstname/last(name). W3C and Schema.org are leading thinkers in the way the world needs to go(at some point), but has not yet gone is any sizable way.
[...]

PR #215, opened 2018-04-02
Revert person-name from given/sur(name) back to first/last(name) for industry compatibility

kstreeter:
Please link to the issue #…
PR for [Issue #177](https://github.com/adobe/xdm/issues/177)

### DIFF
--- a/schemas/context/person-name.schema.json
+++ b/schemas/context/person-name.schema.json
@@ -13,13 +13,13 @@
   "definitions": {
     "personname": {
       "properties": {
-        "xdm:givenName": {
-          "title": "Given name",
+        "xdm:firstName": {
+          "title": "First name",
           "type": "string",
           "description": "The personal, given, or first name."
         },
-        "xdm:surname": {
-          "title": "Surname",
+        "xdm:lastName": {
+          "title": "Last name",
           "type": "string",
           "description":
             "The inherited family name, last name, surname, patronymic, or matronymic name."
@@ -36,7 +36,7 @@
           "description":
             "Normally an abbreviation of a persons *title*, *honorific*, or *salutation*.\nThe `courtesyTitle` is used in front of full or last name in opening texts.\ne.g Mr. Miss. or Dr J. Smith.\n"
         },
-        "xdm:name": {
+        "xdm:fullName": {
           "title": "Full name",
           "type": "string",
           "description":
--- a/schemas/context/person-name.example.1.json
+++ b/schemas/context/person-name.example.1.json
@@ -1,5 +1,5 @@
 {
-  "xdm:givenName": "John",
+  "xdm:firstName": "John",
   "xdm:middleName": "S",
-  "xdm:surname": "Doe"
+  "xdm:lastName": "Doe"
 }
--- a/schemas/context/person-name.example.2.json
+++ b/schemas/context/person-name.example.2.json
@@ -1,5 +1,5 @@
 {
-  "xdm:givenName": "三",
-  "xdm:surname": "张",
-  "xdm:name": "张三"
+  "xdm:firstName": "三",
+  "xdm:lastName": "张",
+  "xdm:fullName": "张三"
 }
--- a/schemas/context/person-name.example.3.json
+++ b/schemas/context/person-name.example.3.json
@@ -1,5 +1,5 @@
 {
-  "xdm:givenName": "فلانة",
-  "xdm:surname": "الفلانية",
-  "xdm:name": "فلانة الفلانية"
+  "xdm:firstName": "فلانة",
+  "xdm:lastName": "الفلانية",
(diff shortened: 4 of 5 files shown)


lrosenthol:
Looks good to me - thanks @kstreeter

cdegroot-adobe:
@kstreeter yes, this works very well for me. I approve.

schemas/context/person-name.schema.json
```diff
@@ -36,7 +36,7 @@
           "description":
             "Normally an abbreviation of a persons *title*, *honorific*, or *salutation*.\nThe `courtesyTitle` is used in front of full or last name in opening texts.\ne.g Mr. Miss. or Dr J. Smith.\n"
         },
-        "xdm:name": {
+        "xdm:fullName": {
```
trieloff: We use `xdm:name` for the proper name in other places, I don't see reason to deviate here.

schemas/context/person-name.schema.json
```diff
@@ -13,13 +13,13 @@
...
-          "title": "Given name",
+        "xdm:firstName": {
+          "title": "First name",
           "type": "string",
           "description": "The personal, given, or first name."
         },
-        "xdm:surname": {
-          "title": "Surname",
+        "xdm:lastName": {
+          "title": "Last name",
```
trieloff: Ok, so this is just a title replacement? I think it is making it even more confusing. Now what is labeled `lastName` is indeed a "surname" and should be treated like one.

cdegroot-adobe: @trieloff - the point of this set of changes is quite well thought out considering the complex issues. The generally agreed best principles going forward in person naming is that there is no correct way to separate person name components and that it specifically should be avoided. Hence "fullName" which is inline with these recommendations. The current overwhelming standard in the industry due to legacy RDBMS and data principles has been to have a first/last name. So that is something we cannot ignore. give/sur name actually are significant issues in their own right and just introduce a different version of first/last name problem with unique new semantics in non western cultures. 

So the root proposal is: 
a) we have a longer term correct approach with "fullName" that is more commonly being used in parallel with separated name component semantics.
b) we retain the current standard in data which is first/last name

I continue to agree with this proposal. and the interoperability it provides.

It will be interesting to see if implementation start to use "legalName", which I expect they will in some industries, and that is specifically a reason I am avoiding the "name" usage. It is very common to have a different legalName, fullName and firstName+{space}+lastName, I am an example of that. And the caucasian person who sits beside me has an issues where his firstName != givenName.

schemas/context/person-name.schema.json
cdegroot-adobe: @trieloff So what would you put in "name" a persons preferred name or their legal(called when in trouble or by ones mother) name.

schemas/context/person-name.schema.json
cdegroot-adobe: We need to improve the descriptions.
I propose:
"The preferred first name of the person."

schemas/context/person-name.schema.json
```diff
@@ -36,7 +36,7 @@
           "description":
             "Normally an abbreviation of a persons *title*, *honorific*, or *salutation*.\nThe `courtesyTitle` is used in front of full or last name in opening texts.\ne.g Mr. Miss. or Dr J. Smith.\n"
         },
-        "xdm:name": {
+        "xdm:fullName": {
           "title": "Full name",
           "type": "string",
           "description":
```
cdegroot-adobe: We need to improve the descriptions.
I propose:
"The preferred last name of the person."

schemas/context/person-name.schema.json
kstreeter: +1 on @cdegroot-adobe description of the rationale.  This is not intended to be a choice of first/last versus given/surname.  We are saying that the preference is to use the full name (which is inline with the recommendations from W3C, and compatible with schema.org), but because our own applications and many applications in the industry use first/last we provide those fields to allow a direct translation.

schemas/context/person-name.schema.json
kstreeter: I feel like "fullName" makes it clearer what should go in here.

schemas/context/person-name.schema.json
trieloff: @cdegroot-adobe I think you meant to make this comment on `xdm:lastName`, not `xdm:fullName`.

schemas/context/person-name.schema.json
trieloff: @cdegroot-adobe I'd put in:

> The full name of the person, in writing order most commonly accepted in the language of the name.

Exactly as the description says. I think the `title="Full Name"` makes things clear enough and makes things more consistent as we don't have `*name` properties elsewhere. `fullName` is a special case that in my mind does not have enough justification. (Using the full name in the `name` property is something I absolutely support)

PR #216, opened 2018-04-03
Move common schemas for 0.9 to "stabilizing"

kstreeter:
Starting the process of moving schemas to the "stabilizing" status.  This first PR will contain the ones from "common" that are needed for 0.9.

### DIFF
--- a/schemas/common/auditable.schema.json
+++ b/schemas/common/auditable.schema.json
@@ -53,5 +53,5 @@
         "http://ns.adobe.com/adobecloud/core/1.0#/definitions/date-properties"
     }
   ],
-  "meta:status": "experimental"
+  "meta:status": "stabilizing"
 }
--- a/schemas/common/auditable.schema.json
+++ b/schemas/common/auditable.schema.json
@@ -53,5 +53,5 @@
         "http://ns.adobe.com/adobecloud/core/1.0#/definitions/date-properties"
     }
   ],
-  "meta:status": "stabilizing"
+  "meta:status": "experimental"
 }
--- a/schemas/common/extensible.schema.json
+++ b/schemas/common/extensible.schema.json
@@ -251,5 +251,5 @@
       "$ref": "#/definitions/@context"
     }
   ],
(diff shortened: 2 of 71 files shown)


kstreeter:
@trieloff, @lrosenthol I'd like to promote these schemas to "stabilizing".  We don't have any issues or PRs currently open on these.  These are dependencies for ExperienceEvent and Profile, targeted to go to stabilizing in the 0.9 milestone.

lrosenthol:
@kstreeter Address & Extensible are just fine.  Organization needs a minor change (I think location should use address and not place, since place isn't part of common).

But geo and auditable need major work.  Geo isn't aligned with the geo work that the original DataModel team did that involves actual geospatial coordinates.  Auditable seems to be a minor update to core and is missing a boatload of other useful elements that other types of auditable transactions (eg. signature workflow) would require.

schemas/common/extensible.schema.json
```diff
@@ -251,5 +251,5 @@
       "$ref": "#/definitions/@context"
     }
   ],
-  "meta:status": "experimental"
+  "meta:status": "stabilizing"
```
trieloff: 👍

schemas/common/auditable.schema.json
```diff
@@ -53,5 +53,5 @@
         "http://ns.adobe.com/adobecloud/core/1.0#/definitions/date-properties"
     }
   ],
-  "meta:status": "experimental"
+  "meta:status": "stabilizing"
```
trieloff: 👍

lrosenthol:
Sorry, @trieloff , but I disagree at least on Geo & auditable.  They are nowhere near ready to go.

trieloff:
@kstreeter can you take Geo and Auditable out? This should not hold up the schemas where we have consensus.

@lrosenthol please file two issues for the general improvements you'd like to see in Geo and Auditable.

lrosenthol:
I have added #221  and #222 to cover the geo and auditable issues respectively.

Issue #220, opened 2018-04-04
Remove stateProvinceISO from examples


trieloff:
We still have a number of examples that use the `stateProvinceISO` property, which has been replaced by `stateProvince`

asteed:
@trieloff - looking through the examples, each has both `xdm:stateProvince` and `xdm:stateProvinceISO` defined. Do we just expect to _**remove**_ the latter entry from each example?

```
    "xdm:stateProvince": "California",
    "xdm:stateProvinceISO": "CA",
```

PR #230, opened 2018-04-05
Remove stale 'xdm:stateProvinceISO' from examples

asteed:
Resolves Issue #220


trieloff:
@asteed please take a look at d805fa1

PR #240, opened 2018-04-09
Remove id from namespace

cdegroot-adobe:
The team that owns the namespace concept have identified an issue related to the @id field that means we really prefer customers to use the namespace.code field only. This change implements that and removes the id usage from the examples.


schemas/context/namespace.schema.json
```diff
@@ -9,26 +9,18 @@
...
-          "format": "uri",
-          "description":
-            "The numeric ID associated with this namespace. This would be provided by the individual or system that created the namespace.\n"
-        },
         "xdm:code": {
           "title": "Code",
           "type": "string",
           "description":
-            "The code is a shortcut to the full @id and at least one of the code or @id can be used. Sometimes, this code refered to as the data source integration code."
+            "The code is a human readable identifier for the namespace and can be used to request the technical namespace id which is used for identity graph processing."
```
trieloff: The technical namespace ID is referenced in the description, but not visible in the schema.

How is uniqueness of the `code` ensured?

Who governs the creation of new `code` values?

schemas/context/namespace.description.md
```diff
@@ -0,0 +1,15 @@
+The namespace ensures first, second, and third party user and device identities can be used in customer engagement platforms.
```
trieloff: Is this the only use case for a namespace?

schemas/context/namespace.description.md
```diff
@@ -0,0 +1,15 @@
...
+
+Namespaces are used to attain global uniqueness with user and device identities that originate in distributed, uncoordinated identity environments.
```
trieloff: The claim that they create global uniqueness is not backed up by the description. We do have a mechanism for ensuring global uniqueness (URIs) that for some reason that is not made clear is not being used here.

I can assume that the reason is that URIs might be a bit too long as a wire format, but this is only an assumption (and my assumptions have been false in the past)

schemas/context/namespace.description.md
```diff
@@ -0,0 +1,15 @@
+The namespace ensures first, second, and third party user and device identities can be used in customer engagement platforms.
+
+Namespaces are used to attain global uniqueness with user and device identities that originate in distributed, uncoordinated identity environments. 
+It can be thought of as an automatic prefix generator that can be used to append a unique prefix to other generators of identities that are only guaranteed to be unique in the scope of the individual generating system (the standard case for cookies and integer based identities).
```
trieloff: What can be thought of as a prefix generator? The namespace? Or the entity that generates the namespace? 

Again, I can only guess here: we have a number of systems that don't know each other that generate IDs. In order to keep the IDs apart, we concatenate the generated ID with an ID of the generator. 

What's not being made clear is:
1. does namespace describe the generator or the generated output
2. who creates IDs for generators?

schemas/context/namespace.description.md
```diff
@@ -0,0 +1,15 @@
...
+
+The namespace does not describe how the additional prefix is to be used to augment the identities it is making unique, that is up to the systems that process the new, globally unique identities.
+
+Namespaces have a simplified code property that is easier for humans to use.
```
trieloff: Right now, it looks like they only have that code property (and other hidden property that should be part of the schema)

schemas/context/namespace.description.md
```diff
@@ -0,0 +1,15 @@
...
+Case 1: Global Namespace Codes: These codes allow reuse of common namespaces for identities that are already globally unique such as e-mail addresses, phone numbers or push notification tokens.
+Case 2: Implementation Specific Codes: Are only for use in a specific context, like an individual brand and have a scope that is specific to that customer engagement context. In this case two implementations can use the same code, but underlying namespaces would be different.
+
+Namespace identifiers are the technical representation of the namespace and are what is actually used as part of user or device identification. Generally, these identifiers should be looked up only at the time of use to ensure contextual correctness.
```
trieloff: Given that there is no ID attribute or identifier attribute, I don't know what this refers to. Is namespace identifier == namespace code or namespace identifier == namespace?

schemas/context/namespace.description.md
```diff
@@ -0,0 +1,15 @@
...
+
+Related:
+Namespaces are similar to datasources in the way they operate. The core difference being datasources are used to uniquely identify specific sources of data in an uncoordinated data environment, such as independent data collection applications.
```
trieloff: To me it sounds like the two concepts can be merged in XDM. Both address the governance of data in a distributed fashion, both do it by labeling data points with the identity of the governor.

The only minor difference is that one deals with data that has been collected and the other with data that has been generated. 

If we need to model data that has been inferred, will we introduce a third concept?

trieloff:
I think there is another inherent problem here: we want the `code` to be:
- compact (it needs to go on the wire)
- human readable
- globally unique

Finally, in order for this to fit into XDM, it needs to be open and extensible.

So if the answer is "Adobe will govern the creation and management of `code` values", with the implication "`code` generation is only available to Adobe enterprise customers", we run afoul of XDM's charter.

schemas/context/namespace.description.md
cdegroot-adobe: Yes

kstreeter:
Hi @trieloff, @cdegroot-adobe I have made some changes to the description that I hope makes it much clearer how namespaces work.  Please take a look and let me know what you think.

lrosenthol:
I am VERY confused here.  What does this “namespace” functionality have to do with our namespacing/prefixing inside of XDM?

trieloff:
@lrosenthol nothing. It's a completely orthogonal concept.

@kstreeter it might make sense to point this out in the description – unless @lrosenthol thinks that it's a severe enough issue that warrants a re-naming.

trieloff:
@lrosenthol is the concept clear now?

@kstreeter reading through the updated description I've wondered if we should call the entire concept "Identity Namespace" or "Identity Space" instead to avoid confusion.

kstreeter:
@trieloff I would be okay with calling it IdentityNamespace if others agree this is helpful in aiding understanding.

kstreeter:
@trieloff, actually after talking to the team here it would be challenging to change the name of this now.  We have several inter-connected components using the field now.  Suggest we keep it as is, and document clearly.

PR #266, opened 2018-04-17
Issue 3 update address

cdegroot-adobe:
Added the missing field that brings broad alignment with CDM “stateProvince”.
Added ISO fields for country and stateProvince to support locale independent processing. These were all in the original PR that created address but were removed during the initial merge.


kstreeter:
Will need to decide this for CDM integration.

trieloff:
I stand by my earlier comment: this is making things more ambiguous and takes us away from using established standards.

cdegroot-adobe:
@trieloff I can see that for the state/province we need to move to the ISO-3166-2 which is a mix of ISO-3166-1(country) and a 1-3 character state/province designation. This the correct representation of the technical/locale neutral identity.

For the naming we need the have both the common name the owner of the record decided to use(and all the vagrancies of that) and the ISO code. Can you propise a better way of naming the property then *+ISO.

kstreeter:
@trieloff, I can confirm what @cdegroot-adobe is describing here.  We have cases (such as interop with CDM) where the "source" does not have an ISO code.  They just have a field.  It is problematic to force an upstream conversion (for example, what happens when the conversion fails?) so we want to be able to at least bring the data in when the code is not available.  Having both code-based and informal names available will help.

kstreeter:
HI @cdegroot-adobe, can we use the suffix "Code" to designate the code-based fields, versus "ISO"?

lrosenthol:
I'm with @trieloff on this one - we need something standard here.  Why not make the standards the primary value and the CDM one the optional variant?

kstreeter:
@lrosenthol, I think we can make the standard version the preferred one, but we can't make it required, since in general if we are getting data from a system that uses the informal designation then the standard code won't be available.  For interop with these systems we need both.

cdegroot-adobe:
@kstreeter  I have updated to use the *Code. We should be ready to merge. It will support/impact CDM interop.

@saurabhere  Can you please approve this for Campaign

@fmeschbe This is due for the Friday milestone, I believe you are the person doing that activity.

@rkpandey Please approve for Profile.

schemas/common/address.schema.json
```diff
@@ -55,13 +55,18 @@
           "title": "Region",
           "type": "string",
           "description":
-            "The region, county, or district portion of the address."
+            "The region, county, or district portion of the address. This is a free-form field that can have the Region name in any language."
+        },
+        "stateProvince": {
```
saurabhere: `xdm:` prefix is missing?

schemas/common/geo.schema.json
```diff
@@ -20,15 +20,15 @@
           "type": "string",
           "pattern": "^[A-Z]{2}$",
           "description":
-            "The two-character [ISO 3166-1 alpha-2](https://datahub.io/core/country-list) code for the country."
+            "The alpha-2 code of the country as defined by [ISO 3166-1](https://www.iso.org/iso-3166-country-codes.html)."
         },
-        "xdm:stateProvince": {
-          "title": "State or province",
+        "stateProvinceCode": {
```
saurabhere: `xdm:` prefix is missing?

schemas/context/experienceevent.example.1.json
```diff
@@ -88,7 +88,7 @@
     "xdm:geo": {
       "@id": "https://data.adobe.io/entities/geo/tokyo",
       "xdm:countryCode": "JP",
-      "xdm:stateProvince": "JP-13",
+      "xdm:stateProvinceCode": "JP-13",
```
saurabhere: Value needs to be changed too. Validation(`maxLength=2`) did not fail because the field name is missing the prefix in the schema.

schemas/context/profile.example.1.json
```diff
@@ -23,7 +23,7 @@
     "xdm:primary": false,
     "xdm:street1": "345 Park Ave",
     "xdm:city": "San Jose",
-    "xdm:stateProvince": "US-CA",
+    "xdm:stateProvinceCode": "US-CA",
```
saurabhere: Value needs to be changed too.

schemas/context/experienceevent.example.3.json
```diff
@@ -25,7 +25,7 @@
     "xdm:geo": {
       "@id": "https://data.adobe.io/entities/geo/tokyo",
       "xdm:countryCode": "JP",
-      "xdm:stateProvince": "JP-13",
+      "xdm:stateProvinceCode": "JP-13",
```
saurabhere: Value needs to be changed too.

cdegroot-adobe: Correcting this.

schemas/context/profile.example.1.json
cdegroot-adobe: Correcting this.

schemas/context/experienceevent.example.1.json
cdegroot-adobe: Correcting this.

schemas/common/geo.schema.json
cdegroot-adobe: Correcting this.

schemas/common/address.schema.json
cdegroot-adobe: Correcting this.

cdegroot-adobe:
@saurabhere Thanks for the comments, please review the changes. Does this break your usage today?

@kstreeter  Can you please review and see if this meets your expectations. I will create a second PR for the Post Box field as this PR is really about the fixing of stateProvinceCode  and and countryCode

fmeschbe:
@saurabhere I have the impression, this just fell through the cracks. I have now tentatively labeled it for the 0.9.4 drop.

cfraser:
Could we please add the breaking label to this PR? Thanks.

cfraser:
@saurabhere Does this need to be part of v0.9.4 or can it be delayed to v0.9.5?

prabhum2:
Closing this as this more than an year old. Please create a new PR if this is needed

Issue #271, opened 2018-04-21
Add listdetails to analytics extension and move customdimensions&customevents back to experienceevent


jwen-adobe:
## What are the schemas that are affected by the issue

under extensions/adobe/experience/analytics folder:
-schema added: listdetails
-schems removed: customdimensions and customevents
-schema modified: experienceevent

## What are examples of products that are impacted by the issue
platform xdm and tooling.

PR #272, opened 2018-04-21
Add listdetails to analytics extension and move dimensions/events back to experienceevent

jwen-adobe:
This PR links to the issue #271 

@kstreeter @trieloff @lrosenthol @cmathis @cdegroot-adobe 

The following changes have been made in this PR.

1. Add listdetails.schema.json as analytics extension. (This schema was removed originally from context folder during the merge as it was only used by extension, so add it back here as analytics requested).

2.Remove the customdimensions and customevents schema and move their fields back to experienceevent as we did before. This is also analytics requirement to make the tooling generate the similar structure as before.


jwen-adobe:
@trieloff thanks.

PR #275, opened 2018-04-25
Simplify representation of privileges in access control entries

agudian:
Optionally epresenting lists that may possibly contain only one value in non-array notation may look convenient in examples, but adds unnecessary complexity to the code that needs to handle this (both for providers and for clients). We stumbled upon that while implementing a provider of an API using the `accesscontrolentry` type.

I therefor propose to keep it simple and only use the array notation.

@trieloff / @fmeschbe please review / merge.

Btw, the polymorphic representation of the property `principal` (an enum or the complex type `principal`) is also problematic and requires custom mappings in our code and for clients. Plus, it is not translatable in our generated swagger docs, which are used for convenience during testing and by clients to jump-start their code. We don't have a great idea on how to model that differently, though (except pushing down the authenticated/unauthenticated/all to the `principal` type). But maybe keep that in mind for future extensions.


agudian:
> I'd suggest you go one step further and include an update to Contributing.md, so that we have this policy written down.

Good idea, I've pushed another commit for that. Let me know if that covers it or if you had something else in mind.

CONTRIBUTING.md
```diff
@@ -117,6 +117,7 @@ Additional aspects of standard design that aid with consumability are:
 * principle of least astonishment: don't surprise the consumer
 * avoid unnecessary complexity: don't introduce indirections that are not needed
 * the principle of minimal verbosity: make keep things as short as possible, but not shorter
+* avoid unnecessary polymorphism that is hard to consume, e.g. offering both singular and array notations, or introducing type variants that don't provide a common discriminator property.
```
trieloff: +1

fmeschbe:
I agree @agudian, @trieloff. It looks like when I created these schemas, I went ahead doing that. The original schemas only declare the `privilege` property to be a simple array as well.

The difference being that this PR asks for a plural form `repo:privilege**s**` whereas the API specification uses the singular form. Which in turn needs an update on the source, which I have to cross-check first before approving this PR.

lrosenthol:
This all seems good to me!

trieloff:
@fmeschbe @agudian please check the merge I've done in d9ca62d and report back on the resolution of plural vs. singular for `privileges`

agudian:
All good from my side.

fmeschbe:
>  report back on the resolution of plural vs. singular for privileges

We settled on plural `repo:privileges`.

Issue #264, opened 2018-04-13
Add AdCloud extensions to ExperienceEvent


gabebarcelos-adobe:
## What are the schemas that are affected by the issue
ExperienceEvent Schema will need to be augmented in order to support AdCloud's use cases. New schemas will be created in the process

## What are examples of products that are impacted by the issue
AdCloud, Universal Profile

PR #281, opened 2018-04-26
Feature 264 - Viewability: Implement viewability schema

gabebarcelos-adobe:
PR to add viewability schema to core ExperienceEvent. This is related to open issue #264.
@kstreeter  @cdegroot-adobe


schemas/context/viewability-details.schema.json
```diff
@@ -0,0 +1,102 @@
...
+          "type": "integer",
+          "description":
+          "The number of iFrames the ad is displayed inside of.",
+          "minimum": 0
+        },
+        "xdm:viewportHeight": {
+          "title": "Viewport Height",
+          "type": "integer",
+          "description":
+          "The number of vertical pixels of the window the ad is displayed in as measured by the viewability library.",
```
cdegroot-adobe: "the viewability library." is specific to the Adobe implementation, can we express this in a more general way.

schemas/context/viewability-details.schema.json
```diff
@@ -0,0 +1,102 @@
...
+  "description":
+  "Description of viewability measured properties of an advertisement",
+  "definitions": {
+    "viewabilityDetails": {
+      "properties": {
+        "xdm:windowDepth": {
+          "title": "Window Depth",
+          "type": "integer",
+          "description":
+          "The number of iFrames the ad is displayed inside of.",
```
cdegroot-adobe: If this is really iFrame specific I think it would be better to name it related to iFrame. "xdm:windowdepth" can have multiple meaning and once it is used here it has to have the same general meaning(it is the index of the iFrame). Maybe IFrameDepth?

schemas/context/viewability-details.schema.json
```diff
@@ -0,0 +1,102 @@
...
+          "type": "integer",
+          "description":
+          "The number of horizontal pixels of the player as measured by the viewability library. This can be larger than the size of the ad if the player has extra controls or thumbnails.",
+          "minimum": 0
+        },
+        "xdm:playerVolume": {
+          "title": "Player Volume",
+          "type": "integer",
+          "description":
+          "The volume of the player was at video start.",
```
cdegroot-adobe: Is this a percentage? if so can we say so?

Also is mute=0?

schemas/context/viewability-details.schema.json
```diff
@@ -0,0 +1,102 @@
...
+          "minimum": 0,
+          "maximum": 100
+        },
+        "xdm:viewabilityInstrumented": {
+          "title": "Viewabiltity Measurement Instrumented",
+          "type": "boolean",
+          "description":
+          "Whether or not viewability was enabled on for this ad instance."
+        },
+        "xdm:version": {
```
cdegroot-adobe: This needs to be done in the instrumentation schema https://github.com/adobe/xdm/blob/master/extensions/adobe/experience/implementations.schema.json

schemas/context/viewability-details.schema.json
```diff
@@ -0,0 +1,102 @@
...
+          "title": "Viewable Ad",
+          "type": "boolean",
+          "description":
+          "Indicates whether ad was measured as viewable during ad play."
+        },
+        "xdm:activeWindow": {
+          "title": "Active Window",
+          "type": "boolean",
+          "description":
+          "Indicates whether ad was shown on active window on users' browser."
```
cdegroot-adobe: What is the meaning of this on a mobile app? Is there maybe a description that covers non browsers?

lrosenthol:
This should be an extension schema and not part of XDM proper. It is **much** too specific to a particular product and library

kstreeter:
@gabebarcelos-adobe, in response to @lrosenthol's comments, I believe that the definition of digital advertising "viewability" is standardized by [IAB/MRC guidelines](https://www.iab.com/wp-content/uploads/2015/06/MRC-Viewable-Ad-Impression-Measurement-Guideline.pdf), and the properties captured here appear to be what would be required to fulfill those guidelines.  If that's true, I feel like we should keep this as part of core XDM, and perhaps just make sure the descriptions make it clear that this is used to fulfill the guidelines.

gabebarcelos-adobe:
@kstreeter - agreed, these changes implement IAB/MRC guidelines which is why we originally put it under core XDM. I've gone ahead and moved it back to core with the latest push. The descriptions have also been further tweaked to clarify viewability standards, happy to add more details is needed.

schemas/context/viewability-details.schema.json
```diff
@@ -7,18 +7,17 @@
...
   "$id": "https://ns.adobe.com/xdm/context/viewability-details",
   "$schema": "http://json-schema.org/draft-06/schema#",
-  "title": "Viewability",
+  "title": "Viewability Details",
   "type": "object",
-  "description":
-  "Description of viewability measured properties of an advertisement",
+  "description": "Browser or app settings measured at runtime describing the viewability of an advertisement or content associated with each page load.",
   "definitions": {
     "viewabilityDetails": {
```
trieloff: this should be `viewability-details`

schemas/context/viewability-details.schema.json
```diff
@@ -7,18 +7,17 @@
   ],
   "$id": "https://ns.adobe.com/xdm/context/viewability-details",
   "$schema": "http://json-schema.org/draft-06/schema#",
-  "title": "Viewability",
+  "title": "Viewability Details",
   "type": "object",
-  "description":
-  "Description of viewability measured properties of an advertisement",
+  "description": "Browser or app settings measured at runtime describing the viewability of an advertisement or content associated with each page load.",
```
trieloff: This needs way more detail. Probably best to move it into a separate file.

extensions/adobe/experience/implementations.schema.json
```diff
@@ -28,6 +28,11 @@
       "$ref": "https://ns.adobe.com/xdm/context/implementationdetails",
       "description":
         "Client details interacting with Adobe Experience Cloud ID Service."
+    },
+    "https://ns.adobe.com/experience/adcloud": {
+      "title": "Client Implementation for Adobe AdCloud",
```
lrosenthol: No references to specific Adobe products

extensions/adobe/experience/implementations.schema.json
```diff
@@ -28,6 +28,11 @@
...
+      "$ref": "https://ns.adobe.com/xdm/context/implementationdetails",
+      "description": "Client details interacting with Adobe AdCloud."
```
lrosenthol: No references to specific Adobe products

schemas/context/viewability-details.schema.json
```diff
@@ -0,0 +1,102 @@
...
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/context/viewability-details",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Viewability Details",
+  "type": "object",
+  "description": "Browser or app settings measured at runtime describing the viewability of an advertisement or content associated with each page load.",
+  "definitions": {
+    "viewabilityDetails": {
```
lrosenthol: this should be "adViewabilityDetails"., since this is clearly about ads?

schemas/context/viewability-details.schema.json
```diff
@@ -0,0 +1,102 @@
...
+      "properties": {
+        "xdm:adUnitDepth": {
+          "title": "Ad Unit Depth",
+          "type": "integer",
+          "description":
+          "Publishers may embed ad units inside containers/iFrames in order limit the ad's access solely to the code of the page. This describes how many 'containers' or iFrames the ad unit is displayed inside of.",
```
lrosenthol: What about ads used in other contexts than the web?

schemas/context/viewability-details.schema.json
```diff
@@ -0,0 +1,102 @@
...
+          "minimum": 0
+        },
+        "xdm:viewportHeight": {
+          "title": "Viewport Height",
+          "type": "integer",
+          "description":
+          "The number of vertical pixels of the window the ad is displayed in, measured at runtime.",
```
lrosenthol: What about non-pixel-based media?

schemas/context/viewability-details.schema.json
```diff
@@ -0,0 +1,102 @@
...
+          "minimum": 0
+
+        },
+        "xdm:viewportWidth": {
+          "title": "Viewport Width",
+          "type": "integer",
+          "description":
+          "The number of horizontal pixels of the window the ad is displayed in, measured at runtime.",
```
lrosenthol: What about non-pixel-based media?

schemas/context/viewability-details.schema.json
```diff
@@ -0,0 +1,102 @@
...
+          "minimum": 0
+        },
+        "xdm:adHeight": {
+          "title": "Ad Height",
+          "type": "integer",
+          "description":
+          "The number of vertical pixels of the player, measured at runtime. This can be larger than the size of the ad if the player has extra controls or thumbnails.",
```
lrosenthol: What about non-pixel-based media?

schemas/context/viewability-details.schema.json
```diff
@@ -0,0 +1,102 @@
...
+          "minimum": 0
+        },
+        "xdm:adWidth": {
+          "title": "Ad Width",
+          "type": "integer",
+          "description":
+          "The number of horizontal pixels of the player, measured at runtime. This can be larger than the size of the ad if the player has extra controls or thumbnails.",
```
lrosenthol: What about non-pixel-based media?

schemas/context/viewability-details.schema.json
```diff
@@ -0,0 +1,102 @@
...
+          "title": "Active Window",
+          "type": "boolean",
+          "description":
+          "Indicates whether ad was shown on active window on users' device."
+        },
+        "xdm:percentViewable": {
+          "title": "Percent Viewable",
+          "type": "integer",
+          "description":
+          "Percent of pixels in the ad deemed viewable at measurement time.",
```
lrosenthol: What about non-pixel-based media?

extensions/adobe/experience/implementations.schema.json
trieloff: Should be ok in this context. We have a couple of other Adobe-specific references here, too.

schemas/context/viewability-details.schema.json
gabebarcelos-adobe: MRC only defines this on web based ads. At the moment, mobile apps/ott/ctv/etc are not eligible for viewability measurements/standard.

schemas/context/viewability-details.schema.json
gabebarcelos-adobe: In this exact situation yes, this is ad centric. However, nothing stops viewability measurements to be collected for content rather than ads. I would argue for reusability sake we should keep as it medium agnostic.

schemas/context/viewability-details.schema.json
gabebarcelos-adobe: IAB and MRC guidelines are only defined for pixel-based media. Following industry standard guidelines for consistency.

schemas/context/viewability-details.schema.json
gabebarcelos-adobe: IAB and MRC guidelines are only defined for pixel-based media. Following industry standard guidelines for consistency.

schemas/context/viewability-details.schema.json
gabebarcelos-adobe: IAB and MRC guidelines are only defined for pixel-based media. Following industry standard guidelines for consistency.

schemas/context/viewability-details.schema.json
gabebarcelos-adobe: IAB and MRC guidelines are only defined for pixel-based media. Following industry standard guidelines for consistency.

schemas/context/viewability-details.schema.json
gabebarcelos-adobe: IAB and MRC guidelines are only defined for pixel-based media. Following industry standard guidelines for consistency.

extensions/adobe/experience/implementations.schema.json
gabebarcelos-adobe: Other adobe-specific references exist here. Should we move all of them out then?

schemas/context/viewability-details.schema.json
gabebarcelos-adobe: Added viewability-details.description.md

schemas/context/viewability-details.schema.json
gabebarcelos-adobe: Thanks for pointing this out, also corrected allOf node

schemas/context/viewability-details.schema.json
lrosenthol: So then if this whole schema is based on this MRC work - let's make it clear in the description right up front (including the link to the spec).  I am tempted to also want it in the name - but that's overkill.

schemas/context/viewability-details.schema.json
lrosenthol: but even HTML doesn't require the use of pixels.  You can use a variety of other measurement units via CSS.  Or you can use SVG for your ads, which also has no pixels measurements.

schemas/context/viewability-details.schema.json
lrosenthol: See previous comment on HTML, CSS and SVG

schemas/context/viewability-details.schema.json
lrosenthol: See previous comment on HTML, CSS and SVG

schemas/context/viewability-details.schema.json
lrosenthol: See previous comment on HTML, CSS and SVG

schemas/context/viewability-details.schema.json
lrosenthol: GREAT description, especially the reference to the spec.  Thanks!

extensions/adobe/experience/implementations.schema.json
lrosenthol: Didn't realize this was inside an extension schema - this is fine.

extensions/adobe/experience/implementations.schema.json
lrosenthol: Didn't realize this was inside an extension schema - this is fine.

schemas/context/viewability-details.schema.json
lrosenthol: Saw the new description file - VERY helpful, thanks!

schemas/context/viewability-details.schema.json
```diff
@@ -0,0 +1,102 @@
+{
+  "meta:license": [
+    "Copyright 2018 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/context/viewability-details",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Viewability Details",
```
trieloff: Maybe we should change the title (just the title, not the schema name) to "MRC Viewability"

schemas/context/viewability-details.description.md
```diff
@@ -0,0 +1,39 @@
+Browser or app environment measured at runtime describing the viewability of an advertisement or content associated with each page load.
```
trieloff: We should probably mention that this relates to display advertising only.

schemas/context/environment.schema.json
```diff
@@ -32,6 +32,12 @@
           "description":
             "The browser specific details such as browser name, version, javascript version, user agent string, accept language."
         },
+        "xdm:viewabilityDetails": {
```
trieloff: I don't think the name `Details` carries too much information or clarifies the intent vs. `xdm:viewability`. Maybe `xdm:adViewability` makes it clear enough.

trieloff: Also, from the examples it is not entirely clear what ad unit this refers to.

schemas/context/viewability-details.example.1.json
```diff
@@ -0,0 +1,16 @@
+{
+  "xdm:adUnitDepth": 0,
+  "xdm:viewportHeight": 1250,
+  "xdm:viewportWidth": 1600,
+  "xdm:adHeight": 250,
+  "xdm:adWidth": 300,
+  "xdm:playerVolume": 85,
+  "xdm:measurementEligible": true,
+  "xdm:implementationDetails" : {
```
trieloff: Paging @cdegroot-adobe 

This is the first time that we are using the `implementationdetails` schema. Should we have one `xdm:implementations` array for the whole experience event, or should we allow scattered pointers to `implementationdetails` in specific places?

schemas/context/viewability-details.schema.json
trieloff: I agree. Great jon, @gabebarcelos-adobe

schemas/context/viewability-details.schema.json
gabebarcelos-adobe: For viewport measurement, browser js standard is to always return a pixel count by definition. This is analogous to window.innerHeight (https://developer.mozilla.org/en-US/docs/Web/API/Window/innerHeight)

schemas/context/viewability-details.schema.json
gabebarcelos-adobe: For viewport measurement, the browser standard is to always return a pixel count by definition. This is analogous to window.innerHeight (https://developer.mozilla.org/en-US/docs/Web/API/Window/innerHeight)

schemas/context/viewability-details.schema.json
gabebarcelos-adobe: This is also a JS measured field which quantifies size by pixel rather than any relative measures such as em, %, etc. Should always be in pixels.

schemas/context/environment.schema.json
gabebarcelos-adobe: Changed name of schema.

As far as what ad unit, a viewability schema will be tied to a advertisement schema with PR 282. A single experience event should only describe a single ad, thus the ad is implied by its' other contexts.

schemas/context/viewability-details.schema.json
lrosenthol: The correct standard is the CSS View module, which identifies the measurement as CSS Pixels (https://www.w3.org/TR/cssom-view-1/#css-pixels).  So let's go ahead and use that correct terminology and then I'm good. Thanks!

schemas/context/viewability-details.schema.json
lrosenthol: The correct standard is the CSS View module, which identifies the measurement as CSS Pixels (https://www.w3.org/TR/cssom-view-1/#css-pixels).  So let's go ahead and use that correct terminology and then I'm good. Thanks!

schemas/context/viewability-details.schema.json
lrosenthol: The correct standard is the CSS View module, which identifies the measurement as CSS Pixels (https://www.w3.org/TR/cssom-view-1/#css-pixels).  So let's go ahead and use that correct terminology and then I'm good. Thanks!

schemas/context/viewability-details.schema.json
lrosenthol: The correct standard is the CSS View module, which identifies the measurement as CSS Pixels (https://www.w3.org/TR/cssom-view-1/#css-pixels).  So let's go ahead and use that correct terminology and then I'm good. Thanks!

schemas/context/viewability-details.schema.json
lrosenthol: If this is a percentage, can you specify min:0, max:100, please?  Otherwise,  I am good.

schemas/context/viewability-details.schema.json
gabebarcelos-adobe: I'm not sure I follow, what would we be changing? The description?

schemas/context/viewability-details.schema.json
gabebarcelos-adobe: The max and min values should be there already under the description field.

extensions/adobe/experience/implementations.schema.json
cdegroot-adobe: I believe all concerns are now addressed from these comments.

extensions/adobe/experience/implementations.schema.json
cdegroot-adobe: I believe all concerns are now addressed from these comments.

kstreeter:
@gabebarcelos-adobe I've got one outstanding comment on the experience event example: it has a "version" field which does not exist in the schema.

gabebarcelos-adobe:
@kstreeter - Thanks for pointing that out, i've updated the example file to align with the schema.

kstreeter:
Hi @lrosenthol when you have confirmed your comments have been addressed we'll merge this in.

cdegroot-adobe:
@lrosenthol - It has been 7 days since you were requested to confirm your comments were addressed. Please do so as we need to merge in this body of schema.

Issue #250, opened 2018-04-12
Channels in Experience Event does not allow extension well and needs to be more data oriented.


cdegroot-adobe:
## What are the schemas that are affected by the issue
https://github.com/adobe/xdm/tree/master/schemas/channels/*.*.*
https://github.com/adobe/xdm/blob/master/schemas/context/experienceevent.schema.json

## What are examples of products that are impacted by the issue
All digital marketing solutions that use channel data.

## Issue
Currently the implementation is highly prescriptive:
- There is a field in ExperiencEvent that is of type channel. So it does not work as an extension point for custom extensions of channels.
- The channel schemas describe the channel types, so the data is not self describing, a service needs to inspect the schema to understand the channel type.

## Solution
1. ExperienceEvent `xdm:channels` changes to a dictionary 
1. The new `xdm:channels` dictionary adds a field for `xdm:type` that is the generalized communications channel type (web, mobile, push)
1. The new `xdm:channels` dictionary adds a field for `xdm:strategy` (paid, earned, owned)
1. The new `xdm:channels` dictionary adds a field for `xdm:marketing` with the name space of Analytics to cover marketing channels.
1. The new `xdm:channels` dictionary adds a field for `xdm:technical` that is the same as the current `xdm:channels` field (points to instances of https://github.com/adobe/xdm/tree/master/schemas/channels/*.*.* but is named).
1. The  https://github.com/adobe/xdm/tree/master/schemas/channels/*.*.* change to be https://github.com/adobe/xdm/tree/master/schemas/technicalchannels/*.*.* .
1. the entry currently in channels for 'Agency' gets deleted. That is not a technical channel, but is a marketing channel and is is certainly not unique in the https://github.com/adobe/xdm/tree/master/schemas/technicalchannels/*.*.* for example an Agency could use a web channel.

trieloff:
I need to understand this better to judge it.

- What is the difference between a marketing channel and a technical channel?
- What does it mean if an experience event is happening in multiple channels?
- I like the ability to distinguish between paid, owned, and earned – is `strategy` the right term, however?
- 👍 on clearing up the Agency channel.

cdegroot-adobe:
@trieloff 
Channels are broadly used is multiple ways are once. So there always must be a distinguishing refining term. Individual customers will normally have additional channel groupings that they will also use.

Channels `types` are the largest grain definition of technical channels and are often used in "cross channel" query cases to select a subset of channel types. "web and e-mail" for example.

[...]

PR #289, opened 2018-05-02
Make xdm:channel a reference

trieloff:
This fixes the issues outlined in #250.

**Note**: this is a breaking change to 0.9.0


trieloff:
@kstreeter @cdegroot-adobe any feedback?

cdegroot-adobe:
@kstreeter  I believe this is a simple fix to the issue we have with the current channels representation and it gives us the ability to innovate more in future. I propose we take this solution now and then consider additions later.

PR #290, opened 2018-05-02
Improve Documentation of Schema Attributes

trieloff:
adobe/jsonschema2md#60 and others add support for linking schema attributes in the generated documentation to sections in the narrative documentation. This PR provides descriptions for each supported attribute and links from the generated docs.

To keep the PR small, I have not included any of the updates in `docs/reference` which will be generated automatically upon merge into master.


docs/status.md
```diff
@@ -0,0 +1,11 @@
+## Schema Status
+
+The status of each schema in XDM describes the likelihood of change of a schema.
+In particular for XDM Versions prior to XDM 1.0.0, this can inform users and providers of XDM-compatible APIs of the stability of a given schema.
+
```
cdegroot-adobe: Having this at the schema level is fine, do we need to also have this at the field level when it deviates from the schema's "status"? For example a new field added to a stable schema but that field would not yet be stable.

trieloff: I agree. #303 covers parts of that.

PR #291, opened 2018-05-02
Update copyright-owner.schema.json

trieloff:
No changes since Mon Mar 12 2018 19:06:43 GMT+0000 (UTC)


lrosenthol:
This item is already covered in https://wiki.corp.adobe.com/display/AdobeCloudPlatform/Core+Asset+Data+Model and our core schemas.

trieloff:
@lrosenthol does this mean we should leave it at experimental? What is the path to stabilizing it?

lrosenthol:
no, kill it as we don't need it.  copyright_owner is already part of the core asset schema.

trieloff:
@lrosenthol you might be seeing something that I don't see, but the Copyright Owner schema got introduced into XDM via the Core Asset Data Model, i.e. the file `copyright-owner.schema.json` is the expression of that part of the Core Asset Data Model.

It is not expressed anywhere else, so if we remove it from XDM, it would no longer be formally part of XDM and introduce an inconsistency with the Core Asset Data model.

lrosenthol:
I see - sorry about that!   Approving this.

PR #295, opened 2018-05-02
Stabilize rectangular

trieloff:
No update since Mon Mar 12 2018 19:06:43 GMT+0000 (UTC)


lrosenthol:
except that it is using terms like imageWidth and referring to values in pixels - which don't apply to numerous uses for a "rectangular measurement", even in images!

trieloff:
We also have a rectangular schema that allows for variable units. The imageWidth and imageLength properties are inherited from TIFF.

Is there another name you'd propose?

lrosenthol:
I have to think about it...

trieloff:
@lrosenthol ping?

lrosenthol:
So this only represents a rectangular area on a pixel-based surface, but without any indication of origin.  Is that is what is intended, then rename accordingly and go for it.  Otherwise, we continue to need a LOT of additional stuff here...

trieloff:
I've updated Schema Title and Description.

Issue #301, opened 2018-05-03
Change Campaign resource name to handle protobuf parsing


saurabhere:
## What are the schemas that are affected by the issue
extensions/adobe/experience/campaign/experienceevent.schema.json

## What are examples of products that are impacted by the issue
Adobe Campaign

According to @cmathis  "One of the limitations of Protobuf is having one object/message named “campaign” contain another object named “campaign” as well. "
This is because the uri of the campaign resource becomes "https://ns.adobe.com/adobe/experience/campaign/campaign". 

This can be avoided by changing the name of the campaign resource from `campaign` to `marketingCampaign`.

lrosenthol:
Or stop using Protobuf....
(I have no objections to the rename)

PR #284, opened 2018-04-30
Update CONTRIBUTING with guidance on using namespaces

kstreeter:
This is a first PR to work on #277 by adding some basic guidelines for use of namespaces versus nested structure in our schemas.  When we agree on the basics I will work on posting PRs for some of the specific issues identified so far.

Issue #277, opened 2018-04-26
Improve consistency and provide guidelines for use of nested JSON structure versus namespaces


kstreeter:
I think we need to develop more consistent guidance on when it is appropriate to create new namespaces, and when nested JSON structures should be used instead.  We use both ways to organize things in different schemas, and many schemas mix both methods at the same time.

While this could be considered a matter of style, as we have continued to use these schemas in the data platform we have found that the practice of mixing methods creates significant usability issues.  For example, UI tools that create or consume XDM (like data pickers) are generally implemented as tree widgets.  What does the widget show?  Does it show the the JSON nested structure?  Does it show the namespace structure?  Both...sometimes or always?

I feel we need to be much more prescriptive about namespaces.  We had previously agreed to limit the number of prefixed namespaces, saying only major areas should have a prefix and those areas should then be consistent internally.  But in many places (channels, metrics, Analytics and Campaign extensions) we have introduced many namespaces in full URIs.  So I don't think we are really following the policy of "namespace moderation" we had discussed early on.

I have a proposal for clear guidelines on when to use namespaces versus when to use structure:

  * Namespaces are *not* used to organize data.  They are used to prevent name conflicts.
  * Name conflicts occur across organization and domain boundaries.  We should have namespaces to keep different organizations or groups working in parallel.
  * Within an organization or domain, there should only be *one* namespace.  JSON structure or more explicit naming of fields should be used to differentiate names.

If we adopt this, our tools will in most cases be able to drop the namespaces completely.  Conflicts will happen of course, but they will be rare cases that can handled in the tools in ways that may be less user-friendly (for example, showing the URI in a detail box or popup).

To implement this, we need to:

 * Revisit the namespaces used in channels, metrics, and the extensions.
 * Add a section on these guidelines to contributing.  

## What are the schemas that are affected by the issue

All, but this issue is particularly apparent on the definitions of channels, of the definitions of metrics, and in the Analytics and Campaign extensions.

## What are examples of products that are impacted by the issue

Campaign, Analytics

trieloff:
I think it would help to point out a couple of schemas where this is a problem today. 

In some of the places you've mentioned, we are using URIs not as namespaces per se, but for entity identification. 

In cases where we have many entities, it seems to be a simpler and more consistent way to use URIs everywhere instead of having a mix where some areas use URIs and others use simple names.

[...]

PR #302, opened 2018-05-03
Campaign ExperienceEvent extension tweaks.

saurabhere:
*Make Campaign link Label and Category explicit fields.
*Change Campaign resource name to `marketingcampaign` to handle protobuf parsing.
*Fix examples.

Fixes issue #301 

Out for review: @cmathis @trieloff @cdegroot-adobe

### DIFF
--- a/extensions/adobe/experience/campaign/experienceevent.schema.json
+++ b/extensions/adobe/experience/campaign/experienceevent.schema.json
@@ -146,7 +146,7 @@
           "type": "integer",
           "description": "The campaign activity originating this message."
         },
-        "https://ns.adobe.com/experience/campaign/campaign/id": {
+        "https://ns.adobe.com/experience/campaign/marketingCampaign/id": {
           "title": "CampaignID",
           "type": "integer",
           "description":
@@ -207,14 +207,15 @@
           "minimum": 1,
           "maximum": 32767
         },
-        "https://ns.adobe.com/experience/campaign/link/tags": {
-          "title": "Tags",
-          "type": "array",
-          "description":
-            "A list of tags qualifying the link. The URL category is one of those tags.",
-          "items": {
-            "type": "string"
-          }
+        "https://ns.adobe.com/experience/campaign/link/label": {
+          "title": "URL Label",
+          "type": "string",
+          "description": "The user-friendly label clicking on which the link opens."
+        },
+        "https://ns.adobe.com/experience/campaign/link/category": {
+          "title": "URL Category",
+          "type": "string",
+          "description": "The category of the link. It may be `subscription`, or a user-defined category."
         },
         "https://ns.adobe.com/experience/campaign/message/external/id": {
           "title": "External ID",
--- a/extensions/adobe/experience/campaign/experienceevent.example.click.json
+++ b/extensions/adobe/experience/campaign/experienceevent.example.click.json
@@ -5,6 +5,23 @@
       "xdm:value": 1
     }
   },
+  "xdm:channel": "https://ns.adobe.com/xdm/channels/email",
+  "xdm:endUserIDs": {
+    "https://ns.adobe.com/experience/mcid": {
+      "@id": "https://data.adobe.io/entities/identity/92312748749128",
+      "xdm:namespace": {
+        "@id": "https://data.adobe.io/entities/namespace/4",
+        "xdm:code": "ECID"
+      }
+    },
+    "https://ns.adobe.com/experience/campaign": {
+      "@id": "https://data.adobe.io/entities/identity/1234567",
+      "xdm:namespace": {
+        "@id": "https://data.adobe.io/entities/namespace/2756",
+        "xdm:code": "CPGN"
+      }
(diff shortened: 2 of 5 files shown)


extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -146,7 +146,7 @@
           "type": "integer",
           "description": "The campaign activity originating this message."
         },
-        "https://ns.adobe.com/experience/campaign/campaign/id": {
+        "https://ns.adobe.com/experience/campaign/marketingCampaign/id": {
```
trieloff: This would work better as `https://ns.adobe.com/experience/campaign/marketingCampaignID`, see #277 and #284 for details.

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -207,14 +207,15 @@
...
         },
-        "https://ns.adobe.com/experience/campaign/link/tags": {
-          "title": "Tags",
-          "type": "array",
-          "description":
-            "A list of tags qualifying the link. The URL category is one of those tags.",
-          "items": {
-            "type": "string"
-          }
+        "https://ns.adobe.com/experience/campaign/link/label": {
```
trieloff: This would work better as `https://ns.adobe.com/experience/campaign/linkLabel`, see #277 and #284 for details.

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -207,14 +207,15 @@
...
+          "title": "URL Label",
+          "type": "string",
+          "description": "The user-friendly label clicking on which the link opens."
+        },
+        "https://ns.adobe.com/experience/campaign/link/category": {
```
trieloff: This would work better as `https://ns.adobe.com/experience/campaign/linkCategory`, see #277 and #284 for details.

saurabhere:
@trieloff As I understand, eventually we'd like to rename all these fields under the common namespace "https://ns.adobe.com/experience/campaign" and then ref to the entities in the separate schemas they are part of. In this case, it would lead to a new schema `extensions/adobe/experience/campaign/link.schema` which would have all these `properties` which will be $ref'ed by these properties. 
If I change this here, that PR would lose context on which fields are part of the `link` schema. No?

cmathis:
@trieloff - Would something like the following be allowed:

```
properties: {
   "https://ns.adobe.com/experience/campaign/link": {
       "type": "object",
       "properties": {
              "label": { //Alternatively this could be xdm:label
              },
              "category": {
              }
        }
}
```

Is it necessary to repeat the namespace for those unlined embedded object fields. What we want to do is to avoid making the data files and fields path look like the following. https://ns.adobe.com/experience/campaign/link**.**https://ns.adobe.com/experience/campaign/label

The final schema and field path needs to be very simple and look something like adobe.experience.campaign.link.label. 
and
adobe.experience.campaign.link.category

trieloff:
@cmathis I'd say yes, having plain names below a fully namespaced object is permitted, but I'd like to get @kstreeter's point of view.

You should be aware that it takes away second-level extensibility, i.e. nobody else would be able to extend this particular object.

trieloff:
@saurabhere creating a `link.schema.json` can have multiple implications:

- if you create it as an abstract schema and just `$ref` or mix it into your existing schema, the property names need to be unique enough so that you don't mix up the "campaign ID" and the "link ID"
- if you create it as a concrete schema and use it as the type of the `link` object, then even a short name like `https://ns.adobe.com/experience/campaign/id` could work.

saurabhere:
@trieloff @cmathis OK, this PR was not about this conversion, but just wanted to add 2 new fields but seeing a need for the change akin to #306 , I will change the schema as well. So, don't review it yet.

saurabhere:
@trieloff @cmathis @kstreeter Out for review.

kstreeter:
This looks really good to me @saurabhere.  The only additional comment I had was around the use of `@id`.   There are a couple of places where this is used with a property of type integer.  `@id` is actually a JSON-LD concept as is defined as a URI, so these would either have to change to a string with format='uri' or be given a different name (like deliveryID, etc.).

saurabhere:
@kstreeter does a simple `id` work? I see it being used in the `user.schema`
I am not sure if that would break the convension of prepending `xdm:` to all fields.

kstreeter:
Hey @saurabhere, yes you could just use xdm:id.  It kind of depends on what the intention of the field is.  If you want to define a Campaign ID primarily for the purpose of identifying that campaign across systems, I would go with redefining it to be an `@id` URI, as that is how this generally is done across XDM.  On the other hand, if the primary purpose is to connect the data to a specific system-generated ID (which is an integer) then it would be appropriate to just call it xdm:id and make it an integer.

saurabhere:
Thanks @kstreeter . Fixed.

saurabhere:
@trieloff good to go?

PR #306, opened 2018-05-04
Rearrange analytics extension fields to use nested structure

kstreeter:
This PR moves the Analytics extension fields into nested objects, versus having them organized with distinct namespaces.

Please link to the issue #…

This is one of the issues described in #277.


kstreeter:
@cmathis, @jwen-adobe can you check this out and make sure it continues to work with our tooling?

jwen-adobe:
@kstreeter It works well with the tooling.

cdegroot-adobe:
Looking at this, one aspect we are realizing is that extensions should generally be organized into grouping fields in the hierarchy rather then over populating the root nodes of the schemas they are extending.

cmathis:
We should do this same thing to https://github.com/adobe/xdm/pull/302.

Issue #79, opened 2018-01-19
Experience Cloud Product Description needs work


trieloff:
I'm working on the [product](https://github.com/adobe/xdm/blob/experience-cloud-staging/schemas/content/product.schema.json) schema for experience cloud right now and I need some help in making sure the descriptions are accurate.

1. Is this the schema for a product variant or a master product?
2. What is the difference between a product variant and a master product?
3. Are these the only product relationships that matter?
4. Should `product` be `auditable`, i.e. would it make sense to include the date when the product record was created or last modified?
5. Should `category`, `department`, and `fabrication` be controlled vocabularies or free text as they are right now?
6. Given that the name and description is to be localized, would it make sense to include the language in the schema?

mhaack:
Looking at the product schema with the commerce hat on there are some more things which I think require some clarification/adjustment:

7. attributes like `size`, `gender`, `brand` etc should not be hard defined and moved to an attribute list
8. attributes must support meta data (like units)
9. attributes must be localized

trieloff:
@mhaack thanks for the feedback. 

> attributes like `size`, `gender`, `brand` etc should not be hard defined and moved to an attribute list

Attributes that are common enough *should be* hard defined, although a separation of customer-visible and customer invisible attributes might make sense. We do have the ability to define custom attributes through the [XDM extensibility mechanism](https://github.com/adobe/xdm/blob/master/docs/extensions.md), but would benefit from defining things that are common to enough products. Right now, it looks very fashion-centric, and we might come up with a more extensive list of attributes that are common.

> attributes must support meta data (like units)

Do you think the approach we have taken for [Metrics](https://github.com/adobe/xdm/blob/master/docs/reference/data/metric.schema.md) should be copied here?

> attributes must be localized

This is tying back to two questions:

A. What is the difference between a master product and a product variant? One viewpoint could be that every product variant is localized in exactly one language.
B. #4 – @lrosenthol proposed a solution, not for localizing XDM itself, but for localized expressions within XDM

mhaack:
> Attributes that are common enough should be hard defined, although a separation of customer-visible and customer invisible attributes might make sense. 

I agree, using a flexible attribute list to me makes most sense here. That is also the approach almost all commerce vendors follow.
Regarding customer-visible vs. customer invisible, this is any important use case I agree. I think this could be done as a meta information setting to flag each attribute. But I would not make it mandatory. 

[...]

PR #310, opened 2018-05-08
Updated version of commerce product

mhaack:
Proposal for an updated more generic and flexible product XDM for #79 

The new proposal is align with the Adobe Commerce Integration Framework product data model (https://git.corp.adobe.com/pages/CQ/commerce-ccif-api/swagger.json) and was reviewed and mapped to the data structures used by the major commerce vendors (Hybris, IBM WebSphere Commerce, Salesforce/Demandware, Magento, commercetools, Shopify).


cdegroot-adobe:
@mhaack - Thanks for the contribution. 

I agree with adding the attribute concepts. 

One critical concern I have is that the design of this schema provides very little standardization of data. Besides `@id`, SKU and name everything else is completely dynamic. The side effect of this is that it becomes slower to query and is less guiding to customers. 
We need to consider two things:
1. When there are very very commonly used fields, how can we make them first class citizens int he schema, rather than just 100% dynamic. Examples like this would be "department", manufacturer, fabrication. But there are likely several more.
2. We cannot have customers decide common keys all by themselves, we need to make a recommendation on certain naming of attributes like 'color', 'size' etc or else this proposal is not helping add semantic meaning to our customers data.

I look forward to your next update, very much appreciated C.

schemas/content/product-attribute.schema.json
```diff
@@ -0,0 +1,52 @@
...
+        "schema:name": {
+          "title": "Name",
+          "type": "string",
+          "description": "The human-readable name of the attribute."
+        },
+        "xdm:value": {
+          "title": "Value",
+          "description": "The attribute value."
+        },
+        "xdm:unit": {
```
cdegroot-adobe: Should be an enumeration of common units.

schemas/content/product-attribute.schema.json
```diff
@@ -0,0 +1,52 @@
...
+          "title": "Key",
+          "type": "string",
+          "description": "The technical key / identifier of this attribute."
+        },
+        "schema:name": {
+          "title": "Name",
+          "type": "string",
+          "description": "The human-readable name of the attribute."
+        },
+        "xdm:value": {
```
cdegroot-adobe: This has no type. That is not allowed.

schemas/content/product-attribute.schema.json
```diff
@@ -0,0 +1,52 @@
...
+        "xdm:unit": {
+          "title": "Unit",
+          "type": "string",
+          "description": "The unit of measure for the attribute value."
+        },
+        "xdm:variantAttribute": {
+          "title": "Variant Attribute",
+          "type": "boolean",
+          "default": false,
+          "description": "Whether this attribute is a defining variant attribute which is used to distinguishe different product variants from each other."
```
cdegroot-adobe: How does a boolean of true for being a variant "distinguishe different product variants from each other."

Typo: distinguish

schemas/content/product.schema.json
```diff
@@ -17,9 +18,8 @@
         "@id": {
           "title": "Identifier",
           "type": "string",
-          "format": "uri",
```
cdegroot-adobe: @id if used must be a URI

schemas/content/product.schema.json
```diff
@@ -17,9 +18,8 @@
...
           "description":
-            "The internal unique ID of the variant in the commerce backend system."
+            "The internal unique ID of the product or variant in the commerce backend system."
```
cdegroot-adobe: "commerce backend system" is not always right. t could be from a PIM etc. Best to not be that explicit.

schemas/content/product.schema.json
```diff
@@ -60,81 +53,68 @@
           "description":
             "The internal unique ID of the product in the commerce backend system."
         },
-        "xdm:masterProductSKU": {
-          "title": "Master Product SKU",
-          "type": "string",
+        "xdm:prices": {
```
cdegroot-adobe: Best to make this a abstract schema so it can be easily reused and extended.

schemas/content/product.schema.json
```diff
@@ -60,81 +53,68 @@
...
+                "pattern": "^[A-Z]{3}$",
+                "description":
+                  "The ISO 4217 alphabetic currency code used for cost and pricing."
+              },
+              "amount": {
+                "title": "Amount",
+                "type": "number",
+                "minimum": 0,
+                "description":
+                  "The amount value for that price."
```
cdegroot-adobe: strange wording to my eye.

schemas/content/product.schema.json
```diff
@@ -60,81 +53,68 @@
...
+              },
+              "type": {
+                "title": "Price",
```
cdegroot-adobe: "Price Type" maybe a better title.

schemas/content/product.schema.json
cdegroot-adobe: Does the price structure also need a GEO property?

schemas/content/product.schema.json
```diff
@@ -60,81 +53,68 @@
...
+              "https://ns.adobe.com/xdm/content/product-attribute#/definitions/attribute"
+          },
           "description":
-            "The ISO 4217 alphabetic currency code used for cost and pricing, including `listPrice` and `COGS`."
-        },
-        "xdm:originalSaleDate": {
-          "title": "Original Sale Date",
-          "type": "string",
-          "format": "date",
+            "List of attributes associated with this product/variant."
```
cdegroot-adobe: We seem a bit inconsistent in the descriptions where we refer to product, product/variant etc. Can you do a pass please and ensure we are consistent please.

schemas/content/product.schema.json
```diff
@@ -155,16 +135,6 @@
           "type": "string",
           "description":
             "The URL for the primary Product View of the product variant page."
-        },
-        "xdm:manufacturerName": {
```
cdegroot-adobe: This proposal is being far to open, we know that many customers have to represent simular concepts for attributes and variants. But we are not giving them guidance on how to be consistent in the schema. This must be improved.

schemas/content/product-attribute.schema.json
```diff
@@ -0,0 +1,52 @@
...
+  "$id": "https://ns.adobe.com/xdm/content/product-attribute",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Product Attribute",
+  "type": "object",
+  "meta:extensible": true,
+  "description": "Generic product attribute of a product in the product catalog.",
+  "definitions": {
+    "attribute": {
+      "properties": {
+        "xdm:key": {
```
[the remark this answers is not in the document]
mhaack: I agree with giving more context and guidance. Can an enumeration be open, since we can not foresee all possible attributes.

schemas/content/product.schema.json
mhaack: Maybe, but then we instead of GEO we should be more generic and add "channel", "scope" or "context".  Prices could be per region, channel, customer group, location etc.

schemas/content/product-attribute.schema.json
mhaack: Can we define "object" as type. Product variants can be strings, numbers, dates, etc. thats why I left this out.

schemas/content/product.schema.json
mhaack: Ok, I will convert this into an abstract schema. Which leads to the follow up questions:

* should `product-attribute`schema be abstract as well? I think so yes.
* should the abstract schemas have `product-` prefix like product-attribute, product-price or no prefix?
* what is the best location of these sub schemas used by product? Still be `content` or better `data`?

schemas/content/product-attribute.schema.json
cdegroot-adobe: See: https://github.com/adobe/xdm/issues/334

We need to get a decision on how we want to implement open enums.

schemas/content/product-attribute.schema.json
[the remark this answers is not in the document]
cdegroot-adobe: See: #334

schemas/content/product-attribute.schema.json
mhaack: Ok I think we need to wait for #334 then.

schemas/content/product-attribute.schema.json
mhaack: Since "object" is no valid type. How can we make this generic? Since defining this as string for example is to short.

schemas/content/product.schema.json
mhaack: Already updated the description.

schemas/content/product.schema.json
mhaack: You mean `amount`, it is widely used for price or cost objects in commerce. A price always has currency and value/amount/number value.

schemas/content/product-attribute.schema.json
mhaack: A true/false marker should be enough. For example for a shirt color & size attributes this is set to true, for attributes like brand or fabrication it is not.
This additional information for color & size give an indication the client that these attributes will influence the product variant. While brand for example has only a descriptive character.

Clients can for example display these as selections boxes and hence has influence on the "selected" SKU.

schemas/content/product-attribute.schema.json
mhaack: I added `"type": "string"` for now, but I'm not really happy with it. I think there should be a type=object or type being optional, to have the flexibility to have product attributes with different types like number, date, etc.

schemas/content/product-attribute.schema.json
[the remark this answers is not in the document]
trieloff: Yeah, don't do this. This is the "bag of key-values" anti pattern spelled out. You can do better by just defining a `patternProperties` object with `://` (regex), so that any URL is a valid extension, but I second @cdegroot-adobe's point that we should standardize as much as we can.

schemas/content/product-attribute.schema.json
trieloff: You can `oneOf` multiple `type`s.

schemas/content/product-attribute.schema.json
trieloff: `xdm:unit` is fine for now, I'd standardize the list of known units here: #358

schemas/content/product.example.1.json
```diff
@@ -1,21 +1,53 @@
 {
-  "@id": "https://commerce.adobe.io/entities/product/product-203766910",
+  "@id": "product-203766910",
```
trieloff: No: `@id` must be a URI.

schemas/content/product.schema.json
trieloff: Also, it's much easier to turn any string into a URI than to provide the right context for a string ID to make it resolvable.

schemas/content/product.schema.json
```diff
@@ -37,21 +37,14 @@
           "type": "string",
           "description": "The localized description of the product."
         },
-        "xdm:category": {
+        "xdm:categories": {
           "title": "Category",
-          "type": "string",
+          "type": "array",
+          "items": {
+            "type": "string"
```
trieloff: This is hard to resolve. If your IDs are plain strings, you have a black-box API that does not allow resolution of metadata of a category.

schemas/content/product.schema.json
trieloff: Maybe we introduce an `xdm:condition` property, which holds multiple properties. Each property equals the XDM property name that is being compared. The value of the property would be (for now) `xdm:equals`.

Resulting in something like this:

```json
"xdm:condition": {
  "xdm:country": {
    "xdm:equals": "DE"
  },
  "xdm:channel.@type": {
    "xdm:equals": "https://ns.adobe.com/xdm/channel-types/web"
  }
}
```

schemas/content/product-attribute.schema.json
mhaack: I reworked the entire attribute value to be either a `simple` string or `https://ns.adobe.com/xdm/data/measure` be be more standardized and relay on things which are already there.

schemas/content/product-attribute.schema.json
mhaack: Changed to 'oneOf`.

schemas/content/product.schema.json
mhaack: As discussed with @trieloff changed back to `@id` & `uri`, same for category identifier.

schemas/content/product.schema.json
mhaack: Done.

schemas/content/product.schema.json
mhaack: I think we can go with the simple approach for now, without conditions. But I like the conditional properties, they should be standardized. I created #362 for that.

schemas/content/product-price.schema.json
```diff
@@ -0,0 +1,50 @@
...
+  "meta:extensible": true,
+  "meta:abstract": true,
+  "description": "Price model of a product in the product catalog.",
+  "definitions": {
+    "price": {
+      "properties": {
+        "xdm:amount": {
+          "title": "Amount",
+          "type": "number",
+          "minimum": 0,
```
trieloff: Are you sure about this? I'd assume you will find real data in PIMs that has negative amounts.

schemas/content/product-price.schema.json
```diff
@@ -0,0 +1,50 @@
...
+          "type": "string",
+          "examples": ["USD", "EUR"],
+          "pattern": "^[A-Z]{3}$",
+          "description":
+            "The ISO 4217 alphabetic currency code used for cost and pricing."
+        },
+        "xdm:type": {
+          "title": "Product Price Type",
+          "type": "string",
+          "examples": ["Sales Price", "MSRP", "Net Price"],
```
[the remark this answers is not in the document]
mhaack: I agree, will change it `meta:enum`

schemas/content/product-attributes.schema.json
```diff
@@ -0,0 +1,101 @@
...
+                {
+                  "type": "string"
+                },
+                {
+                  "$ref": "https://ns.adobe.com/xdm/data/measure"
+                }
+              ],
+              "description": "The value of the product attribute."
+            },
+            "xdm:variantAttribute": {
```
mhaack: 'xdm:variantAttribute' is not defined for attributes of type "type": "object" only, right?

trieloff: If the attribute is a primitive (string, number, boolean), then `xdm:variantAttribute` should be interpreted as having the default value, i.e. `false`.

schemas/content/product-attributes.schema.json
```diff
@@ -0,0 +1,101 @@
...
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/content/product-attributes",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Product Attributes",
+  "type": "object",
+  "meta:extensible": false,
+  "meta:abstract": false,
```
cdegroot-adobe: What does this mean, I don't think I have seen it used before in the XDM repo

schemas/content/product-attributes.schema.json
```diff
@@ -0,0 +1,101 @@
...
+        }
+      },
+      "patternProperties": {
+        ".+://.+": {
+          "$ref": "#/definitions/attribute"
+        }
+      }
+    },
+    "attribute": {
+      "oneOf": [
```
cdegroot-adobe: @kstreeter - Please look at this.
While I know JSON-Schema allows this, I don't think it is easy for us to represent this form of dynamic data in our applications.  
@mhaack see this: https://github.com/adobe/xdm/blob/master/CONTRIBUTING.md
`avoid unnecessary polymorphism that is hard to consume, e.g. offering both singular and array notations, or introducing type variants that don't provide a common discriminator property.`

schemas/content/product-price.schema.json
cdegroot-adobe: I have seen what Lars allude to here in the real world. For example a discount as a product.

mhaack: I removed the minimum constraint, even I don't agree on negative prices. Using it for example to model discounts smells wrong to me, there are other/better ways to do that.

schemas/content/product-attributes.schema.json
trieloff: https://github.com/adobe/xdm/blob/master/docs/abstract.md

schemas/content/product-attributes.schema.json
trieloff: I guess we can take away the simple option, which will lead to more verbose notation on the wire, but make it a bit easier to consume.

trieloff:
I second @cdegroot-adobe's comment that we can benefit from additional structure and known properties in `product-attributes`. Via Slack, @mhaack has expressed the concern that many of these attributes are vertical-specific and therefore not easily generalizable.

My suggestion:
* start with http://schema.org/Product – which is widely accepted and generic
* then look at common marketplace APIs like Amazon, eBay, Alibaba, Google Shopping
* as a final step, we should invite customers from verticals to help with extensions of `product-attributes` for common verticals like fashion, electronics, etc.

schemas/content/product-attributes.schema.json
mhaack: I simplified it now by removing the simple attributes as suggested by @trieloff

mhaack:
I added a bunch of generic product attributes from various "references" (Ebay, Shopify, Amazon, Aliexpress & Google).

schemas/content/product-attributes.schema.json
cdegroot-adobe: @trieloff is `"meta:abstract"` false by default if omitted?

schemas/content/product-attributes.schema.json
```diff
@@ -0,0 +1,193 @@
...
+            "$ref": "#/definitions/attribute"
+          },
+          "description": "Descriptive feature list of the product or varinat."
+        },
+        "xdm:productLine": {
+          "title": "Product Line",
+          "$ref": "#/definitions/attribute",
+          "description": "The product line of the product or varinat."
+        },
+        "schema:weight": {
```
cdegroot-adobe: I see no reason to use third party schemas for these simple definitions, It may even be wrong unless the `#/definitions/attribute` is from schema.org too. And considering it includes `xdm:` properties I think it does not.
@trieloff What do you think, I recommend converting all these `schema:*` attributes to first class `xdm:*` attributes.

schemas/content/product-attributes.schema.json
```diff
@@ -0,0 +1,193 @@
...
+          "description": "A pointer to another, functionally similar product (or multiple products)."
+        }
+      },
+      "patternProperties": {
+        ".+://.+": {
+          "$ref": "#/definitions/attribute"
+        }
+      }
+    },
+    "attribute": {
```
cdegroot-adobe: Can we move this to an external file so it is easy to incorporate in other schemas please

schemas/content/product-attributes.schema.json
```diff
@@ -0,0 +1,193 @@
...
+      "title": "Structured Product Attribute",
+      "properties": {
+        "schema:name": {
+          "title": "Name",
+          "type": "string",
+          "description": "The human-readable name of the attribute."
+        },
+        "xdm:value": {
+          "title": "Value",
+          "oneOf": [
```
cdegroot-adobe: This cannot work. We need concrete types.

schemas/content/product.schema.json
```diff
@@ -60,111 +59,50 @@
...
-          "description":
-            "The date when this product variant was created. The time using RFC3339 with a stated timezone offset such as \"2001-07-04T12:08:56-07:00\". An example formatting pattern is \"yyyy-MM-dd'T'HH:mm:ssXXX\"."
-        },
-        "xdm:productLastModified": {
-          "title": "Product Last Modified",
-          "format": "date",
-          "type": "string",
+            "Descriptive attributes associated with this product. "
+        },
+        "xdm:variants": {
```
cdegroot-adobe: This creates a highly nested structure, which is not what I think we want to do. What we have here now is a single product master with multiple variants. This will be expensive to query as it is a table inside a table where almost all joins will be to the inner table. This is a very big issue.

schemas/content/product.schema.json
```diff
@@ -60,111 +59,50 @@
           "description":
             "The internal unique ID of the product in the commerce backend system."
         },
-        "xdm:masterProductSKU": {
-          "title": "Master Product SKU",
-          "type": "string",
+        "xdm:prices": {
```
cdegroot-adobe: How do we handle each variant having a different price?

mhaack: Since variants are product objects and use the same schema the can also have prices as well. If a variant has a same price like product but with different value than it can be set on variant level with same price type. 

For example a shirt with 5 different sizes defines the sales price (=type) on product level to USD 29 and variant with size S is USB 19. For this variant you define the sales price. All other variants inherit the sales price from the product.

schemas/content/product-attributes.schema.json
trieloff: It is valid JSON Schema.

schemas/content/product-attributes.schema.json
trieloff: We should use `schema:*` attributes whenever we can, as they make the meaning unambiguous. XDM is derived from Schema.org, so there is no need to re-define these properties.

schemas/content/product.schema.json
trieloff: We will have more than one implementation of XDM, so I think there are bigger concerns than the query efficiency of one single implementation. In addition, almost any PIM implements some relationship of product masters to variants, so this is a join that is quite common and not enough rationale for me to reject the PR.

cdegroot-adobe: @trieloff - I disagree. The current design is not designed to support the primary use case we see for Products which is that of an experience in the context of a product. Int his case the product being experienced is almost always a product variant. So it does not makes sense to me for this to be embedded in the master, which is more of a seldom used formality that is used for PIM data designs. 

To make an analogy, this is like building an CDN, but the only records it has is of templates with an inner list of the actual media that is being requested.

The previous design had variants as first class records and still maintained the concept of master records. While there are many improvements in this PR, this is not one and needs to be revisited.

mhaack: @cdegroot-adobe can you elaborate on
> The current design is not designed to support the primary use case we see for Products which is that of an experience in the context of a product.

I think there are different "we" views. We (the AEM commerce team) look into broader commerce, not only the client experience. For example I can think of importing PIM catalog data into ACP and running ML task on that data.

I think with the current proposal you can realize all kinds of product views. From experience context (the product schema has one needs, there is no need to use the nested variant structure) up to complex PIM data structures.

prabhum2:
@mhaack -  Can we close this PR

Issue #309, opened 2018-05-07
Extend Analytics to include operating system ID and browser ID


stormybriggs1:
## What are the schemas that are affected by the issue
extensions/adobe/experience/analytics/experienceevent.schema.json

## What are examples of products that are impacted by the issue
Adobe Analytics ingestion and all down stream consumers.

lrosenthol:
This information is included in browserDetails referenced from https://github.com/adobe/xdm/blob/9783604737f91692a9529d9e3a0b9d9bb6411b9d/docs/reference/context/environment.schema.json (among other places)

cdegroot-adobe:
@lrosenthol The IDs by which different systems know that browser or OS by is not part of the current schemas. Different systems have different IDs. Device Atlas etc. In this case the issue is outlining something in the scope of Analytics as it is in the Analytics extension.

Make sense?

PR #311, opened 2018-05-08
Extend Analytics to include operating system ID and browser ID

stormybriggs1:
Issue #309 Extend Analytics to include operating system ID and browser ID.

This is an analytics specific concern that allows us to retain legacy browser and operating system references as we move to XDM.


stormybriggs1:
I would like to get this merged in for the May 18th milestone. Thanks!

stormybriggs1:
@kstreeter Would you mind looking at this short PR?

kstreeter:
@stormybriggs1, what scheme is being used for these identifiers?  If they are something standard or commonly used I would suggest we add to the browserdetails or environment so they can be used elsewhere.  Are these from something like (for example) DeviceAtlas?

stormybriggs1:
@cluby-adobe What are your thoughts on kstreeters comments?

cluby-adobe:
@kstreeter, these are Analytics internal lookups.  They are not from a sharable standard.

extensions/adobe/experience/analytics/experienceevent.schema.json
```diff
@@ -7212,6 +7212,18 @@
               "description": "Custom event 1000."
             }
           }
+        },
+        "https://ns.adobe.com/experience/analytics/environment/browserID": {
```
trieloff: Should be just `https://ns.adobe.com/experience/analytics/browserID`

extensions/adobe/experience/analytics/experienceevent.schema.json
```diff
@@ -7212,6 +7212,18 @@
...
+          "title": "Browser ID",
+          "type": "integer",
+          "description":
+            "The Adobe Analytics Identifier for the browser used."
+        },
+        "https://ns.adobe.com/experience/analytics/environment/operatingSystemID": {
```
trieloff: Should be just `Should be just `https://ns.adobe.com/experience/analytics/operatingSystemID `

lrosenthol:
If these values represent an *internal* concept, then they need to be in an extension schema and not a public one

stormybriggs1:
@kstreeter @trieloff Please take another look. I have updated this PR as suggested by @cluby-adobe. Thanks.

Issue #265, opened 2018-04-14
Move experienceevents metrics to each of the different focus fields


cmathis:
There was an idea talked about a while ago to remove that the Analytics group was expecting where we remove the experienceevent.metrics field (which today references metrics.schema.json) and instead move each of the metric fields under the the different experienceevent channel fields that relate to those metrics (web, marketing, commerce). 

For examples, instead of having: 
experienceevent.**xmd:metrics**.http://ns.adobe.com/xdm/data/xdm/metrics/commerce/purchase
We would instead have:
experienceevent.**xdm:commerce**.xdm:purchase (no need for this to be a URI) 

Commerce metrics become members of the https://ns.adobe.com/xdm/context/commerce
Direct-Marketing metrics become members of https://ns.adobe.com/xdm/context/marketing
Web metrics become members of https://ns.adobe.com/xdm/context/webinfo 

This keeps all the data associated with each of these different channels together and aligns better with cases where a solution/partner may create a different channel specific field (experience:campaign extension, experience:analytics extension, etc.) with its own set of metrics. 

cc @cdegroot-adobe , @kstreeter

cluby-adobe:
@cdegroot-adobe, There are a couple problems introduced by having them separate from the verticals with the related dimensions:

* We don't have a consistent naming convention for the paths between the metrics and the rest of the dimensions.  The customer will have to deal with 2 independent concepts based on what type of field they are dealing with.  This will be a poor customer experience.
* With the solution and customer extension they won't have the pattern of a separate, combined metrics structure.  They will mix the metrics in with the dimensions.  So on top of dealing with inconsistent paths just in the core between metrics and dimensions, they will have to mentally switch to a different pattern based on using the core vs solution and customer extensions.

Given that customers don't read documentation to explain the inconsistency.  The inconsistency will negate any benefit of the consolidated metrics structure.  We need to make the structure consistent, obvious, and self documenting.

cdegroot-adobe:
@cluby-adobe thanks for the comments, much appreciated.

If we have to convert IRIs to XED formats for persistance then I think there is some consideration for doing as you say. But generally I'm not seeing it.

[...]

PR #316, opened 2018-05-09
Issue #265 - Part1 - Moved web metrics to the web contexts

cdegroot-adobe:
Shortened the URI field name for the web events to
https://ns.adobe.com/xdm/metrics/web/* by dropping the 'data'.
Updates all examples.
Updarted the change log

Please link to the issue #265

### DIFF
--- a/schemas/context/webinteraction.example.1.json
+++ b/schemas/context/webinteraction.example.1.json
@@ -1,5 +1,8 @@
 {
   "xdm:type": "other",
   "xdm:URL": "https://www.example.com/products/store/?view=1",
-  "xdm:name": "product store"
+  "xdm:name": "product store",
+  "https://ns.adobe.com/xdm/metrics/web/link-clicks": {
+    "xdm:value": 1.0
+  }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,3 +11,4 @@
 [Unreleased](https://github.com/adobe/xdm/releases/tag/v0.9.1)
 
 * Renames all `Metric` schemas to `Metric Definition` #254
+* Moved web metrics used in ExperienceEvent \\metrics to the webinteraction and webpageview contexts #316
--- a/schemas/context/experienceevent.example.1.json
+++ b/schemas/context/experienceevent.example.1.json
@@ -51,11 +51,6 @@
     "https://ns.adobe.com/xdm/data/metrics/commerce/purchases": {
       "xdm:value": 1.0,
       "xdm:unit": null
-    },
-    "https://ns.adobe.com/xdm/data/metrics/web/page-views": {
-      "@type": "https://ns.adobe.com/xdm/data/metrics/web/page-views",
-      "xdm:value": 1.0,
-      "xdm:unit": null
     }
   },
   "xdm:productListItems": [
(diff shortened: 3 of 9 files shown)


cdegroot-adobe:
@kstreeter this is the first break out part of the Metrics re-org. Can you please review asap. I am now working on the next ones Analytics requires.

schemas/context/webinteraction.example.1.json
```diff
@@ -1,5 +1,8 @@
 {
   "xdm:type": "other",
   "xdm:URL": "https://www.example.com/products/store/?view=1",
-  "xdm:name": "product store"
+  "xdm:name": "product store",
+  "https://ns.adobe.com/xdm/metrics/web/link-clicks": {
```
kstreeter: @cdegroot-adobe, we'll need to decide what we want the namespace for these to be.  I'm not sure having a metrics namespace is even necessary: would we want metric and non-metric fields that have the same name in the same object?  If we choose to have a 'web' namespace, it will be somewhat redundant to the containing object.  So I actually feel like we should just put this directly into the XDM namespace.  @trieloff, what are your thoughts on that?

cdegroot-adobe: @kstreeter are you saying 
We change:
`https://ns.adobe.com/xdm/metrics/web/link-clicks`
to be:
`https://ns.adobe.com/xdm/link-clicks`
Which then is the same as
`xdm:link-clicks`

Then we need to refactor the metric definitions.

trieloff: I can live with the shorter namespace URI, but it would be a breaking change. So far, this change looks non-breaking.

schemas/data/page-views.schema.json
```diff
@@ -5,7 +5,7 @@
     "you may not use this file except in compliance with the License. You may obtain a copy",
     "of the License at https://creativecommons.org/licenses/by/4.0/"
   ],
-  "$id": "https://ns.adobe.com/xdm/data/metrics/web/page-views",
+  "$id": "https://ns.adobe.com/xdm/metrics/web/page-views",
```
trieloff: Now it's a breaking change. I suggest to leave it in `data/metrics`.

cdegroot-adobe: Because this is moving in the hierarchy it is already a breaking change. I strongly recommend we move tot he URIs we want at this point. The teams are already absorbing a number of breaking changes so this can be absorbed at this point.

kstreeter: +1 on this, moving the fields to a different parent is already breaking, better we do it now versus having another breaking change in the same area later.  Can we agree on putting everything in "https://ns.adobe.com/xdm/metrics"?

trieloff: If we move it to `xdm/metrics`, we will have to move all files from the `schemas/data` folder to the `schema/metrics` folder and introduce this new folder. I just don't think that's worth it.

cdegroot-adobe:
@kstreeter can you look at this update please and let me know if it is what you expected with the URI changes. Thanks C.

kstreeter:
This is exactly what I was thinking @cdegroot-adobe.  We've now got good consistency between dimensions and metrics in each area, and are still able to leverage the `MetricDefinition` pattern to provide additional detail about how each measure relates to the metric that defines it.

cdegroot-adobe:
@kstreeter You stated:
"As we go through this, I really feel like we should get rid of the 'metrics' namespace entirely. I think we should just put all the metrics that are part of the core schemas in the 'xdm' space. Others have stated this, but it just seems unnatural to have dimensions and metrics in different namespaces. In general, we would not want to have dimensions and metrics in a single object that differ only by namespace. The use of the MetricDefinition and Measure schemas should be enough to organize the metrics into a consistent framework.
If we do this, we don't need to re-organize the folders. I agree with @trieloff that just removing one portion of the namespaces and creating conflicts with the folder structure is probably not worth it."

Can you give us simpler folk an example please?
Do you mean the following? `https://ns.adobe.com/xdm/data/page-views` as an example?

cdegroot-adobe:
@trieloff - please review as soon as you can. once one of these metric changes is agreed the rest will be able to follow the same pattern.

PR #317, opened 2018-05-09
Issue #265 - Part2 - Moved the commerce metrics from ExperienceEvent \metrics to the \commerce context

cdegroot-adobe:
Moved the commerce metrics from ExperienceEvent \metrics to the \commerce context
Shortened the URI field name for the web events to
https://ns.adobe.com/xdm/metrics/web/* by dropping the 'data'.
Updates all examples.
Updarted the change log

Please link to the issue #265

Please link to the issue #…


cmathis:
FYI - In it's compatibility form the path from the root of experienceevent will be commerce.mertrics.commerce.checkout or commerce.commerce.checkouts, etc. Either way you are repeating the "commerce" string. 

This is because you are putting a URI field name with the string "commerce" under another field already named "commerce. 

Yes, we can add special handling for these metric specific URI's, but I would ideally like to avoid doing anything special to that path as you must have a reason for naming the field that way.

cdegroot-adobe:
@cmathis - I was just reviewing the URIs with Kevin this morning and I updated them in the last couple commits just after you commented. The URI is now `https://ns.adobe.com/xdm/metrics/{propertyName}` for all the metrics going forward. 

Yes for the XED conversion you will need a special rule to strip out `metrics`

cdegroot-adobe:
@kstreeter  @trieloff  please review.

CHANGELOG.md
```diff
@@ -11,4 +11,7 @@
 [Unreleased](https://github.com/adobe/xdm/releases/tag/v0.9.1)
 
 * Renames all `Metric` schemas to `Metric Definition` #254
+* Moved `Commerce` metrics used in ExperienceEvent `metrics` to the `commerce` context #317
+* Shortened the URI field name for the web events to `https://ns.adobe.com/xdm/metrics/*` by dropping the 'data'. #317
```
trieloff: This change is incomplete (you also need to create a new directory, explain it in all documents that mention the repo structure, adjust all URLs), breaking (lots of stuff) and overall not worth it.

schemas/data/abandons.schema.json
```diff
@@ -5,7 +5,7 @@
     "you may not use this file except in compliance with the License. You may obtain a copy",
     "of the License at https://creativecommons.org/licenses/by/4.0/"
   ],
-  "$id": "https://ns.adobe.com/xdm/data/metrics/commerce/abandons",
+  "$id": "https://ns.adobe.com/xdm/metrics/cartAbandons",
```
trieloff: The file name is conflicting with the metric name and the metric name is conflicting with the established patterns. My suggestion:

1. rename the `$id` to `https://ns.adobe.com/xdm/data/metrics/commerce/cart-abandons`
2. rename the file to to `cart-abandons.schema.json`

PR #326, opened 2018-05-15
Fix #325 Add mutations to campaign experienceEvent.

### LINKED ISSUE #325: Introduce Mutations concept to Campaign's experienceEvent
experienceEvent
Campaign

In the current scenario, expeirenceEvents are said to be immutable. However, there may be cases where a transition needs to be taken care of.
For example, in Campaign, an Email gets Sent(generates an eE:metric:sends=1), then gets bounced(generated another eE:metric:bounces=1) and then retried to get finally sent again(generates an eE:metric:sends=1).
Question: How many Emails were sent?
There are 2 types of downstream applications to answer these:
1/ Which handle each event as immutable and report the number as 2. The email was tried twice.
2/. Treat these as transitions when the same email delivery status has mutated across states. 

To facilitate the functioning of downstream applications like 2/ , we need to pass some information of the previous event that was sent for this. This can be handled by sending a clipped version of the previous eE in a special field `mutation` which lists the old values of just those fields that were changed in this new event.

This requirement is currently for Campaign only. However, we may argue if this could be a core concept.

Discussions in comments at an internal wiki page : display/~hsawhney/XDM+1.0+review+for+the+Campaign+events 

Tag: @cmathis

saurabhere:
Fix #325 

To Review: @cmathis @cdegroot-adobe @trieloff


kstreeter:
Hi @saurabhere, can you explain this use case in a little more detail?  I believe that generally the model for ExperienceEvent is that it captures state at the time of the current event, and that a processor understands how that state changes over time by looking across all of the events in a session or time window.

saurabhere:
There is some description in #325 .

Campaign would initially be sending 2 types of experienceevents:
1. Delivery Events
2. Tracking Events

Delivery Events describe the state of a message. It may transition from error->sent->bounce->sent... So, there needs to be a track of the final state.
Tracking events are rather simple and immutable. When someone opens the message or interacts with it, a tracking event is sent out. So mutations don't apply there.

The use-case is that one of the consumers of this XDM would be campaign's in-house analytics engine which is designed in such a way that each time an event proceeding an event happens,, it expects a negative count of the previous metric. So, when asked the overall aggregation of the metric, it returns a correct value. A more detailed description can be found at the internal wiki page titled: "Realtime Reporting Design" in Adobe Campaign space.
The idea here would be to send the details of the previous event in every event so that a negative event can be generated by the reporting engine.

While your point is valid in the context of sessions, there is no such concept here. A message is described by a messageId and a bounce event may come even after weeks of sending it. Hence, It becomes hard to query the engine(on a dimension with a very high cardinality) every time you ingest one record. 

This idea can be extended to other services as well, but since it is mission critical for campaign's existing functionalities, i have introduced it for Campaign's extension at this time.

cmathis:
@saurabhere  - Does this mean an ExperienceEvent coming from Campaign may contain a nested ExperienceEvent, which contains another nested ExperienceEvent, and so on until you read the original ExperienceEvent? If each of those are fully populated the payload of that message is going to be huge and will likely fill up the max 1MB message size on Pipeline.

saurabhere:
@cmathis It would be like the example shown.
Only the fields that changed between this and the previous event will be sent. Just one level nesting.

cdegroot-adobe:
@saurabhere - Can you explain this a bit more. How for example will the event not be counted twice by systems that do not understand the Campaign Extension?

saurabhere:
@cdegroot-adobe This doesn't cause double counting because in case of a mutation, we are still sending only one experienceEvent. Just that, it will have an additional information about the previous event. 
It would help downstream applications to be able to delete/negate the previous events by constructing a fresh row of data based on the current fields and the previous fields.
So for example, as shown in https://github.com/saurabhere/xdm/blob/ea6ba6dc39f2011f1f8a1b42d8c64b4521f26607/extensions/adobe/experience/campaign/experienceevent.example.sms-bounce.json
It shows a bounce event for SMS. It implies that a "Sends" event was previously sent. 
A downstream application can now choose to generate a dummy row:
```javascript
{
    "xdm:timestamp": "2017-09-26T14:52:25+00:00",
    "xdm:metrics": {
      "https://ns.adobe.com/xdm/data/metrics/direct-marketing/sends": {
        "xdm:value": -1
      },
    "xdm:reason": "undefined",
    "xdm:reasonMessage": "* Sent.",
    "xdm:quarantine": false,
...
}
```
(Note the negative xdm:value)

to negate the previous `sends` event. They will ingest the current `bounce` event normally by excluding the `xdm:mutation` object.
A more detailed description can be found at the "OLTP data to OLAP data" section of the internal wiki page titled: "Realtime Reporting Design" in Adobe Campaign space.

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -257,6 +257,11 @@
                 "Business qualifier that identifies the event sent by the data source. It informs on the business reason for sending the event. It is unique per organization. This is used by Campaign orchestration to identify the event without inspecting its payload to determine which action should be triggered when the event is received. The value of this field is a contract between Campaign orchestration and the data source."
             }
           }
+        },
+        "xdm:mutation": {
+          "title": "Mutated fields from the previous experienceEvent.",
+          "$ref": "https://ns.adobe.com/experience/campaign/experienceevent",
+          "description": "Information of the fields that have changed since the previous experienceevent to keep track of the state changes across events. For e.g. if when an email gets sent, an eE with metric `sends` will be sent. This may get bounced and an eE with metric `bounces` will be sent then. It may be important for a downstream application to know that the previous eE with metric `sends` is now to be invalidated. This field will contain the values from the previous eE that have a different value in the current eE."
```
trieloff: @saurabhere please change "eE" to "Experience Event"

saurabhere: damn! I was hoping it would stick.

rennieja:
@saurabhere over what period can a delivery event be mutated?  can you provide an example how to determine if a delivery record indicates a "successfully received by the recipient" event or "not successfully received by the recipient" event?  is their a limit to retries?  in the scenarios described it sounds like there can be multiple sent and multiple bounce events.  what causes "xdm:quarantine" to move from false to true?  is it based on # of bounces, or on a specified smtp message.  in ACC, there is a undetermined state, meaning the smtp message is not recognized or has to be evaluated; not classified as hard or soft bounce.  how do we recognize those?  which attributes from experience events from campaign indicate a hard vs soft bounce.  (terminal vs temporary failure).  appreciate additional context.

saurabhere:
@rennieja These are great many questions. Can you connect with me on the organization's email? I think some detail you ask may require info on campaign internals.

rennieja:
@saurabhere will do.

Issue #328, opened 2018-05-15
New experienceevent schema which includes all solution extensions


jwen-adobe:
## What are the schemas that are affected by the issue
New schema: experienceevent under extensions/adobe/experience folder

## What are examples of products that are impacted by the issue
Platform XDM and tooling

PR #329, opened 2018-05-15
Added new experienceevent schema which includes all solution extensions

jwen-adobe:
This PR links to the issue #328 

This PR create a uber experienceevent schema which contains core/standard fields as well as the solution extension experienceevent fields.


cmathis:
Looks great Joe. 

Editors, let's not merge this in yett. Having discussions with the PM group that may affect some of this.

cmathis:
We can merge this in. This is used to create the equivalent protobuf and POJO library classes that will be used by Adobe solutions.

PR #273, opened 2018-04-23
[WIP]Issue 265 move metrics under experienceevent contexts

cdegroot-adobe:
Completed via other PRs that broke down the changes. 

UPDATE: This is being broken up into smaller more manageable chunks.

PR to update ExperienceEvent so that metrics are under the contexts that they relate to.

Relates to issue #265

PR #331, opened 2018-05-16
Issue #265 part4 moved direct marketing metrics to its own contexts

cdegroot-adobe:
This is a break out of PR #273 related to materializing the direct-marketing name pace as a context in ExperienceEvent


cdegroot-adobe:
@kstreeter  - would like to get in with other metric reorgs.

schemas/context/direct-marketing.schema.json
```diff
@@ -0,0 +1,63 @@
...
+          "description": "The direct marketing message (email, SMS, Push Notifications etc) has been opened/swyped/touched by the recipient."
+        },
+        "xdm:unsubscriptions": {
+          "$ref": "https://ns.adobe.com/xdm/data/measure",
+          "description": "The recipient of the direct marketing message (email, newsletters, etc) has opted out from receiving them."
+        },
+        "xdm:notSent": {
+          "$ref": "https://ns.adobe.com/xdm/data/measure",
+          "description": "Asynchronous message (email, SMS, MMS etc) was not dispatched due to cancellation or expiration of the marketing activity."
+        }
```
saurabhere: Should `xdm:messageClicks` also be added to this list?

schemas/context/experienceevent.schema.json
```diff
@@ -88,6 +88,12 @@
           "description":
             "The information related to web page and link of the ExperienceEvent."
         },
+        "xdm:directMarketing": {
+          "title": "Direct Marketing",
+          "$ref": "https://ns.adobe.com/xdm/context/direct-marketing",
+          "description":
+            "The events and properties related to direct/outbound marketing such as email, direct mail, texts and in-app notifications."
+        },
```
saurabhere: Can you also change the examples at https://github.com/adobe/xdm/tree/master/extensions/adobe/experience/campaign with this change?

schemas/data/messageClicks.schema.json
```diff
@@ -0,0 +1,47 @@
+{
+  "meta:license": [
+    "Copyright 2018 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/data/metrics/direct-marketing/message-clicks",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Message Clicks",
```
saurabhere: I am not sure if we really need a third metric after `web/linkclicks` and `advertising/clicks` already. While it may make sense for each paradigm to have their own metric for their click activities, I guess all the three are essentially represent weblink clicks.

kstreeter: I could see this either way.  Agreed they are all just clicks on a link.  But the context of each is very different, and therefore the meaning or intent of the click is very different.  On the other hand, we capture all that context, so a processor could always sort it out later.  @cluby-adobe can you weigh in on this?

saurabhere:
@chrisdegroot I think the idea was to either keep a single click metric OR have context specific metrics.
If we remove `direct-marketing/message-clicks`, either of the `advertising/clicks` and `web/link-clicks` should be removed too. At this point, while advertising and web contexts have their own metrics, others(direct-marketing) need to pick between the two.

schemas/context/experienceevent.schema.json
cdegroot-adobe: done

cdegroot-adobe:
@saurabhere - that is not my understanding. Previously you made a comment that a mail click is the same as a web/link click, and therefore direct-marketing should really use that. This was a really great insight. Is this no longer your opinion.
@kstreeter - Do not merge this until we clarify the go forward strategy here please.

saurabhere:
But isn't `advertising/clicks` also same as `web/link-click`? If yes, then let's merge them. If not, let's define which would be used when and based on that direct-marketing can use one of them. 
Although if metrics are defined to be context specific, then a dedicated `message-clicks` indeed makes sense.

cdegroot-adobe:
@kstreeter - this is ready to merge now.

PR #333, opened 2018-05-16
EndUserID naming made more consistent.

cdegroot-adobe:
- "analytics" -> "aaid"
- "campaignid" -> "acid"

### DIFF
--- a/schemas/context/enduserids.example.1.json
+++ b/schemas/context/enduserids.example.1.json
@@ -5,7 +5,7 @@
       "xdm:code": "ECID"
     }
   },
-  "https://ns.adobe.com/experience/analytics": {
+  "https://ns.adobe.com/experience/aaid": {
     "@id": "https://data.adobe.io/entities/identity/2394509340-30453470347",
     "xdm:namespace": {
       "xdm:code": "AVID"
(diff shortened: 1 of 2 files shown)


cdegroot-adobe:
@kstreeter  - Required for Analytics Beta.

lrosenthol:
is there a reason we shortened these to terms that most users won’t know?

cdegroot-adobe:
@lrosenthol we are choosing the shortened names because in most cases they are the standard ones customers use. Customers that have this data will be using them with these shortened names most commonly. The Campaign Id is really a new concept and Analytics Id has is also a new term that is not in common use yet. mcid and tntid are the standard. So we are moving to be consistent.

cdegroot-adobe:
@saurabhere can you please review this change for campaign. Thanks C.

kstreeter:
confirmed this is how the teams would like to go with the naming

saurabhere:
@kstreeter @cdegroot-adobe Althought for `acId` the `@id` attribute will be an integer and not uri based. How will `aaid` and `mcid` will be sent though? I understand mcId is of the form "1234567890ABCDEF@AdobeOrg'. No?

PR #336, opened 2018-05-16
Fixed blocking issue with name of webPageView.

cdegroot-adobe:
Updated:

- Renamed to WebPageDetails

One blocking issue raised during the 0.9 reviews was that the experimental schema and referenced field for `context\webpageview` was named like a metric and that was very problematic. 

The consuming teams agree that `webPageDetails` is a better term. Their preferred term was `webPage` but that is problematic because it really should refer to the contents of an asset.


kstreeter:
@cdegroot-adobe, the rationale for changing this to "webPageView" was that the event being tracked was a user viewing a page, as distinct from a user clicking or otherwise interacting on that page.  The new naming seems to have lost that distinction.  Can we find a name that makes this clear, without confusing it with the metric of the same?  Or perhaps change the name of the metric?

cluby-adobe:
@kstreeter, the confusion is that that struct inside ExperienceEvent was never meant to just be about a page-view.  It's just where you put the details about a page regardless of the reason you have that data.  It's just grouping fields like URL, name, etc that are related to a web-page.  If it's a page-view then the page-views measure field in that structure would be populated.  If it's a link-click and you are representing the containing page then you would not populate page-views and instead populate link-clicks under the webInteraction structure.

kstreeter:
Thanks @cluby-adobe, that makes sense.

kstreeter:
@trieloff, @lrosenthol can you comment on this?  I know you had input on the original change.

trieloff:
Looks good.

Issue #337, opened 2018-05-17
Analytics specific state and zip code


stormybriggs1:
## What are the schemas that are affected by the issue
extensions/adobe/experience/analytics/experienceevent.schema.json

## What are examples of products that are impacted by the issue
Analytics ingestion and downstream products.

## Description
In addition to the geo information that we lookup, the client provides location information such as state and zip code that we would like to retain in XDM.

PR #338, opened 2018-05-17
Add analytics specific state and zip

stormybriggs1:
Please link to the issue #337.

In addition to the geo information that we lookup, the client provides location information such as state and zip code that we would like to retain in XDM.


cdegroot-adobe:
Yes, this is good with me. It is provided by the client side and it makes sense to keep it in the Analytics extension if they need to continue using it.

kstreeter:
@stormybriggs1, how are these different than the same fields available in common/address?

stormybriggs1:
@kstreeter Within Analytics we have multiple values for state and zip. We have those for which use a service to derive geo information and we have client supplied values. These fields are to retain the client supplied values. The difference is the source.

PR #339, opened 2018-05-18
Add fields for custom dimensions in Campaign.

saurabhere:
This adds 20 new cusField fields to Campaign's ExperienceEvent.
This is very similar to the analytics extension of experienceEvents adding eVars to the schema.
The use-case is to allow customers to send upto 20 variables in experienceevents and configure them. This is handled by a multitenant OLAP cube which provides Reporting and Analysis to the customers.
The long term plan is to be able to use the aliasing feature in catalog where one could create aliases to these cusFields so that they can be addressed using friendly names as well.

To Review: @cdegroot-adobe @kstreeter .

### DIFF
--- a/extensions/adobe/experience/campaign/experienceevent.example.email-send.json
+++ b/extensions/adobe/experience/campaign/experienceevent.example.email-send.json
@@ -66,5 +66,6 @@
   },
   "https://ns.adobe.com/experience/campaign/orchestration": {
     "xdm:businessReason": "onJourneyEnter"
-  }
+  },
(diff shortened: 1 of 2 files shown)


cdegroot-adobe:
@saurabhere - Can you explain how these are used in todays Campaign schema. This is a very new request to me as we have been designing the schemas for Campaign for well over a year and I have not seen a request for these before. It goes against the design of XDM to do this. The only reason we allow this extension concept for Analytics is because it is central to how that solution works. By default the answer is "no". Your team will need to make a very strong case for why these need to exist.

PR #320, opened 2018-05-11
[DO NOT MERGE] Issue #265 part3 Added the mobile/application metrics to the a new application context #317

cdegroot-adobe:
Break down of #266

PR #340, opened 2018-05-20
Issue #265 - part 3 (try 2) adding mobile/application context and metrics

kstreeter:
This PR is intended to replace #320, which has somehow gotten completely out of sync with master.  I wasn't able to fix it w/o hand reverting dozens of commits, so instead I just pulled all the changes from #320 and made a new branch.

Note this loses some of the discussion history, so please see that PR for historicals on this.  We had (to my understanding) addressed all comments before the branch got messed up.  Let's do a quick pass on the changes and get it in.


schemas/context/application.schema.json
```diff
@@ -0,0 +1,75 @@
...
+        "xdm:upgrades": {
+          "$ref": "https://ns.adobe.com/xdm/data/measure",
+          "description":
+            "Upgrade of an application that has previously been installed. Triggered on first launch after upgrade."
+        }
+      }
+    }
+  },
+  "allOf": [
+    {
```
cdegroot-adobe: Add the context to the all of

Issue #357, opened 2018-05-25
Commerce Schema is not extensible


trieloff:
See #317 the commerce schema declares `meta:extensible` but is not actually extensible.

cdegroot-adobe:
@kstreeter  lets pull this into the June 15th milestone. I need to review other schemas to ensure they are correct in this regard.

PR #368, opened 2018-05-30
Stricter Meta Schema, Fixed Issues Found

trieloff:
As discussed in #357, we have a couple of inconsistent schemas that declare one thing in `meta:*` and then disregard the implementation. This was partly caused by a bug in our tests that skipped the `meta.schema.json` validation, so that even invalid schemas would pass.

This PR:
- makes sure every schema is validated against `meta.schema.json`
- adds rules to `meta.schema.json`, in particular
  - when using `meta:extends`, make sure there is an `allOf` with at least two `$ref`s and each `$ref` points either to a URL or to something in `#definitions`
  - `meta:status` is present for every schema
- fixes all violates of the above rules (our metric definitions are a mess)
- fixes two remaining metrics that have not been converted into metric definitions


cdegroot-adobe:
This is great, thanks Lars for adding.

schemas/data/poi-entries.schema.json
```diff
@@ -42,6 +42,9 @@
   "allOf": [
     {
       "$ref": "#/definitions/metric"
```
cdegroot-adobe: This looks like an error too, it should be `"$ref": "#/definitions/metricdefinition"`

schemas/data/poi-exits.schema.json
```diff
@@ -41,6 +41,9 @@
   "allOf": [
     {
       "$ref": "#/definitions/metric"
```
cdegroot-adobe: This looks like an error too, it should be "$ref": "#/definitions/metricdefinition"

cmathis:
@cdegroot-adobe  - With this change all of those solution extensions would need to include in their allOf the definition of the base context/experienceevent.  In doing this they are no longer smaller schema fragments that include only the new fields, but instead represent a complete schema which includes the fields of the schema it is extending plus the new fields.  It's really a pre extended version of experienceevent. 

I'm wondering how this affects the workflow you've talked about where these were intended to be only small fragments that are mixed in later via and API call and creation of a customer extended schema. Do we need another way to define extension fragments that can be selected and mixed in later rather?

trieloff:
We already have a way of creating schema fragments: these are abstract schemas, i.e. `definitions`: yes, `$ref`: no.

cmathis:
@trieloff - From what I understand the UI is to eventually allow a user to select the base context/profile schema, and then a second drop down or selection shows the extensions that can be added. How will the UI/API know what schemas to show in that second selection? I think the PM group has been thinking the meta:extends contains the list it CAN extend.

trieloff:
@cmathis please open a separate issue for this, paging @kstreeter I'm not opposed to changing the semantics of `meta:extends`,  especially that right now it is redundant to `allOf/…/$ref`

Issue #371, opened 2018-05-31
Analytics custom ID


stormybriggs1:
## What are the schemas that are affected by the issue
schemas/context/enduserids.schema.json

## What are examples of products that are impacted by the issue
Analytics ingest and downstream products.

## Description
Within Analytics we have a field that we would like to preserve that is a custom identifier passed in by the client.

PR #372, opened 2018-05-31
Analytics custom ID

stormybriggs1:
Issue #371 . Add a custom identifier to preserve an Analytics specific identifier. This identifier is passed in by clients as an override.


lrosenthol:
+1 from me - looks fine.

Issue #373, opened 2018-05-31
Received Timestamp


stormybriggs1:
## What are the schemas that are affected by the issue
schemas/context/experienceevent.schema.json

## What are examples of products that are impacted by the issue
Analytics ingest and downstream products.

## Description
In addition to a client specified timestamp, Analytics also needs to specify the timestamp at which time the server received the experience event.

PR #374, opened 2018-05-31
Received timestamp

stormybriggs1:
Issue #373 . Analytics would like to preserve the timestamp of when a server received an experience event.


lrosenthol:
+1 this also seems fine to me

Issue #375, opened 2018-05-31
Analytics custom lists needs to support keys on each list value


stormybriggs1:
## What are the schemas that are affected by the issue
extensions/adobe/experience/analytics/experienceevent.schema.json

## What are examples of products that are impacted by the issue
Analytics ingest and downstream products.

## Description
Analytics custom lists sometimes have key values.

PR #376, opened 2018-05-31
Alter analytics custom lists to allow keys

stormybriggs1:
Issue #375 , alter the data type of custom lists in Analytics to support key-value pairs.


extensions/adobe/experience/analytics/keyedlist.schema.json
```diff
@@ -0,0 +1,45 @@
...
+  "description": "List of keyed values.",
+  "definitions": {
+    "keyedlist": {
+      "properties": {
+        "xdm:list": {
+          "title": "Keyed List",
+          "type": "array",
+          "items": {
+            "type": "object",
+	          "title": "Key-value pairs",
```
cluby-adobe: Inconsistent whitespace

stormybriggs1: Should be fixed now.

extensions/adobe/experience/analytics/keyedlist.schema.json
```diff
@@ -0,0 +1,45 @@
...
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/experience/analytics/keyedlist",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Keyed List",
+  "type": "object",
+  "description": "List of keyed values.",
+  "definitions": {
+    "keyedlist": {
```
cluby-adobe: For other reviewers the reason we are defining this type vs just using an object with additionalProperties (a map) is for our usage in Analytics list variables the key is optional and we are also keeping open the option of adding the original delimiters to this struct later like we have for listdetails - https://github.com/adobe/xdm/blob/master/extensions/adobe/experience/analytics/listdetails.schema.json#L24

lrosenthol:
I would like to see a more detailed justification for this before we move forward.

cluby-adobe:
@lrosenthol, We clarified more details on the data and usage for these 3 fields we didn't have clarity on before.  We need to preserve the keys from the original source of this data so `listdetails` for these 3 fields is the wrong data type.  We talked about adding an optional key to listdetails but that is polluting the intent of that data type and would cloud what's actually supported for the other fields that are correctly using listdetails.
This is a breaking change to this schema but it's one we have to make.

extensions/adobe/experience/analytics/keyedlist.schema.json
trieloff: I don't think it's the most elegant way to express something, but fine for an extension.

Issue #361, opened 2018-05-25
Circular References in XDM


cmathis:
https://github.com/adobe/xdm/blob/744e787bbd8cafcc404c979556e49ba86fa94c4a/extensions/adobe/experience/campaign/experienceevent.schema.json#L263

We have an issue here where ExperienceEvent is extended with a Campaign field which in turn defines a field that references ExperienceEvent. This causes all kinds of issues for the UI trying to dynamically show the nested fields, and has also caused a problem with the dynamic code generation based on the schemas. 

If this type of circular reference is needed in the XDM schema, can we come up with a meta tag or hint that the UI and code generators can use to know it should not follow that reference?

kstreeter:
@cmathis, I think we need to consider how to handle this in a general way (perhaps even disallow cycles).  In the short term I recommend that the tooling/UI simply not follow circular references.

kstreeter:
Hi @cmathis, can you confirm that the latest in #377 addresses the issue?

jwen-adobe:
@kstreeter @cmathis I included the changes in my converter test and it looks good: https://git.corp.adobe.com/experience-platform/sdm-specification/tree/issue-371/json/release

kstreeter:
Thanks @jwen-adobe I will pull the changes

cmathis:
@kstreeter  - Hold off on pulling this in. It's not quite right - the mutationevent is not including all of the standard context/experienceevent fields as it should. Looking into a fix right now.

cmathis:
@kstreeter  - Here is the problem. It was decided a few weeks ago that we wanted a single experienceevent schema to be ingested and references by the Profile Service. To do this, we adjusted the tooling to automatically include the solution fields into the standard context/experienceevent. In doing this we no longer have a base schema without that campaign mutation field. There is no way to create a new mutationevent schema that includes everything from the original base context/experienceevent plus a portion of the campaign extension fields. The only way we are seeing to do this is to define a new campaign/baseevent schema where we copy and paste the fields from the original context/experienceevent (without the solution fields) and have the mutationevent include it along with the portion of campaign fields. Yes, the schemas could get out of sync over time, but until we have non-extended base to reference we cannot think of another way. 

The is the reason in 0.9.9 we kept the context/experienceevent non extended and instead used an adobe/experience/experienceevent as the uber schema. If doing this we would have not problem, but then we are back to having two different types of experienceevent schemas.

kstreeter:
[...]

PR #377, opened 2018-05-31
Issue 361 - Fix circular reference in Campaign EE extension

kstreeter:
Please link to the issue: #361 

This fixes the circular reference by breaking out the main body of the extension from the portion containing the circular field.


cmathis:
Found a problem with this last PR - please hold off

Issue #392, opened 2018-06-12
Utilize the created/modified dates from auditable


jwen-adobe:
## What are the schemas that are affected by the issue
-common/address
-context/person
-context/profile

Currently, the platform xdm schemas abobe do not utilize the created/modified dates from "http://ns.adobe.com/adobecloud/core/1.0#/definitions/date-properties" in the auditable schema. 

We need to include those per Campaign's request.

## What are examples of products that are impacted by the issue

platform xdm and tooling

PR #393, opened 2018-06-12
Utilize the created/modified dates from auditable for profile, address and person schemas

jwen-adobe:
This PR links to the issue #392 

@kstreeter @trieloff @cmathis @cdegroot-adobe @saurabhere 

Currently, the platform xdm schemas below do not utilize the created/modified dates from "http://ns.adobe.com/adobecloud/core/1.0#/definitions/date-properties" in the auditable schema.

-common/address
-context/person
-context/profile

We need to include those per Campaign's request.

### DIFF
--- a/schemas/common/address.schema.json
+++ b/schemas/common/address.schema.json
@@ -99,7 +99,7 @@
       "$ref": "https://ns.adobe.com/xdm/common/geo"
     },
     {
-      "$ref": "https://ns.adobe.com/xdm/common/auditable#/definitions/auditlog"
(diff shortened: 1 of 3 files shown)


cdegroot-adobe:
@kstreeter - Please include for June 15th milestone, this is really a fix.
I think the pattern should be that we include the auditable schema when we have the `"meta:auditable": true,` set.

kstreeter:
I am good with this, any feedback @trieloff @lrosenthol ?

Issue #397, opened 2018-06-15
Fixing new campaign extension schemas' $id namespace


jwen-adobe:
## What are the schemas that are affected by the issue
-xdm/extensions/adobe/experience/campaign/orchestration/reportingeventmetrics.schema.json
-xdm/extensions/adobe/experience/campaign/orchestration/reportingexternalevent.schema.json

## What are examples of products that are impacted by the issue
platform xdm and tooling

PR #398, opened 2018-06-15
Issue397: Fix $id namespace for newly added campaign extensions

jwen-adobe:
This PR links to the issue #397 

@kstreeter @trieloff @cmathis @cdegroot-adobe 

Please review and approve for 0.9.9.2 release.

This is to fix the $id namespace of the newly added campaign extension for platform tooling.


jwen-adobe:
@kstreeter I see in-consistency here related to the newly added campaign extensions under orchestration folder, for example, the reportingevent's $id is "https://ns.adobe.com/experience/campaign/orchestration/reportingevent" while the others does not contain orchestration in it. 

Currently, the conversion translated the $id to the file location to deref and build protobuf files etc. If we do not make the $id similar like other existing solution schemas, I will have to add special cases in the conversion to handle this and we may have more special cases down the road which is not ideal.

jwen-adobe:
I also see the following two $ids are inside the $ref of reportingevent.schema.json, so my fix is fixing their typos.

-"https://ns.adobe.com/experience/campaign/orchestration/reportingeventmetrics"
-"https://ns.adobe.com/experience/campaign/orchestration/reportingexternalevent"

kstreeter:
Thanks @jwen-adobe, I am going to push this fix, but will file an issue about why we are putting these into their own namespace.

jwen-adobe:
@kstreeter ok thanks. Just fyi that we have $id matching the folder structures for almost all other schemas (except /adobe/ portion for special handling). These are the first ones created under their own sub-folders. If we are expecting $id can be anything not related to its folder structure, then the conversion will need to build a map to maintain the $id/schema_locaiton mapping and the protobuf conversion also need to be changed to adapt this.

PR #404, opened 2018-06-15
Removed Circular Reference in Org and Geo Units

cdegroot-adobe:
Remove the concept of a Parent to Geo and Org for now as we workout the correct longer term structures for this. Keeping the original breaking change.


cdegroot-adobe:
@kstreeter  - This is the fix for the circular ref that allows us to decide the right implementation later with a reduced change of a breaking change.

PR #411, opened 2018-06-18
Payload value types

harleensahni:
Please link to the issue #…
Add a concept of payload value types to segment membership. Need feedback on if this is an acceptable use of OneOf or if we prefer to enforce restrictions on multiple values in APIs and not the json schema itself.


schemas/context/experienceevent.schema.json
```diff
@@ -118,6 +118,15 @@
           "title": "Advertising",
           "$ref": "https://ns.adobe.com/xdm/context/advertising",
           "description": "The information related to advertising activity related to the experience event"
+        },
+        "xdm:segmentMemberships": {
```
harleensahni: Move xdm:segmentmemberships in experience event to PR request around profile stitch (profile merge)

schemas/context/segmentmembership.schema.json
```diff
@@ -29,6 +29,66 @@
...
+                  "const": "number",
+                  "description": "The type of payload."
+                }
+              },
+              "required": ["xdm:payloadNumberValue", "xdm:payloadType"]
+            },
+            {
+              "type": "object",
+              "properties": {
+                "xdm:payloadPropensityValue": {
```
harleensahni: The thinking is that propensity can live outside of payload and belongs more with the core of a what segment is.

kstreeter:
Hey @harleensahni, in general I am supportive of this.  I did have a question about the "payload" concept.  Is there a more semantically specific way to describe this concept?  It seems like this is describing the "strength" of the membership, or some characteristic of placement into the segment.  Can we use language that describes that?

Regarding the "oneOf" construct, I think this is a good idea.  One thing we should do is out in some unit tests that make sure that we don't allow property names to be defined differently across choices.  I can see this technique being abused :)

harleensahni:
@kstreeter That is a good question. Propensity or strength of membership may more naturally fall as it's own property. But we have use cases for other values that should be tied to the membership and have the same ttl. Here are some examples:

- Instead of creating segments that have distinct ranges that are not granular to capture age, have an age segment that stores an integer value in the membership (where this granularity is helpful for down stream data science that might be done with this data)
- Storing gender as an enum
- Rewards level for a hotel (silver, gold, platinum)

With these, and other use cases, I am hoping to come to a generic solution to this problem of storing values on segments that is broader than just propensity.

schemas/context/experienceevent.schema.json
cluby-adobe: @harleensahni, What would you think about breaking this up into 2 PRs.  One to just add `xdm:segmentMemberships` to ExperienceEvent and a separate one for the payloads?  Then we can more forward with mapping the segment IDs Analytics and others get from AAM now to XDM while the new payload functionality gets worked out.

harleensahni: Will do that, see #444 . Will remove segment membership from this one.

schemas/context/segmentmembership.schema.json
```diff
@@ -29,6 +29,83 @@
...
+                  "type": "string",
+                  "const": "string",
+                  "description": "The type of payload."
+                }
+              },
+              "required": ["xdm:payloadStringValue", "xdm:payloadType"]
+            }
+
+          ]
+        },
```
rkpandey: I don't see any metadata related to the nature of payload itself. The metadata is related to the payload value. Just by looking at data, I won't be able to find out what it represents.

harleensahni: I think this belongs in the segment definition (which we haven't defined an xdm about yet). I think you'll have to join between segment definition and segment membership. Does that fit with how you see it, @rkpandey ?

harleensahni:
Propensity will live here instead: https://github.com/adobe/xdm/pull/452

this pull request (411) is concerned with everything other than propensity.

harleensahni:
We're moving towards storing this payload as map to allow storing multiple values that are semi structured.

harleensahni:
We will proceed with this PR with its simple mechanism for storing a single value for the most basic use cases, like propensity. #452 has been closed and will be added back to this PR

schemas/context/segmentmembership.schema.json
harleensahni: We revisited this with @kstreeter and decided to bring back propensity here.

harleensahni:
@fmeschbe, I think @chrisdegroot had meant for you to review #452, but it's been supplanted by #411 (which #452 itself supplanted).  
cc: @kstreeter

Issue #419, opened 2018-06-24
Change Identity schema identifier to unstructured string from URI


kstreeter:
After working with the current definition of `Identity` it has been decided that the current definition of `Identity.@id` is too restrictive due to the need to ensure that all characters in a contributed identity are URI safe.  The preference is to define this field in a less structured way, and let implementations manage the details of the representation.

Since `@id` is defined as a URI, the suggestion is to change this field from `@id` to `xdm:id` and designate it as a string versus a URI.  This is consistent to how we have represented unstructured identifiers in other places in XDM.

## What are the schemas that are affected by the issue
Identity, EndUserIds, Profile, ExperienceEvent (and their extensions)

## What are examples of products that are impacted by the issue
Analytics, Campaign, Ad Cloud, Target

jbeckert:
Is this explicitly only for `Identity`? Meaning all other usages of `@id` remain unchanged with the same `"type": "string"` and `"format": "uri"`?

jbeckert:
Does `xdm:id` for `Identity` need a `"pattern"` property to reject ids with prohibited characters, e.g. characters that would mess up the usage of the value in URL path components?

lrosenthol:
Please don't use xdm:id - it will cause too much confusion with @id.

fmeschbe:
@jbeckert:

> Is this explicitly only for Identity?

Yes, all others keep the `@id` property

> Does xdm:id for Identity need a "pattern" property to reject ids with prohibited characters

That is a good call-out. We should indeed to this at least in the description if not formalizing with a pattern. Your intent would be for this to be the "URL-safe" characters, correct ?

For example we could amend the description to state, that only characters should be chosen for the `xdm:id` which do not need to be percent-encoded when used in an URL path.

Or we could (in addition) declare this explicit pattern:
 
```

pattern = "^[a-zA-Z0-9._~!$&'()*+,;=:@-]+$"
```

This is the full ability of RFC 3986 `pchar` except `percent-encoded` which would be forbidden:

    pchar         = unreserved / pct-encoded / sub-delims / ":" / "@"
    unreserved    = ALPHA / DIGIT / "-" / "." / "_" / "~"
    sub-delims    = "!" / "$" / "&" / "'" / "(" / ")" / "*" / "+" / "," / ";" / "="

I think it is debatable to support all those characters and we might want to actually limit to just the `unreserved` set.

fmeschbe:
[...]

PR #420, opened 2018-06-24
Issue 419 - changing Identity id from URI format to plain string

kstreeter:
This PR is to address issue #419, where the definition of `@id` in the `Identity` schema is too constrained for real usage.  This PR changes the field from `@id` to `xdm:id`, and updates all of the examples.


jwen-adobe:
@kstreeter This looks good from tooling perspective based on my test.

cdegroot-adobe:
Technically this PR does remove the need for URIs, but the examples still are URIs.

What is the reason we are not using URL encoding? The cases where there are non URL safe characters is a corner case.

saurabhere:
Looks good. So, `_id` becomes `id` in XED for identities. Hope this is communicated to all the consumers(UPS, siphon)

saurabhere:
Can this also include similar changes in the `orgunit.schema` and `geounit.schema` ?

cdegroot-adobe:
@saurabhere this change is limited to the Identity schema. The standard in XDM is to use @id for the globally unique identifier of all entities.

cdegroot-adobe:
@fmeschbe - we need to include this in the July 16th milestone.

PR #423, opened 2018-06-27
Add post office box to align with Microsoft CDM

cdegroot-adobe:
Improving common\address with Post Office Box property.


cdegroot-adobe:
@fmeschbe - can you please put this in the July 16th milestone.

Issue #439, opened 2018-07-19
Move/Rename datasetID


cdegroot-adobe:
## What are the schemas that are affected by the issue
extensions/adobe/experience/experienceevent.schema.json

## What are examples of products that are impacted by the issue
Unified Profile

Introduced in 0.9.3 was a new field that is added to context/experienceevent named datasetID. This field has unique usage properties that cause concern at the level it was exposed and the generic naming of the field.

Decision by the impacted teams is to move this field to a more constrained location and rename it in a more constrained fashion.

fmeschbe:
PR #440 has been signed-off of and merged. Therefore closing this issue as well.

PR #440, opened 2018-07-20
Issue 439 correct representation of datasetID

cdegroot-adobe:
Please link to the issue #439.

Improve/correct the representation of Dataset ID in the Adobe Experience namespace.

Final proposal


jwen-adobe:
@cdegroot-adobe for the schema "https://ns.adobe.com/experience/profile/experienceevent", I am not seeing "meta:extends": ["https://ns.adobe.com/xdm/context/experienceevent"] included. If you need this profile/experienceevent schema to be included in "context/experienceevent" as well, you will need this meta:extends. Or it will be only in "adobe/experience/experienceevent" schema.

tcuevas-adobe:
👍

Issue #434, opened 2018-07-14
Clarify @id for Identity


fmeschbe:
Issue #419 (PR #420) propose to drop the URI formatted `@id` property in favor of a simple string `id` property. 

There is a train of thought along the lines of _All schemes SHALL have the @id property. But it MAY be optional_ (quoting @lrosenthol). I am sympathetic to this thought.

Essentially, the schema looks like

```json
{
  "xdm:id": { "type": "string" },
  "xdm:namespace": { "$ref": "https://ns.adobe.com/xdm/context/namespace" },
  "xdm:xid": { "type": "string" },
  "xdm:primary": { "type": "boolean", "default": false }
}
```
(where `xdm:namespace` is a single mandatory property `xdm:code` of type `string`)

The issues are:

* None of the fields is mandatory, so technically an empty object is a "valid" Identity
* `xdm:id` requires `xdm:namespace` to be a proper identity
* `xdm:xid` by itself is unique and is a mapping for `xdm:id` plus `xdm:namespace`
* `@id` is not defined for this schema

So we have to fix the real optionality of the properties and we have to define how to look at the `@id` property.

First of all, having an `@id` property makes sense as a unique identity of something. But having this property optional does not really add value to such an important field as an identity.

Proposal for discussion:

> For the `@id` property we could consider defining an URN namespace, say _adobe-id_ which requires the `xdm:id` and `xdm:namespace.xdm:code` properties in a form like `urn:adobe-id:<xdm:namespace.xdm:code>:<xdm:id>`. For example `urn:adobe-id:Email:someone@example.com`

(Such a definition, of course, would require the values for `xdm:code` and `xdm:id` to be restricted. See also #432)

*Note*: This issue summary represents the current state of knowledge and discussion. As we evolve the discussion around this issue, the scope of this issue may shift.

## What are the schemas that are affected by the issue

[Identity](https://github.com/adobe/xdm/blob/master/schemas/context/identity.schema.json) I in fact see a number of practical problems.

## What are examples of products that are impacted by the issue

Analytics, Campaign, Ad Cloud, Target

PR #441, opened 2018-07-23
Make the change - @id present but optional

lrosenthol:
This addresses #434

### DIFF
--- a/docs/id.md
+++ b/docs/id.md
@@ -1,6 +1,6 @@
 ## Identifying Entities
 
 Instances of most schemas in XDM are uniquely identifiable based on the value of the `@id` attribute.
-This attribute is always a URI, but may not be present or required in all schemas.
+This attribute is always a URI, and will always be present all schemas.  However, some schemas do not require that all instances be identifiable, so the `@id` attribute will be optional in those cases.
 
-Schemas that have an `@id` attribute are considered identifiable, schemas that do not have this attribute are typically only used as child-objects in identifiable schemas.
+Schemas that have a required `@id` attribute are considered identifiable, schemas where this attribute is optional are typically only used as child-objects in identifiable schemas. However, since not all uses for the schema are known, we hace left it present and optional should a given use case require the need to make it identifiable.


cdegroot-adobe:
@lrosenthol - can you please add to the description of this PR the issue we are solving. There are so many discussions on this topic I think it would be helpful to have it clearly articulated here.

lrosenthol:
The reason we are doing this is because it can't be done after the fact.

`@id` is [defined in](https://w3c.github.io/json-ld-syntax/#node-identifiers) the JSON-LD spec as the way in which you refer to an node externally - basically giving *that instance* of that schema a unique ID.  While it may be that in Adobe's _current_ use of a given schema, they don't ever need to refer to a specific instance - we do not know how other users of these schemas will use them.

And since you can't add `@id` to an instance of a schema that doesn't have it defined in the schema itself (if you assume that somewhere in the flow there may be validation) - we must define it in all schemas.

However, just because it is defined in each, doesn't mean it has to be there - which is why it may be optional.

cdegroot-adobe:
@lrosenthol - Thank you for the reasoning. One clarification, we use the term "schema" here, but we really mean every instance of an object and maybe array too.

e.g looking at the assets/asset.schema.json we would want to add `@id` in both these cases which are not schema "files"
```
        "xdm:milestone": {
          "title": "Milestone",
          "description": "A label given to a version of the asset.",
          "type": "object",
          "properties": {
<--- Add @id here? -->
            "label": {
              "title": "Milestone Label",
              "type": "string",
              "description": "Label for the version of an asset"
            },
            "description": {
              "title": "Milestone Description",
              "type": "string",
              "description": "Description for the version of an asset"
            }
          }
```

And 
```
        "xdm:aliasIDs": {
          "type": "array",
          "meta:usereditable": false,
          "title": "Alias IDs",
          "items": {
            "properties": {
<--- Add @id here? -->
              "system": {
                "title": "External System",
                "type": "string",
                "description":
                  "A string used to identify the external systems like google, facebook etc."
              },
              "id": {
                "type": "string",
                "title": "External ID",
                "description":
                  "An ID under which external systems track the asset."
              }
            }
          },
```

Thoughts?

lrosenthol:
Actually, when I am using it, I mean the schema definition as contained in a `.schema.json` file.  Perhaps that is what is confusing you?

So in your examples, you wouldn't add @id into either.  You would have it at the top of an instance of a schema.

cdegroot-adobe:
That seems a bit arbitrary considering we do not necessarily break down schemas into sub/abstract schemas consistently across committing teams. On the XC side we tend to prefer reusability of schemas, and thus have normally got only one properties definition per schema. 

I certainly read the intent of `@id` to mean node, not schema file. So any collection of properties that represent something, an `aliasID`, or a `milestone` may need a predictable address if we follow the thinking of the proposal. 

I suppose I'm saying the proposal is still a but unclear in my mind because you are saying an "`@id` needs to be added if we somehow deemed it right to break out an object in a dedicated schema file." Which in itself is fairly stylistic decision, not a canonical one.... or else I'm showing my ignorance in this area....

lrosenthol:
I agree with you @cdegroot-adobe that you can use a schema either by itself or as a child of some other schema.  However, you seem to be considering properties as instances of schemas, but they are not. 

On the other hand, consider something like a directory listing that has a list of assets such as:

`
"xdm:dirListing" : [
{ xdm:asset { "@id" : "12345", .... },
{ xdm:asset { "@id": "67890", .... }
]
`

Since xdm:asset is an actual schema, each instance of it can then have an `@id` associated with - even though it is inside/child of the xdm:dirListing

ogoldman:
I want to point out that how we organize schema definitions into files *is* a canonical decision, in that references to those definitions from other schemas reference that filename. I worry this may be a looming issue for us, at least where we currently have done a poor job of grouping definitions into files.

PR #444, opened 2018-07-24
profile stitch

harleensahni:
This PR addresses issue #468.

Add a concept of profile stitch / merge rules to experience event. I had to introduce a concept of profile stitch identity similar to what we have for segment identity, but I'd prefer if we just took a much simpler approach with segment and profile ids as simple strings. 

I also add segment membership to experience event and modify identity to support multiple ids for the same namespace (which is a hard requirement for merging).


schemas/context/identity.schema.json
```diff
@@ -20,6 +20,14 @@
           "type": "string",
           "description": "Identity of the consumer in the related namespace."
         },
+        "xdm:additionalIds": {
```
cdegroot-adobe: additionalIDs

schemas/context/identity.schema.json
```diff
@@ -20,6 +20,14 @@
...
+          "title": "Additional Identifiers",
+          "type":"array",
+          "items": {
+            "type": "string"
+          },
+          "description": "If consumer has more than one of the given id, additional ids are listed here."
```
cdegroot-adobe: If endUser has more than one identity in the namespace, additional identity ids are listed here.

schemas/context/profileStitch.example.1.json
```diff
@@ -0,0 +1,22 @@
...
+  "xdm:profileStitchID": {
+    "@id": "https://data.adobe.io/entities/profileStitchIdentity/1",
+    "xdm:namespace": {
+      "xdm:code": "AAM"
+    }
+  },
+  "xdm:version": "1.0",
+  "xdm:endUserIds": {
+    "https://ns.adobe.com/experience/mcid": {
+      "@id": "https://data.adobe.io/entities/identity/92312748749128",
```
cdegroot-adobe: this has now changed to "xdm:id" as of 0.9.3

schemas/context/profileStitch.example.1.json
```diff
@@ -0,0 +1,22 @@
...
+      "additionalIds": [
```
cdegroot-adobe: additionalIDs

schemas/context/profilestitchidentity.schema.json
```diff
@@ -0,0 +1,41 @@
...
+    ],
+    "$id": "https://ns.adobe.com/xdm/context/profileStitchIdentity",
+    "$schema": "http://json-schema.org/draft-06/schema#",
+    "title": "Identity",
+    "type": "object",
+    "description": "ProfileStitchIdentity is used to clearly distinguish Profile Stitching from multiple sources. ProfileStitchIdentity is established by an Profile Stitch identity provider, which itself is referenced in the `namespace` attribute. Within each `namespace`, the profileStitchIdentity is unique.",
+    "definitions": {
+        "profileStitchIdentity": {
+            "properties": {
+                "@id": {
```
cdegroot-adobe: `@id`->"xdm:id"

schemas/context/profilestitchidentity.schema.json
```diff
@@ -0,0 +1,41 @@
...
+            }
+        }
+    },
+    "allOf": [
+        {
+            "$ref": "#/definitions/profileStitchIdentity"
+        }
+    ],
+    "meta:status": "experimental"
+}
```
cdegroot-adobe: add newline please.

schemas/context/experienceevent.example.1.json
```diff
@@ -114,5 +114,48 @@
   },
   "xdm:marketing": {
     "xdm:trackingCode": "marketingcampaign111"
-  }
+  },
+  "xdm:profileStitching": [
+    {
+      "xdm:profileStitchID": {
+        "@id": "https://data.adobe.io/entities/profileStitchIdentity/1",
```
cdegroot-adobe: `@id`->"xdm:id"

cdegroot-adobe:
@harleensahni - I really like the additionalIDs. To simplify things I recommend that you pit that in a separate PR as that is quite a distinct thing and we should be able to get closure on it quickly.

I do not quite understand the concept of the core of this addition to ExperienceEvent. I am assuming that the merge rules and profile stitching are added during data-collection. So this is really a log of that. Can you explain the interaction flow a bit please?

schemas/context/experienceevent.schema.json
```diff
@@ -126,6 +126,31 @@
           "description": "The media activity information related to the experience event"
         }
       },
+      "xdm:profileStitching" : {
+        "title": "Profile Stitching",
```
lrosenthol: This seems like a very Adobe-centric thing, yes?  If so, it belongs in an extension and *not* in the primary schema

schemas/context/experienceevent.schema.json
```diff
@@ -126,6 +126,31 @@
...
+        "description": "Details about the ids that were joined by profile stitching.",
+        "$ref": "https://ns.adobe.com/xdm/context/profileStitch"
+      },
+      "xdm:segmentMemberships": {
```
lrosenthol: What is a "segment", where it is defined, is it specific to an Adobe workflow or generic?

schemas/context/profileStitch.schema.json
```diff
@@ -0,0 +1,40 @@
+{
+  "meta:license": [
+    "Copyright 2018 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/context/profileStitch",
```
lrosenthol: Also needs `@id`, as per new plans

schemas/context/profileStitch.schema.json
```diff
@@ -0,0 +1,40 @@
...
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Profile Stitch",
+  "type": "object",
+  "description": "Details about the ids that were joined by profile stitching.",
```
lrosenthol: This seems like a very Adobe-centric thing, yes? If so, it belongs in an extension and not in the primary schema

schemas/context/profilestitchidentity.schema.json
```diff
@@ -0,0 +1,41 @@
+{
+    "meta:license": [
+        "Copyright 2018 Adobe Systems Incorporated. All rights reserved.",
+        "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+        "you may not use this file except in compliance with the License. You may obtain a copy",
+        "of the License at https://creativecommons.org/licenses/by/4.0/"
+    ],
+    "$id": "https://ns.adobe.com/xdm/context/profileStitchIdentity",
+    "$schema": "http://json-schema.org/draft-06/schema#",
```
lrosenthol: also needs `@id`

schemas/context/profilestitchidentity.example.1.json
```diff
@@ -0,0 +1,6 @@
+{
+  "@id": "https://data.adobe.io/entities/profileStitchIdentity/1",
+  "xdm:namespace": {
+    "xdm:code": "AAM"
```
lrosenthol: Don't you also need an `xdm:xid` here?  What use is the namespace without the thing you are namespacing?

schemas/context/profileStitch.schema.json
```diff
@@ -0,0 +1,40 @@
+{
```
cdegroot-adobe: Lower case the name of this schema, and the example too please.

schemas/context/experienceevent.schema.json
harleensahni: I don't think so. It's about which set of profiles segmentation is running against. Adobe has some functionality around this, but I think others in the industry do too.

lrosenthol: well, I would want to understand how it fits into the larger industry context before accepting this in a common schema.

schemas/context/experienceevent.schema.json
harleensahni: It's not defined yet in it's own xdm. This pull request didn't introduce the concept of segment membership, it already exists on profile and the PR for the segment membership is merged. I'm just adding it experience event. It's meant to be generic, but this will definitely be used by Audience Manager and I think Target.

schemas/context/experienceevent.schema.json
harleensahni: Understood, should we talk offline? Or do you have something in particular you're looking for?

lrosenthol: Certainly happy to chat offline, but what I am looking for is either an existing standard for this that we can point to from this common schema...or we move it to an Adobe extension.

schemas/context/experienceevent.schema.json
lrosenthol: Got it.   OK - no issue there.

schemas/context/profilestitchidentity.example.1.json
harleensahni: Wouldn't this as-is namespace `@id`.

lrosenthol: @id is the ID of that particular instance of that schema - if that is what you are doing, then yes, I would agree.  I thought you were referring to some other object...which would be via xdm:id

harleensahni:
@cdegroot-adobe For splitting into two different PRs, one for addiitonalIDs and one for profile stitching, is it necessary? I ask because the profile stitching PR is useless without the additionalIDs and I want to call out as part of the PR for profile stitching that we'll be using additiionalIDs to have multiple ids for the same namespace.

As far as the question about what this PR does with profile stitch / merge rule and the flow:

1. Merge rules / profile stitch are defined elsewhere (not addressed in this PR really, apart from declaring the ID format for those). This is similar to the punting approach we've taken with segment memberships but not defining segment yet.
2. What is recorded in the Experience Event here is the result of a specific invocation of a profile stitch / merge rule that was done to connect this event and the ids in it to other profiles. Basically, what's recorded is that profile stitching that was done in this way (profileStichId, [linkedid_1, linkedid_2, linkedid_n]). And then adding segment membership on experience event with an additional field to refer back to the profile stitch id to find what other profiles that segment realization applies to (besides the one that came in on the event).

lrosenthol:
@harleensahni what do you mean by "multiple IDs for the same namespace"?  That just seems wrong.

As long as the "merge rules" you describe could be implemented by anyone else using the same schemas (but perhaps with their own ID format) - that sounds reasonable.  If this is Adobe-centric, it needs to move to an extension.

harleensahni:
@lrosenthol 

> What do you mean by "multiple IDs for the same namespace"? That just seems wrong.

I think it'll be helpful if I give an example. Here the namespace will be ECID, and id will the cookie associated with the ECID and well us two values, `ECID_1` (for device 1) and `ECID_2` (for device 2). There'll be another namespace for identity of `CRMID` (possibly based off a hash of customer's email).  For this example, we'll have one CRM IDs, `CRMID_1`. 

Let's say that we've received one experience events in the past, `EE_1` . On `EE_1`, we saw that the event came from `ECID_1` and the user was logged in with `CRMID_1`.  Now we have another experience event `EE_2`, the event came from `ECID_2` and the user logged in was the same user with `CRMID_1`. If we have a profile stitching that's operating where it can join ECIDs based on the logged in CRMID, you will have two ECIDs of value `ECID_1` and `ECID_2` in the same namespace. This is recording that. 

As far as what joining those two ECIDs, one thing is segmentation that operates against the joint profiles of `ECID_1` and `ECID_2`.  It's important to know which profiles were joined and to have that recorded so that you can record the segmentation back on both `ECID_1` and `ECID_2`.

lrosenthol:
ECID isn't a real namespace - at least as far as XDM is concerned.  It's something specific to how you define the string that represents your IDs, yes?  Since IDs are just arbitrary strings - that seems out of scope for XDM proper since it doesn't care what you stick in there...

harleensahni:
It's defined in XDM (under it's old name of MCID): https://github.com/adobe/xdm/blob/master/schemas/context/enduserids.schema.json#L20 
When I'm saying namespace, I mean the identity space for end user ids.

lrosenthol:
Gotcha - thanks for that link!  (just filed an issue against that schema, #449 , but that's separate)

OK - so if you need multiple items, use an array.  What am I missing??

cdegroot-adobe:
@lrosenthol the design of the endUserIDs is very carefully thought out. Having specific properties in known locations is very important for efficiency when there are very large numbers of records, as is the case with ExperienceEvent. That is why we have this design and it is predicated on multiple identities being an edge case. Changing how endUserIDs implements this is *not* and option. The additionalIDs is a way of catering for the edge case and is fine because they are not used in the same way as the "xdm:id" for the identity. additionalIDs is an array, so covers this case. As I say this specific area is quite carefully curated and what Harleen is doing here is the approach that multiple teams have discussed and come to agreement on.

harleensahni:
The reason I chose to model this using EndUserId here is that is what ExperienceEvent uses for the primary ids that come in with the event here: https://github.com/adobe/xdm/blob/master/schemas/context/experienceevent.schema.json#L43. So I want this to be consistent with that. I think EndUserId was modeled the way it is to support being able to do lookups based on a namespace very quickly since it'll be it's own column on disk. It would be weird if I modelled this as an array of EndUserIds since you could have multiple ids under the same namespace in separate instances of EndUserIds (and you can have multiple different namespaces in EndUserIds).  

By adding the array of additionalIDs to EndUserIds, you may not be able to quickly find if a particular id is in a namespace since it might be in the array, but you can at least still tell very quickly if an event has ANY ID in a particular namespace, which is valuable.

lrosenthol:
Well, personally, I am not thrilled with this design but I will accept that we can't change the model at this time.

schemas/context/identity.schema.json
```diff
@@ -20,6 +20,14 @@
           "type": "string",
           "description": "Identity of the consumer in the related namespace."
         },
+        "xdm:additionalIDs": {
```
biswas-adobe: Additional Ids are also needed to support multiple ECID --> CRMID mappings that are provided by co-op and private graph customers. The support for additional Identities should be done outside of the Identity object where each id specified, is of type xdm:Identity. Either in the EndUserIds structure or in a separate Identity extension for ExperiencEvent.

schemas/context/experienceevent.example.1.json
```diff
@@ -114,5 +114,48 @@
...
+  "xdm:segmentMemberships": [
+    {
+      "xdm:segmentID": {
+        "@id": "https://data.adobe.io/entities/identity/92312748749128",
+        "xdm:namespace": {
+          "xdm:code": "AAM"
+        }
+      },
+      "xdm:profileStitchID": {
+        "xdm:id": "https://data.adobe.io/entities/profileStitchIdentity/1",
```
harleensahni: need to change this back to `@id`

schemas/context/identity.schema.json
cdegroot-adobe: @harleensahni  please review with Rahul, I agree on your proposal, this is likely just something you need to talk through.

schemas/context/profileStitch.schema.json
cdegroot-adobe: Agreed, we need to get that approved first and then we will do the needful. I would like Kevin to return for the `@id` decision and exactly how we roll it out. In the XC we cannot automatically add them, so we need to do something to limit the explosion of fields in our operational store.

schemas/context/profileStitch.schema.json
cdegroot-adobe: Please see earlier comments on this. I am confident this is not an Adobe-centric things at all.

schemas/context/profilestitchidentity.schema.json
cdegroot-adobe: see earlier comment on it, we will pick up all the `@id`s in a dedicated PR.

schemas/context/experienceevent.schema.json
lrosenthol: @cdegroot-adobe if I do a web search on "profile stitching", only one of the hits is about this (and it actually calls it visitor stitching).
Is there any existing industry reference document that we can point to that defines this?

schemas/context/profilestitchidentity.schema.json
harleensahni: Done

schemas/context/experienceevent.example.1.json
```diff
@@ -114,5 +114,48 @@
   },
   "xdm:marketing": {
     "xdm:trackingCode": "marketingcampaign111"
-  }
+  },
+  "xdm:profileStitching": [
```
mohrahit: A couple of questions:
- What is the use case for adding profileStitching to ExperienceEvent? 
- Is the user expected to populate this via ETL jobs?

schemas/context/experienceevent.example.1.json
```diff
@@ -114,5 +114,48 @@
...
+  ],
+  "xdm:segmentMemberships": [
+    {
+      "xdm:segmentID": {
+        "@id": "https://data.adobe.io/entities/identity/92312748749128",
+        "xdm:namespace": {
+          "xdm:code": "AAM"
+        }
+      },
+      "xdm:profileStitchID": {
```
mohrahit: Is there a reason to add this to each segment realization and each ExperienceEvent? This information can be looked up on demand from the segment metadata.

schemas/context/experienceevent.example.1.json
harleensahni: To support profile stitching in realtime for use with realtime segmentation
Someone could populate this with an ETL job if they wanted, but that wasn't the use case I was considering. I think it's open enough to support that.

schemas/context/experienceevent.example.1.json
harleensahni: @mohrahit, another alternative existed also. That was to have segmentmembership be a property under each profile stitch here in ExperienceEvent. That felt too nested and more importantly, assumed that all segmentation happened with profile stitching.

schemas/context/identity.schema.json
harleensahni: @biswas-adobe  has concerns with Identity itself being modified. It's beginning to look like a better approach to using EndUserID here (as well as in Experience Event) is to go with an array of Identities. 

Modeling with additional IDs is not compatible with #459 since it does not allow for different authentication states for each id in a namespace.

schemas/context/experienceevent.example.1.json
harleensahni: @mohrahit, are you okay with this approach?

schemas/context/identity.schema.json
kstreeter: @harleensahni I also feel like if would be better to use an array of Identity.  This is how the same problem was handled in Profile.

harleensahni: @kstreeter @biswas-adobe I've changed this to be an array of Identity

PR #445, opened 2018-07-25
Change repo:assetID to repo:id

mburbidgadobe:
Every content type (file, composite, entity, directory) in the Platform API has an identifier that uniquely identifies the content item. The property name of this identifier is currently "repo:assetID". The term asset does not particularly fit many content item types such as directories, datasets, and other entities. Furthermore, since this identifier is specific to the repository, it seems more appropriate to use the property name "repo:id" for the repository identifier. A few of the data XDM use the term assetID to mean a different identifier.

This change set updates the repository XDM to use "repo:id" for the repository identifier property name.

Issue: https://jira.corp.adobe.com/browse/ACP-217

### DIFF
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -22,7 +22,7 @@ Some XDM schemas might be declared non-extensible, which means that any property
 
 XDM generally separates two kinds of properties:
 
-1.  Standard properties are part of the XDM specification, and follow a pattern of `prefix:name`. For instance, `repo:id` is the unique identifier of assets in the content repository.
+1.  Standard properties are part of the XDM specification, and follow a pattern of `prefix:name`. For instance, `repo:id` is the unique identifier of assets in a repository.
 2.  Extension properties are not part of the XDM specification and have been defined by a customer, a partner, or by Adobe for things that are specific to one implementation. Extension property names are always URIs like `http://example.com/ns/xdm/my_property`
 
 XDM comes with a small list of allowed prefixes that correspond to the namespace URIs of the standards that XDM is incorporating.
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -22,7 +22,7 @@ Some XDM schemas might be declared non-extensible, which means that any property
 
 XDM generally separates two kinds of properties:
 
-1.  Standard properties are part of the XDM specification, and follow a pattern of `prefix:name`. For instance, `repo:assetID` is the unique identifier of assets in XDM.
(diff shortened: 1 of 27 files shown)


docs/extensions.md
```diff
@@ -22,7 +22,7 @@ Some XDM schemas might be declared non-extensible, which means that any property
 
 XDM generally separates two kinds of properties:
 
-1.  Standard properties are part of the XDM specification, and follow a pattern of `prefix:name`. For instance, `repo:assetID` is the unique identifier of assets in XDM.
+1.  Standard properties are part of the XDM specification, and follow a pattern of `prefix:name`. For instance, `repo:id` is the unique identifier of assets in the content repository.
```
lrosenthol: replace "the content" with "a" (since we're talking about any type of repo, anywhere)

fmeschbe:
This opens a can of worms, I fear. We just introduced the property `xdm:id` into the Identity schema with PR #420 and had a good debate about whether it is ok to replace `@id` with `xdm:id`.

Now we replacing `repo:assetId` with `repo:id`. This is in and of itself is a good thing.

But now we end with two `XXX:id` properties essentially doing the same thing: Identifying the object (as opposed to identifying the message).

Can we rationalize this situation ?

/cc @kstreeter, @chrisdegroot

ogoldman:
@fmeschbe These two IDs do not have the same semantics. Repo:id is the identity associated with the binding of an asset to a "storage slot": if you delete the asset and re-upload it, it gets a new repo:id. It is used primarily to refer to something else--the thing in that storage slot, if you will.

On the other hand, xdm:id is embedded in the object it identifies. Delete and re-upload the entity, and its xdm:id doesn't change.

Finally, consider the example of an "asset reference", such as appears in directory listings. In that entity, an xdm:id identifies the reference; repo:id identifies the referenced asset.

lrosenthol:
what's to rationalize, @fmeschbe ?  repo:id and xdm:id are completely different things - they are in separate namespaces.

@ogoldman be careful - xdm:id is *not* the same as `@id`, which is what you describing (or at least it sounds like you are describing).

fmeschbe:
Thanks @ogoldman and @lrosenthol. That helps and "rationalizes" the difference and similarity for me:

- `repo:id` is uploaded by the repository as the object is uploaded (and created in the implementation)
- `xdm:id` is an identifier for an object which is not necessarily defined by the repository and not managed by repository but by an application using the repository

PR #453, opened 2018-07-30
Update copyright

cziegeler:
Please link to the issue #…


cziegeler:
Regardless of that, as there hasn't been an official 1.0 release, I think it would look strange if you release a 1.0 in 2018 and it says (C) 2017

lrosenthol:
I agree with @cziegeler - since it's not 1.0 yet, we should update...

cdegroot-adobe:
@lrosenthol  - For clarity, because the eventual question will be do we update all the copyrights in all the files. We had made the decisions to leave the copyright at the year at which the file was added to the repository, not changing them during updates etc.

Are we expressly changing that policy?

Issue #399, opened 2018-06-15
New schema to support direct marketing subscriptions


jfait:
## What are the schemas that are affected by the issue
Profile

## What are examples of products that are impacted by the issue
Campaign

New schema to register subscribes and unsubscribes to directly marketing lists

PR #458, opened 2018-08-01
Issue#399 Change context\pushNotificationToken to context\notificationSubscription

cdegroot-adobe:
Please link to the issue #399 

The Profile PushNotificationToken schema has been found to have some limitations an a new requirement has emerged for a more general purpose representation of subscriptions to notifications.

This PR rolls up changes to merge these two implementations and make the context\pushNotificationToken more generic.

The changes are only a rename and two additional properties, but the nature of this changes to encompass the new scope pretty well.


lrosenthol:
This seems to have uses beyond just marketing notifications & subscriptions, including a connection to event notifications.  @ogoldman @theManikJindal - can you two comment on whether we should have two separate items (this + one for event notifications) or we should come up with a common one?

ogoldman:
@lrosenthol This appears to be about user subscriptions to notification channels. I certainly think we should have only one model for user subscriptions, whether those subscriptions are related to marketing, collaboration (DC/CC), or something else.
I would draw a distinction, however, with client registrations for API/programmatic subscriptions, such as I/O Events. Those subscriptions cross user boundaries, and should have a different model.

cdegroot-adobe:
@kstreeter - can you please review, we need to get this campaign extension in for the next release.

extensions/adobe/experience/campaign/notificationSubscriptionTarget.schema.json
```diff
@@ -0,0 +1,38 @@
+{
+  "meta:license": [
+    "Copyright 2018 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id":
+    "https://ns.adobe.com/experience/campaign/notificationSubscriptionTarget",
```
kstreeter: I believe we are making schema URI all lower case

extensions/adobe/experience/campaign/notificationSubscription.schema.json
```diff
@@ -0,0 +1,62 @@
...
+          "type": "string",
+          "description":
+            "The topic that the enduser has subscribed to and opted into notifications for."
+        },
+        "xdm:type": {
+          "title": "Subscription type",
+          "type": "string",
+          "description": "The type of notification/subscription."
+        },
+        "xdm:subscriptionDate": {
```
kstreeter: The subscription date is at this level, but the unsubscription date is in the unsubscription details.  Is the idea that a reader knows that a user has unsubscribed if the details element is present?  If that is true, it should be described somewhere.

extensions/adobe/experience/campaign/notificationSubscriptionTarget.schema.json
cdegroot-adobe: Corrected the naming and reference

extensions/adobe/experience/campaign/notificationSubscription.schema.json
cdegroot-adobe: Added the following description to the subscriptionDate
          "description":
            "Date and time when the enduser has subscribed to a notification topic. An enduser is subscribed if they have a Subscription Date and the `unsubscriptiondetails/unsubscriptionDate` is empty."

PR #459, opened 2018-08-02
Add authentication states to identity schema

cdegroot-adobe:
Please link to the issue # 
In ExperienceEvents Identities can have different authenticated states for each identity. This is currently part of the Visitor SDK. This PR adds this feature to the `context\identity` schema.

https://marketing.adobe.com/resources/help/en_US/mcvid/mcvid-authenticated-state.html

The state enumeration is being updated to better align with current design.


lrosenthol:
The problem that I see is that you've put information about experience events into a schema (identity) which is general purpose.  Identity has nothing to do with authentication, IMO.  If the event needs that info, add it to the event.  @cdegroot-adobe

harleensahni:
Approved

cdegroot-adobe:
@lrosenthol yes, that is certainly a concern, but I do not think it is too much of a problem. Identity is a representation of context of identity (`context\identity.schema.json`). It would certainly be a serious problem in the case of a schema in `\data` or `\common`

I do not see a good way of representing this anywhere else. The relationship is to the identity instance. An endUser will have multiple identities in an ExperienceEvent, and they each may have different authenticationStates.

Do you have a thought on how to represent it in a different way?

lrosenthol:
I realize that currently the identity schema happens to live in /context, but there is nothing in the definition of identity that puts it specifically about a context.  Read the description:

> "Identity is used to clearly distinguish people that are interacting with digital experiences. Identity is established by an identity provider, which itself is referenced in the `namespace` attribute. Within each `namespace`, the identity is unique.",

It's about people.  Those people can appear in numerous other schemas - and in fact, I would expect them too.  And it's not, necessarily, about a person at a specific time/event - it's just a person.

The authenticatedState *only* applies at a specific time/event, which is why it needs to be part of the event description.   Either that, or create a new schema - `identityUse` (or the like) which combines an Identity with authenticatedState (and potentially other states relevant to a given instance usage.

schemas/context/identity.schema.json
```diff
@@ -32,6 +32,18 @@
...
             "When present, this value represents a cross-namespace identifier that is unique across all namespace-scoped identifiers in all namespaces."
         },
+        "xdm:authenticatedState": {
+          "description":
+            "The state this identity is authenticated as for this observed ExperienceEvent.",
+          "type": "string",
+          "default": "ambiguous",
+          "enum": ["ambiguous", "authenticated", "loggedOut"],
+          "meta:enum": {
+            "ambiguous": "Ambiguous",
```
harleensahni: We should at a later date give examples of ambiguous states.

biswas-adobe:
@harleensahni - it's not clear how on a website/app I can be logged in for one identity and logged out state for another. Interactions for login/logout are done using one widget typically that logs you in/out-of the website/app. Could you please explain how multiple identities can in heterogenous authentication states on the one page/app-state?

harleensahni:
@biswas-adobe One case could be where a company has single sign on but still has local logins for each one of their apps. This could happen when a company acquires another company. When you land on the app's login page, you have an identity for the single sign on that is set to authenticated but the identity for the local app set to something else. It may remember your username from last time and have the authenticated state set to logged out.  Once you click "login with single sign on" you will then have the same authentication state for both. 

Another use case is an application where you may login with a household login (like a video streaming service), but once you're logged in you can pick a "profile" to act as a particular member of that household.

harleensahni:
Another use case could be logins where it's provided by a social network. You could be logged into the social network but not the application yet.

benbeard87:
Similar to the video streaming services, companies that produce household connected devices (TVs, set-top boxes etc) may choose to use the household identity (CRM ID) to represent a specific HH device which may or may not support authentication. Here the individual identity passed on the same call in an authenticated state represents the current user of the household device.

harleensahni:
@biswas-adobe I would think it'd be available in the data source definition. That's not covered in this PR and I don't think exists in XDM yet. Otherwise, it'd be available by the API of the system that created the identity.

PR #464, opened 2018-08-06
Add Journey context and delivery attributes in experience event

ksradobe:
## What are the schemas that are affected by the issue
1) extensions/adobe/experience/campaign/experienceevent
2) extensions/adobe/experience/campaign/orchestration/reportingevent

## What are examples of products that are impacted by the issue
Voyager
Campaign

"reportingevent" schema is having all the attributes which constitute a valid reporting event. Here certain attributes can be pulled out from reporting event and make them available to other experience event. (Ex: Here, "Orchestration Context" in Campaign experience event). This will increase the consumability of common orchestration properties. 

Another change is to add "Delivery name" and "Campaign Name" which is kind of friendly identifier for the user to remember instead of simple integer id. Similarly "Delivery Label" is a descriptive name given to the marketing activity for making it easily searchable.

PR #470, opened 2018-08-15
Revert "Add Journey context and delivery attributes in experience event"

kstreeter:
We have decided to delay including 464 until after the 0.9.4 milestone.  This PR reverts those changes.

Reverts adobe/xdm#464


ksradobe:
@kstreeter I am not very clear on the reason for not including 464 in master. Can you please call out the reason.

PR #472, opened 2018-08-16
Add Journey context and delivery attributes in experience event

ksradobe:
1) extensions/adobe/experience/campaign/experienceevent
2) extensions/adobe/experience/campaign/orchestration/reportingevent

Voyager
Campaign

"reportingevent" schema is having all the attributes which constitute a valid reporting event. Here certain attributes can be pulled out from reporting event and make them available to other experience event. (Ex: Here, "Orchestration Details" in Campaign experience event). This will increase the consumability of common orchestration properties. 

Another change is to add "Delivery name" and "Campaign Name" which is kind of friendly identifier for the user to remember instead of simple integer id. Similarly "Delivery Label" is a descriptive name given to the marketing activity for making it easily searchable.

Few more changes:
changing action's type attribute to "actionType".(It is a breaking change)
ContainerID is an additional attribute that was added to campaign experience event.


ksradobe:
@fmeschbe 
PR640 has been reverted in PR470. So I just merged those 4-5 commits into one cherry picked that one to create this PR. I also modified XDM schema as per your argument in PR640 (ie to eliminate journeyContext from orchestration/experienceEvent.xml). Can you review it and confirm your approval?

fmeschbe:
Thanks @ksradobe.
 
According to @kstreeter he backed those changes out in PR #470 to be able to tag the v0.9.4 drop and open the discussion about the PR.

So this will probably be the basis for the discussion of how we want to include the features from PR #464 in the XDM proper.

ksradobe:
@lrosenthol can you review this PR?

lrosenthol:
@ksradobe it's not clear what I am supposed to review, since it seems that @kstreeter is making some changes based on the comment from @fmeschbe above...

extensions/adobe/experience/campaign/experienceevent.example.email-send.json
```diff
@@ -59,12 +59,16 @@
...
     "xdm:testEnabled": true,
     "xdm:messageClass": "continuous",
-    "xdm:templateID": 1000
+    "xdm:templateID": 1000,
+    "xdm:deliveryName": "DM200",
+    "xdm:deliveryLabel": "Birthday Wishes"
   },
   "https://ns.adobe.com/experience/campaign/marketingCampaign": {
-    "xdm:id": 100
+    "xdm:id": 100,
```
lrosenthol: ID isn't a number - it's a string.

extensions/adobe/experience/campaign/orchestration/experienceevent.schema.json
```diff
@@ -0,0 +1,90 @@
...
+  "definitions": {
+    "journey": {
+      "title": "Journey of an entity",
+      "type": "object",
+      "description": "Journey created for an entity by marketer.",
+      "properties": {
+        "@id": {
+          "title": "Journey unique identifier",
+          "type": "string",
+          "format": "uri",
```
lrosenthol: should this now be uri-reference?

extensions/adobe/experience/campaign/orchestration/experienceevent.schema.json
```diff
@@ -0,0 +1,90 @@
...
+    },
+    "journeyVersion": {
+      "title": "Journey version",
+      "type": "object",
+      "description": "Schema for describing the version of a journey, where version holds the state machine",
+      "properties": {
+        "@id": {
+          "title": "Journey version identifier",
+          "type": "string",
+          "format": "uri",
```
lrosenthol: should this now be uri-reference?

extensions/adobe/experience/campaign/orchestration/experienceevent.schema.json
```diff
@@ -0,0 +1,90 @@
...
+    },
+    "action": {
+      "title": "Action on step transition",
+      "type": "object",
+      "description": "Action taken during a step transition that happened for an entity in state machine",
+      "properties": {
+        "xdm:actionID": {
+          "title": "Action identifier",
+          "type": "string",
+          "format": "uri",
```
lrosenthol: should this now be uri-reference?

extensions/adobe/experience/campaign/experienceevent.example.email-send.json
ksradobe: Here,"id" attribute is an integer as per the XDM schema.

ksradobe: It is custom attribute xdm:id which is of type integer but not "@id" whose type is string

fmeschbe:
We should be discussing the type of the `xdm:id` properties.

This PR currently defines them to be of type `integer`. This works very well with the current implementation of ACS which is very SQL centric. As such, this choice of type IMHO exposes this implementation detail into the exchange and data format.

As we move forward and ACS is being rewritten and redesigned around new and more modern design and architecture principles, I am pretty sure, those identifiers won't remain numbers but will become strings if not even URIs.

As such, I propose the `xdm:id` properties to be defined of type `string`.

This has the added benefit that this would align with the definition of properties named `xdm:id` elsewhere, for example the identity schema.

extensions/adobe/experience/campaign/orchestration/experienceevent.schema.json
```diff
@@ -0,0 +1,90 @@
...
+          "description": "Unique identifier denoting the associated action."
+        },
+        "xdm:actionType": {
+          "title": "Action Type",
+          "type": "string",
+          "description": "The type of action to be performed.",
+          "meta:enum": {
```
fmeschbe: I am a bit at unease with the labels of this enum:

* They seem to be inconsistently named, for example _participe-substantive_ as in `scheduled_notification` or `parametrized_action` vs. _imperative-substantive_ as in `send_journey_notification` or just typed name as in `acs_writer`
* The `acs_*` labels seem very much tied into the ACS product

I think the labels should be consistently structured and not be tied to a current product name (which may change outside of our control and which limits the general applicability of this data structure.

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -191,6 +191,16 @@
...
+            },
+            "xdm:deliveryLabel": {
+              "title": "Delivery Label",
+              "type": "string",
+              "description" : "Human friendly name of the campaign activity which is originating this message."
+            },
+            "xdm:deliveryName": {
+              "title": "Delivery Internal name",
+              "type": "string",
+              "description" : "Friendly identifier of the campaign activity which is originating this message."
```
fmeschbe: I think the topic is that both the `xdm:deliveryLabel` and `xdm:deliveryName` are intended to be _human readable_. I think they should be consistently described as such.

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -191,6 +191,16 @@
               "type": "integer",
               "description":
                 "The delivery template's ID used to initialize this delivery.\n\nThe type the template used in Adobe Campaign can be identified using the `messageClass`.\n\n* `messageClass` = `oneTime`: the template is standard delivery template.\n* `messageClass` = `continuous`: the template is a recurring delivery.\n* `messageClass` = `transactional`: the template is a transactionnal message template.\n"
+            },
+            "xdm:deliveryLabel": {
+              "title": "Delivery Label",
+              "type": "string",
+              "description" : "Human friendly name of the campaign activity which is originating this message."
+            },
+            "xdm:deliveryName": {
```
fmeschbe: I still think the term _delivery_ is not needed here and in the  `xdm:deliveryLabel` since they are part of the _delivery_ schema/object.

Otherwise you could also make the case to call the ID field `xdm:deliveryId`.

extensions/adobe/experience/campaign/experienceevent.schema.json
```diff
@@ -202,6 +212,11 @@
               "title": "Campaign ID",
               "type": "integer",
               "description": "Identifier of the marketing campaign to which activity originating this message belongs to."
+            },
+            "xdm:campaignName": {
```
fmeschbe: Same discussion here for `xdm:campaignName` as part of the _campaign_ schema/object

extensions/adobe/experience/campaign/experienceevent.schema.json
ksradobe: change from "name", "label" to deliveryName and "deliveryLabel" is requested in one of the review comments. 
https://github.com/adobe/xdm/pull/464#discussion_r207927635

ksradobe:
@fmeschbe This PR itself is not introducing delivery and campaign "xdm:id" attribute. They we already merged quite some time back. IMHO we should tackle this change as part of new PR under new issue. If you required, I will create issue and assign it to the concern person

extensions/adobe/experience/campaign/experienceevent.schema.json
ksradobe: https://github.com/adobe/xdm/pull/472#discussion_r213328077

extensions/adobe/experience/campaign/experienceevent.schema.json
ksradobe: done

extensions/adobe/experience/campaign/orchestration/experienceevent.schema.json
ksradobe: 1)This actionType is already introduced XDM reporting event and now I am referring the same in campaign experience event. If inconsistent names needs to be handled here then it would be a breaking change. Should I create a separate issue and assign it to the concern person?
2) I agree with your comment in core XDM schema perspective. But generally xdm extension schemas are the only way to introduce product specific attributes. In that case, it should not be a problem use such names.

fmeschbe: @ksradobe wrote:

> This actionType is already introduced XDM reporting event and now I am referring the same in campaign experience event. If inconsistent names needs to be handled here then it would be a breaking change. 

This is a good call-out. Yet what is evolutionary state of this ? is it draft or stabilizing or finalized ?

ksradobe: It is in stabilizing state.

fmeschbe: Yes, I like that much better.

How about

- renaming `campaign_rest_call` to `campaign_request` and
- renaming `campaign_messageService_call` to `messageService_request` ?

ksradobe: @fmeschbe pushed my changes.

ksradobe:
@fmeschbe created a separate PR to track this independent improvement https://github.com/adobe/xdm/issues/499.

extensions/adobe/experience/campaign/orchestration/experienceevent.schema.json
ksradobe: @lrosenthol  I changed @id's format from uri to uri-reference [observed similar usages in the master branch]

kstreeter:
This is now in.  Please note that on master I added back the old `xdm:type` field and marked it as deprecated.  That allows this PR to go in without a breaking change that will impact current projects.

PR #479, opened 2018-08-20
Fix Segment Identity title conflict 

cfraser:
* updated title so it doesn't conflict with the title of the Identity schema


cfraser:
It would be great if we could get this in 9.9.5. Thanks. :)

Issue #469, opened 2018-08-15
Combine individual Birthday fields to a single field of format: 'date-time'


shieldz:
## What are the schemas that are affected by the issue
Profile
## What are examples of products that are impacted by the issue
Today we deal with most dates by creating a field where the format is "date-time." Examples include:
Anniversary
Subscription date
Purchase date
Renewal date
Membership date

But for birthday, we have three separate fields, all of format: 'integer'
* birthDay
* birthMonth
* birthYear

Request these be deprecated in favor of a single 'date-time' field for consistency and to reduce required workarounds in downstream services (ex. special casing integer fields to support additional date-related functions.

lrosenthol:
Huge +1 to fixing this!

kstreeter:
The birth date fields were represented this was in order to handle cases where a complete date is not known or desired.  For example, you may only know an age (i.e. birthYear) or you may only know the day/month but not the year (birthday, without knowing how old a person is).  Any change here needs to support these other cases.

Also, procedurally this schema is "stabilizing" so we should deprecate, rather than remove, any existing fields, unless we are prepared to treat this as a critical breaking change.

Please note I am not saying that fixing/changing this is a bad idea, just that it would have to be done with care!

shieldz:
Makes sense, Kevin.  The challenge here is that the current classification of these fields as 'integer' makes it difficult to apply date functions to these without building field-specific special cases, which we want to avoid.
Can we perhaps agree on a method for handling dates in which portions of that date (day, month, or year) are not known? That way we can support the cases you mentioned and maintain consistency as well.
We should also apply whatever decision we make here to other date fields, such as anniversary, where year may not be known or needed.

shieldz:
Correction, the ISO standard already dictates how to handle null missing fields.  @kstreeter this means cases with incomplete dates should be manageable provided the incoming data conforms to the standard.  Adding it below for reference:


Writing Omitted Components
An omitted component can be represented by a hyphen ( - ) or an x in the extended datetime form yyyy-mm-ddThh:mm:ss and in the extended duration form Pyyyy-mm-ddThh:mm:ss.

Omitted components in the durations form PnYnMnDTnHnMnS are dropped, they do not contain a hyphen or x. For example, P2mT4H.

The following formats write omitted components that use the hyphen and the x:

Format    Datetime Form    Duration Form    Examples
$N8601H    yyyy-mm-ddThh:mm:ss    PnYnMnDTnHnMnS    --09-15T15:-:53
[...]

PR #481, opened 2018-08-20
Add birthDate field to align on the standard date format

cfraser:
* add birthDate field as described in issue #469

### DIFF
--- a/schemas/context/person.schema.json
+++ b/schemas/context/person.schema.json
@@ -21,6 +21,12 @@
           "$ref": "https://ns.adobe.com/xdm/context/person-name",
           "description": "The person's full name"
         },
+        "xdm:birthDate": {
+          "title": "Birth Date",
+          "type": "string",
+	  "format": "date",
+          "description": "The date a person was born."
+	},
         "xdm:birthDay": {
           "title": "Birth day",
           "type": "integer",


schemas/context/person.schema.json
```diff
@@ -21,6 +21,12 @@
           "$ref": "https://ns.adobe.com/xdm/context/person-name",
           "description": "The person's full name"
         },
+        "xdm:birthDate": {
+          "title": "Birth Date",
+          "type": "string",
+	  "format": "date",
```
fmeschbe: As per the discussion, we don't want to have this to be `date` since the date might be incomplete. We discussed using a regular expression to match a date, so that we can be sure it is properly formatted.

Something along these lines:

```
"pattern": "([0-9X]{4}|XXXX)-([0-9]{2}|XX)-([0-9]{2}|XX)"
```

kstreeter:
Hey @cfraser (and @fmeschbe).  My suggestion is that we have three distinct fields which capture the specific cases we have for birthday.  I feel like this will make the interpretation of each field simpler:

  * xdm:birthDate - the full date, using the "date" format
  * xdm:birthDayAndMonth- the regex "XXXX-MM-DD" as @fmeschbe suggested
  * xdm:birthYear - the birth year, which can be used to compute age

I am a bit concerned about allowing partial dates in the same field as you would expect a full date.  This puts a lot of burden on readers to handle partials, and won't allow standard tooling to expose full dates as standard data types (like the Java Date class).

fmeschbe:
@kstreeter The main concern is indeed how to handle partial dates.  The standard Java classes do not seem to be ready to support this case properly. So an unknown day could default to first, while an unknown month would become January, and an unknown year .... (I don't know).

But that is not of a concern for XDM as a transfer format, is it ?

schemas/context/person.schema.json
HyveMynd: I believe that pattern is an acceptable ISO format, is it not? If so, the date format is enough.

fmeschbe: It is @HyveMynd. The problem is how to deal with incomplete dates. For example, when we don't know the year.

cfraser:
Closing in lieu of #498.

Issue #418, opened 2018-06-22
Change ID format from "uri" to "uri-reference"


kstreeter:
As we have continued to build tooling around XDM, we have encountered some variation in how strict tools are about validating or generating code for properties with a format of "uri".  Some tools allow for both relative URI references as well as absolute ones, while others only accept complete URIs in absolute form.

It appears this variation stems from changes made between JSON Schema draft 4 and draft 6.  In draft 4, the specification was not clear about whether format "uri" was intended to be absolute or not.  Tools went different direction on this, and the issue was addressed in draft 6.  Unfortunately, they chose a method that was not really backwards compatible, designating format "uri" as strictly an absolute URI, and adding a "uri-reference" to encompass the more expansive form.

A particular place where we use format "uri" is for describing the JSON-LD construct `@id`.  In JSON-LD, `@id` is allowed to be a relative reference.  So our use of format "uri" (under draft 6, which we have adopted in XDM) has overly constrained that definition.

We have a number of tools and components that assume it is acceptable to use the relative URI form in `@id` fields, based on the JSON-LD definition.  We are encountering issues in some toolchains that fully honor the definition of format "uri", and reject data that uses the relative form.

Since our use of `@id` is intended to align with its definition in JSON-LD, I would like to propose that we fix our mapping for `@id` to use format "uri-reference" rather than format "uri".  This will fix the tool chain issues, and bring us closer in sync with JSON-LD.

## What are the schemas that are affected by the issue

All

## What are examples of products that are impacted by the issue

All

kstreeter:
@lrosenthol @chrisdegroot @ogoldman please let me know what you think of this proposal.  

This should not be a breaking change from a schema perspective (we are relaxing constraints), but it will impact tooling that generates code (POJOs, etc) from schema definitions.  If we adopt this, I suggest we migrate definitions on a case-by-case basis after considering impact.

cdegroot-adobe:
I agree that the JSON-LD specification is going in a direction that forces us to make this sort of decision.

My concern is that there are three states, and there are only two options, and neither of them are what we want.

`uri-reference` -> spec says stuff about it being local to document base 
`uri-absolute` -> what json-ld really seems to say it wanted to go to, but alas did not.
`uri` -> should conform to `uri-reference` OR `uri-absolute`

Is short I think unless we are ok saying a `uri-reference` can be a absolute or relative uri we may not have a clear option.

cdegroot-adobe:
[...]

PR #482, opened 2018-08-21
Fix @id's format from "uri" to "uri-reference" if exists

jwen-adobe:
This PR links to the issue #418.

@kstreeter Fyi.


kstreeter:
Hey @jwen-adobe thanks for posting this.  It looks like maybe you prettified some of the files and pulled in some non-URI related changes.  While we probably do want to fix the formatting of those files at some point, the extra changes will make it harder to validate this change.  Can we roll it back and commit just the files with URI format changes?  Thanks!

jwen-adobe:
@kstreeter I know what u saying, I did not find a way to do it in my script as it has to save the file by using JSON.stringigy which will change the format, unless I have to do it manually...

PR #488, opened 2018-08-22
Change @id's format from "uri" to "uri-reference" if exists

jwen-adobe:
This PR links to the issue #418 

@kstreeter This PR contains only files with @id's format changed from "uri" to "uri-reference".


lrosenthol:
I am fine with this change...

Issue #489, opened 2018-08-23
Search annotations in XDM


nandaja:
**Problem / Premise**

In the current systems in Adobe Search, there is no standardized way to indicate in events / otherwise as to what fields need to be indexed and made searchable, which ones need to be made available in the search index for querying etc. There is no easy way for clients to discover the nuances of the data sets the search APIs will expose in the responses to search queries.

**Need**
Given the above premise, there is a need to standardize the interfaces across layers with special reference to the data interfaces and specifications around XDM. The proposal covers the approach to standardize annotations in XDM to be able to easily define which fields for a particular object ( as represented by a XDM schema) are to be indexed. Another aspect is to be able to also move the definition of the search API response data aspects including which fields are filter-able, which ones fetch-able, which ones are sortable and which ones are searchable as well to XDM. Once implemented, the XDM spec will be the source of truth for determination of the characteristics of behavior to determine what data sets will be indexed. The XDM spec will also serve as the source of truth to determine the response characteristics for data sets as exposed through the search APIs. 

**Approach**
We will extend the use of meta descriptors on XDM to define search annotations.
This is currently possible in two different ways : one to define them within the schema ( refer https://github.com/adobe/xdm/blob/master/docs/descriptors.md#embedding-schema-descriptors-in-a-schema ) , and the other being defining a standalone schema descriptor ( refer https://github.com/adobe/xdm/blob/master/docs/descriptors.md#example-of-defining-a-new-schema-descriptor )

The proposal is to use a combination of these two ways as enumerated above to define search annotations.

Search annotations proposed are listed below : 
* **search:indexable**: Boolean indicating if the property is  searchable or query-able
* **search:stored**: Boolean indicating if the property's raw values are stored in the index. If this value is false and search:indexable is true that means the field can be used to search on but cannot be used to fetch its value in the result set. This is there as an optimization to keep index size small.
* **search:filterable**: Boolean indicating if this property can be used as a filter in search queries
* **search:fetchable**: Boolean indicating if the property is returned by default in search query results. If search:stored = true that automatically makes an field to be fetched in search response. This field makes it available  by default in search results.
* **search:sort**: Includes properties  : Boolean indicates if this property can be used as a pivot for sorting search results by and Boolean to indicate case sensitiveness for sort
* **search:analyzed**: Include properties : Boolean indicating if the property is normalized using an analyzer, also supports meta to specify the analyzer to be used, and list of enum literals to define common analyzer types
[...]

PR #490, opened 2018-08-23
Defining search descriptors for search annotations : Issue #489

nandaja:
PR detailing changes proposed in Issue #489


nandaja:
@kstreeter @jaganperi @lrosenthol

kstreeter:
This looks pretty good @nandaja.  One thing that I think is missing is to add the "search" namespace and prefix to https://github.com/adobe/xdm/blob/master/schemas/common/context.jsonld

nandaja:
@kstreeter  Thanks for reviewing the PR. 
Sorry if I am missing something here . The addition of search namespace is part of the commit already. 

"search" : "https://ns.adobe.com/search" is added to schemas/common/context.jsonld. Is there anything else that needs to be done for namespace addition ?

jaganperi:
Looks good, @nandaja -- Is there a way to call out the defaults that would apply if the search descriptor is not specified/partially specified for a property?

nandaja:
Thanks for reviewing @jaganperi . I can call out the defaults as part of the documentation , will make these changes and add to this PR.

schemas/common/descriptors/search/asset.example.1.json
```diff
@@ -0,0 +1,448 @@
+{
```
lrosenthol: Is this an attempt to revise the asset schema with the search extensions?  Be aware that asset is undergoing change right now, so this will also probably change...

schemas/common/descriptors/search/searchdescriptor.description.md
```diff
@@ -0,0 +1,32 @@
...
+
+* `search:filterable`: Boolean indicating if this property can be used as a filter in search queries
+
+* `search:fetchable`: Boolean indicating if the property is returned by default in search query results
+
+* `search:stored`: Boolean indicating if the property's raw values are stored in the index, if false, the field can be used as a search pivot but its values cannot be fetched as part of the search results
+
+* `search:sort`: Includes properties  : Boolean indicates if this property can be used as a pivot for sorting search results by and Boolean to indicate case sensitiveness for sort
+
+* `search:analyzed`: Include properties : Boolean indicating if the property is analyzed using an analyzer,  and a list of enum literal values to specify the analyzers to be used
```
lrosenthol: search:analyzer

schemas/common/descriptors/search/searchdescriptor.schema.json
```diff
@@ -0,0 +1,92 @@
...
+    "https://ns.adobe.com/xdm/common/desciptors/schemadescriptor#/definitions/descriptor"
+  ],
+  "meta:abstract": false,
+  "type": "object",
+  "description": "A descriptor that describes search annotations for specific properties of the schemas.",
+  "definitions": {
+    "searchdescriptor": {
+      "properties": {
+        "search:indexable": {
+          "title": "isIndexable",
```
lrosenthol: missing a space

schemas/common/descriptors/search/searchdescriptor.schema.json
```diff
@@ -0,0 +1,92 @@
...
+          "type": "boolean",
+          "description": " When present, Boolean indicating if the property is indexable / searchable"
```
lrosenthol: Add the default value (for all entries)

schemas/common/descriptors/search/searchdescriptor.schema.json
```diff
@@ -0,0 +1,92 @@
...
+        },
+        "search:fetchable": {
+          "title": "is Fetchable",
+          "type": "boolean",
+          "description": "When present, Boolean indicating if the property is returned by default in search query results"
+        },
+        "search:filterable": {
+          "title": "is Filterbale",
```
lrosenthol: spelling error

schemas/common/descriptors/search/searchdescriptor.schema.json
```diff
@@ -0,0 +1,92 @@
...
+        },
+        "search:stored": {
+          "title": "is stored",
+          "type": "boolean",
+          "description": "When present, Boolean indicating if the property's raw values are stored in the index"
+        },
+        "search:sort": {
+          "properties": {
+            "search:sortable": {
+              "title": "isSortable",
```
lrosenthol: missing space

schemas/common/descriptors/search/searchdescriptor.schema.json
```diff
@@ -0,0 +1,92 @@
...
+              "title": "Analyzer to be applied",
+              "type": "array",
+              "items": {
+                "type": "string",
+                "meta:enum": {
+                  "standard": "standard",
+                  "keyword": "keyword",
+                  "simple": "simple"
+                },
+                "description": "When present, value of enum literals in array indicates the type of analyzers to be applied"
```
lrosenthol: is this a closed list?

mburbidgadobe:
@nandaja I think it would be great to give an example of how a search descriptor is used to annotate a property.

schemas/common/descriptors/search/asset.example.1.json
nandaja: This is an example - I can amend it later if required. We will need to later revise actual asset schema with the search extensions .That could be done later once the change underway is complete.

nandaja:
@mburbidgadobe  This is an example of annotating properties on the asset XDM schema : https://github.com/adobe/xdm/pull/490/files/921877981f42ac85ad1d25bc2b1e11b8947a51bd#diff-f3f10be67234909f7ea972ce6c96fc10. Please let me know if I need to add more examples and I can do the needful. Thanks !

schemas/common/descriptors/search/searchdescriptor.description.md
nandaja: Updated.

schemas/common/descriptors/search/searchdescriptor.schema.json
nandaja: Fixed, thanks !

schemas/common/descriptors/search/searchdescriptor.schema.json
nandaja: Updated with default values for all properties.

schemas/common/descriptors/search/searchdescriptor.schema.json
nandaja: Yes, this is a closed list - we may enhance it later to add others. If clients need special analyzers, they will have to work with us to enable those.

schemas/common/descriptors/search/searchdescriptor.schema.json
nandaja: Updaetd.

schemas/common/descriptors/search/searchdescriptor.schema.json
nandaja: Updated.

nandaja:
@lrosenthol  All changes based on the review comments have been incorporated.

lrosenthol:
@nandaja it looks like your merging the branches has created a massive PR with all sorts of changes to lots of other parts of the tree.  Can you either "undo" that *or* create a new branch from master, reapply just your changes, and then try a new PR?
Then let @ogoldman and myself know so we can do the final review and get this merged!

cfraser:
@nandaja I think your best bet is going to be to rebase instead of merging. That should eliminate the merged commits.

nandaja:
Thanks @lrosenthol  and @cfraser  . I will create a new PR and submit today.

nandaja:
Fork is up to date with original repo with changes relevant to this issue reapplied. Changes now include only the relevant ones for the addition of search annotations. 

@lrosenthol  @ogoldman  Please help with the review and merge. Thanks !

nandaja:
Thanks @lrosenthol  !

schemas/common/context.jsonld
```diff
@@ -28,6 +28,7 @@
     "activitystreams": "https://www.w3.org/ns/activitystreams",
     "schema": "http://schema.org",
     "repo": "http://ns.adobe.com/adobecloud/core/1.0/",
-    "iptc4xmpExt": "http://iptc.org/std/Iptc4xmpExt/1.1/"
+    "iptc4xmpExt": "http://iptc.org/std/Iptc4xmpExt/1.1/",
+    "search" : "https://ns.adobe.com/search"
```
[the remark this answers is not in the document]
cdegroot-adobe: There is no reason to create a new namespace, and it is to be avoided. These are additions to the Adobe model which all go under `xdm`. The only cases we use other namespaces is because we want to include third-party or preexisting historical standards.

schemas/common/descriptors/search/asset.example.1.json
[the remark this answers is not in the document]
cdegroot-adobe: Ah, very smart, that totally went over my head...

schemas/common/context.jsonld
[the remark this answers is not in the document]
lrosenthol: Actually, @cdegroot-adobe , @nandaja is absolutely doing the right thing (at my request).  We *do* want multiple namespaces to properly segment things.

schemas/common/descriptors/search/searchdescriptor.schema.json
```diff
@@ -0,0 +1,100 @@
...
+  "type": "object",
+  "description": "A descriptor that describes search annotations for specific properties of the schemas.",
+  "definitions": {
+    "searchdescriptor": {
+      "properties": {
+        "search:indexable": {
+          "title": "is Indexable",
+          "type": "boolean",
+          "description": " When present, Boolean indicating if the property is indexable / searchable",
+          "default": false
```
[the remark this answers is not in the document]
lrosenthol: Correct - that is indeed the default since most properties will not be indexed.

schemas/common/descriptors/search/searchdescriptor.schema.json
```diff
@@ -0,0 +1,100 @@
...
+          "description": "When present, Boolean indicates if this property can be used as  a filter in search queries",
+          "default": false
+        },
+        "search:stored": {
+          "title": "is stored",
+          "type": "boolean",
+          "description": "When present, Boolean indicating if the property's raw values are stored in the index",
+          "default": false
+        },
+        "search:sort": {
```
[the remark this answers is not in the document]
lrosenthol: Because the property needs to tell the Search system/index whether it should allow this field to be sorted or not.

schemas/common/descriptors/search/searchdescriptor.example.1.json
```diff
@@ -0,0 +1,19 @@
+{
+  "@id": "https://ns.adobe.com/xdm/common/descriptors/searchdescriptor",
+  "@type": "xdm:searchdescriptor",
+  "search:indexable": true,
+  "search:fetchable": true,
+  "search:filterable": true,
+  "search:stored": false,
```
mpetersen: Maybe I'm getting this example wrong, but fetchable: true and stored: false are contradicting, right? What is the expected result?

schemas/common/context.jsonld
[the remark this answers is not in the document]
cdegroot-adobe: @lrosenthol - very interesting, this is interesting to know, I think we will want to consider several aspects of this for the XC area too, especially around identity.

schemas/common/descriptors/search/searchdescriptor.schema.json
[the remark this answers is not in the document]
cdegroot-adobe: @lrosenthol - How am I to think of this when the property `search:indexable` is not present on say a Profile record? Would an SDK be required to return `false` or `null` or what?

I driving at there being an unintended consequence of setting the absence of this property impacting other schemas in XDM. Maybe we need a metadata hint on the schemas that expect to be constrained by the use of these search controls.

schemas/common/descriptors/search/searchdescriptor.schema.json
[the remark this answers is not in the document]
cdegroot-adobe: Is there a spec on search I am failing to be aware of? In XC a good case we would have is the Product schema that supports product catalogs. Do you expect XC would have to implement this `search` implementation for how products are searched on ecommerce sites. I assume yes, in which case we need the commerce teams to agree to the "non indexable by default" theme here. It still seems unworkable as it stands to me.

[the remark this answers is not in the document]
lrosenthol: @cdegroot-adobe interesting question. Right now, these search properties are being introduced for use by USS (Adobe's Universal Search System) which is used by various properties in DMe (eg. CC, DC, Stock, Lr).  However, philosophically, I would say that any system that is consuming XDMs and providing a search facility on top of them should use this info as well.  I don't think it's mandatory, just recommended.

schemas/common/descriptors/search/searchdescriptor.description.md
```diff
@@ -0,0 +1,32 @@
+A descriptor that describes the search related annotations on the properties of a schema.  
+
+The following properties are defined as part of this descriptor: 
+
+* `search:indexable`: Boolean indicating if the property is indexable / searchable
+
+* `search:filterable`: Boolean indicating if this property can be used as a filter in search queries
+
+* `search:fetchable`: Boolean indicating if the property is returned by default in search query results
```
cdegroot-adobe: Can you give an example of this `search:fetchable`, it would help me greatly. I do not find it obvious. 

PS: "fetchable" is showing as incorrect english in the chrome spell checker. It suggests "fetch-able". It does not have issues with the other -ables

schemas/common/descriptors/search/searchdescriptor.schema.json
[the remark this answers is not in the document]
cdegroot-adobe: @lrosenthol - I think it is being made mandatory by the "not searchable by default" nature it is defining for all searches on xdm schemas from my current reading.

Are you saying this is just a 'hint' not a permission? So different to robot.txt.

How expansive do you think "search" is? I take a classic search UX as given. But what about an agent searching in the background using ML etc and suggesting unrequested ideas. "simular images...", "Simular products..." etc.

schemas/common/descriptors/search/searchdescriptor.schema.json
```diff
@@ -0,0 +1,100 @@
...
+              "description": " When present, Boolean indicating if the property is analyzed ",
+              "default": false
+            },
+            "search:analyzer": {
+              "title": "Analyzer to be applied",
+              "type": "array",
+              "items": {
+                "type": "string",
+                "meta:enum": {
+                  "standard": "standard",
```
cdegroot-adobe: We typically uppercase the first letter of the descriptions in `meta:enum`

e.g "standard": "standard", -> "standard": "Standard",

schemas/common/descriptors/search/searchdescriptor.schema.json
```diff
@@ -0,0 +1,100 @@
...
+        },
+        "search:analyzer": {
+          "properties": {
+            "search:analyzed": {
+              "title": "isAnalyzed",
+              "type": "boolean",
+              "description": " When present, Boolean indicating if the property is analyzed ",
+              "default": false
+            },
+            "search:analyzer": {
```
cdegroot-adobe: In some search cases there are different analyzers for 'search' vs 'indexing', are you distinguishing between them here and avoiding issues is we are to evolve to supporting both in future. 

Refer:
https://docs.microsoft.com/en-us/azure/search/search-analyzers
```        {
           "name":"text",
           "type":"Edm.String",
           "searchable":true,
           "indexAnalyzer":"whitespace",
           "searchAnalyzer":"simple"
        },```

File: schemas/common/descriptors/search/searchdescriptor.schema.json [hunk 15]
```diff
@@ -0,0 +1,100 @@
...
+              "title": "is Case Insensitive",
+              "type": "boolean",
+              "description": "When present, Boolean indicating if the sort on the property is case insensitive",
+              "default": false
+            }
+          }
+        },
+        "search:analyzer": {
+          "properties": {
+            "search:analyzed": {
```
cdegroot-adobe: Seems superfluous, this boolean can only be true if a `search:analyzer` is defined. So it maybe duplicative.

schemas/common/descriptors/search/searchdescriptor.schema.json
```diff
@@ -0,0 +1,100 @@
...
+              "type": "boolean",
+              "description": " When present, Boolean indicating if the property is analyzed ",
+              "default": false
+            },
+            "search:analyzer": {
+              "title": "Analyzer to be applied",
+              "type": "array",
+              "items": {
+                "type": "string",
+                "meta:enum": {
```
cdegroot-adobe: Is there industry consistency with the definition of a "simple|standard|keyword" analyzer? If not I would like to see how we define them here.

PR #498, opened 2018-08-28
Issue 469 - Create distinct fields for different cases of birth date

kstreeter:
This PR is to address #469.  Add fields to address each case for knowledge of a person's birth date:

* the full date
* the day and month only (no year)
* the year only

The existing fields for "xdm:birthDay" and "xdm:birthMonth" will be deprecated.

Note this is intended to supercede PR #481.  Since that PR is not on a branch in this repo I could not just update it directly.


kstreeter:
@lrosenthol, @fmeschbe take a look at this and let me know if you have any comments.  I spoke with Aaron (who filed #469) this approach seems to address the issues raised.

schemas/context/person.schema.json
```diff
@@ -1,74 +1,91 @@
...
+                    "title": "Birth Date",
+                    "type": "string",
+                    "pattern": "[0-1][0-9]-[0-9][0-9]",
+                    "description": "The day and month a person was born, in the format MM-DD. This field should be used when the day and month of a person's birth is known, but not the year."
+                },
+                "xdm:birthYear": {
+                    "title": "Birth year",
+                    "type": "integer",
+                    "description": "The year a person was born including the century (yyyy, e.g 1983).  This field should be used when only the person's age is known, not the full birth date.",
+                    "minimum": 1,
```
fmeschbe: Is this minimum for a birth year reasonable ? I mean, if we set limits, shouldn't those limits be meaningful ?

In other words, is a birth year of, say, 67, reasonable for our XDM use cases ? And why would -100 (the assumed birth year of Julius Caesar) not be ?

kstreeter: Yes, this needs work.  This definition will let folks move forward, we should improve this under a new issue.

kstreeter:
@fmeschbe, I don't think it is accurate to call XDM an "exchange format", although that is one use.  XDM is also the metadata layer used to describe both the structure and semantics of data in our platform.  Systems we have built read this metadata and make assumptions about how to interpret and display data.  These systems frequently map from definitions made in XDM JSON Schema into the semantically equivalent form of their internal implementation.  So we need to take care not to introduce semantics into JSON Schema that these systems can't understand.

The new (and unpublished) semantics of partial dates in the latest ISO 8601 draft have not been translated into JSON Schema.  There is no support for it in any of the tooling or programming environments we use.  We need a solution that we can adopt now.

kstreeter:
Thank you all for the discussion on this issue, but as we have not gotten any viable counter-proposals we are going to go with this change for now so that teams can move forward.  These schemas are no yet locked down, so if folks want to propose improvements they will definitely be considered!

PR #504, opened 2018-09-04
Correct profile stitching

harleensahni:
There were two issues with the PR I created for profile stitching (#444 ) that weren't caught till now. 
Profile stitching on experience event should be an array of profile stitching entities, not a single profile stitch. The example was right before, but that was corrected to match the incorrect schema. 

Also, move to use Identity array in profile stitch. It was incorrectly put in segment membership instead due to some sloppy copy and pasting.


harleensahni:
@kstreeter, Can we get this merged in as a correction?

cmathis:
FYI - This creates a breaking change to the schema which affects any EE data that was previously ingested from Analytics or Campaign. Doing this will require everyone to throw away data and start over.

PR #513, opened 2018-09-25
PR 504 - Updates to make changes additive

kstreeter:
PR #504 was merged with several breaking changes.  This PR updates these changes to make them additive only.

### DIFF
--- a/schemas/context/profilestitch.schema.json
+++ b/schemas/context/profilestitch.schema.json
@@ -31,6 +31,12 @@
           },
           "minItems": 1,
           "description": "Array of Identities. Condensed, normalized encapsulation of all end user identifiers."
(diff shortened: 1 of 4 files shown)


schemas/context/profilestitch.schema.json
```diff
@@ -31,6 +31,12 @@
           },
           "minItems": 1,
           "description": "Array of Identities. Condensed, normalized encapsulation of all end user identifiers."
+        },
+        "xdm:endUserIDs": {
```
harleensahni: my big worry here is that when we have IdentityMap, we'll have three fields for representing identity here. I get that we'll deprecate the old ones, but just want to document it here on the review.

Issue #517, opened 2018-09-27
Support key/values map on ExperienceEvent


harleensahni:
## What are the schemas that are affected by the issue
ExperienceEvent
## What are examples of products that are impacted by the issue

Please add support for the newly being added map type to store values with a map on experience event. 

@cmathis, @kstreeter

ankushkd:
PR : https://github.com/adobe/xdm/pull/546

PR #523, opened 2018-09-27
Issue 517 - Unify representation of end-user identity between ExperienceEvent and Profile

kstreeter:
This PR addresses issue #517.  

Currently, EE and Profile have different representations of end user identities.  This has caused some confusion by implementors, and in addition the representation in EE cannot carry multiple identities of a single namespace, which is a problem for some use cases.

The new representation uses the newly introduced "map" type to model identities as a map from identity namespace ID to one or more instances of the `Identity` schema.  This makes it possible to carry multiple identities from a single namespace (which was possible with the array representation), while still doing direct lookups for the presence of identities of a specific space (which was not possible with the array representation).


schemas/context/experienceevent.example.1.json
```diff
@@ -5,26 +5,31 @@
...
-    "https://ns.adobe.com/experience/aaid": {
-      "xdm:id": "2394509340-30453470347",
-      "xdm:namespace": {
-        "xdm:code": "AVID"
+    ],
+    "https://data.adobe.io/entities/namespace/10": [
+      {
+        "xdm:id": "2394509340-30453470347",
+        "xdm:namespace": {
+          "xdm:code": "AVID"
```
harleensahni: Is the code redundant? I take it this part is optional and all of xdm:namespace could be considered optional?

harleensahni:
@kstreeter ProfileStitch Identities should also be updated: https://github.com/adobe/xdm/blob/master/schemas/context/profilestitch.schema.json#L26 to use IdentityMap for consistency.

schemas/context/experienceevent.example.1.json
biswas-adobe: If the intention of using an ID is to have a stable reference to the namespace and not have systems use the code in their implementations, then we should remove it. There's nothing stopping any implementation from assuming the code can be used for generating a fully qualified identity.

neerajgoel82: This structure allows for specifying different namespace codes mapped to the same namespace key. Is that OK? If not, xdm:namespace should be outside the id object.

kstreeter: Since the code can change, it makes sense to allow for the _current_ value to (optionally) be captured in the data.

rkpandey: If we want both id and code how we are planning to get id at the time of ETL. Are we saying ETL ecosystem with integrate with Adobe Service to fetch id?

tcuevas-adobe:
This breaks open source tools. Json key names need to compliant with open source tools. Using urls/uris in keys breaks these tools.

schemas/context/experienceevent.example.1.json
rkpandey: Also, if keys are going to be represented as "https://data.adobe.io/entities/namespace/4" how are we use this in json pointer notation. REF-> https://tools.ietf.org/html/rfc6901

kstreeter: @rkpandey the mechanism is that external callers use the integration code (as it is the user-facing "handle" for the namespace), and the system transforms to id for storage as appropriate.  The interface is the code, while the id is used internally.

There is no incompatibility between URIs and JSON Pointer.  JSON Pointer allows for escaping of reserved characters.

cdegroot-adobe:
@kstreeter - we also have another identities[] representation here https://github.com/adobe/xdm/blob/master/schemas/context/profilestitch.schema.json

What do you wan to do with this? considering we are moving away for the identities array?

schemas/context/profilestitch.example.1.json
```diff
@@ -6,17 +6,12 @@
...
-      "xdm:id": "https://data.adobe.io/entities/identity/92312748749128",
-      "additionalIDs": [
-        "https://data.adobe.io/entities/identity/62312748749321",
-        "https://data.adobe.io/entities/identity/49312748749132"
-      ],
-      "xdm:namespace": {
-        "xdm:code": "ECID"
-      }
-    }
+  "xdm:identityNap": {
```
cdegroot-adobe: Misspelling "identity*N*ap"

schemas/context/experienceevent.schema.json
```diff
@@ -38,9 +38,15 @@
           "description": "The time at which this interaction was received by a server."
         },
         "xdm:endUserIDs": {
+          "meta:status": "deprecated",
```
cluby-adobe: I thought the plan was to make this a breaking change and just remove it so we're clean.

kstreeter: we are going to do an initial drop of this change with both to help testing.  We will drop the older fields when we do the larger schema refactor.

Issue #516, opened 2018-09-26
Bug:  XDM requires that productViews.id length be > 20


Sgadamsetty04:
## What are the schemas that are affected by the issue
_xdm.context.experienceevent

## What are examples of products that are impacted by the issue

The constraint on the product ID seems like a bug. We need to remove this. 

Background: I have a ACME Retail Application where user views and purchase products. Application integrates with Platform and sends all experience event as well as Profile update using Streaming via  DCCS.  

Here is JSON message for Product View:

[{"header":{"imsOrgId":"340B40865B7BAD1F0A49403C@AdobeOrg","msgType":"xdmEntityCreate","datasetId":"5b7c58814fabe601e1680d17 ","createdAt":1536896306783,"xdmSchema":{"name":"_xdm.context.experienceevent","version":"0.9.9.3","namespace":""},"source":{"name":"Acme Retail","description":null}},"body": { "xdmMeta": {"xdmSchema":{"name":"_xdm.context.experienceevent","version":"0.9.9.3","namespace":""}} , "xdmEntity":{"web":{"webPageDetails":{"siteSection":"Computers & Tablets,Computer Accessories & Peripherals,Laptop Accessories,Laptop Chargers & Adapters,Computers,Electronics,Computer Accessories","name":"Lenovo - AC Adapter for Select Lenovo Yoga Laptops - Black"}},"endUserIDs":{"_experience":{"acid":{"namespace":{"code":"AcmeID"},"id":"4162807","primary":true}}},"_id":"1536896287398","commerce":{"productViews":{"id":"1008","value":36.74}},"timestamp":"2018-09-13T20:38:07-07:00"}

Now validation for this message is failing because XDM requires that productViews.id length should be > 20 . Is there any reason why this id should be greater than 20 character length? In my case, I am sending the productId  stored in database which is 4 characters long. 

               "productViews": {
                    "description": "View(s) of a product have occurred.",
                    "meta:xdmField": "xdm:productViews",
                    "type": "object",
                    "properties": {
                        "_type": {
                            "type": "string",
                            "format": "uri",
                            "description": "Links to the `Metric` that this measure refers to. The `@type` can be omitted in contexts where it is obvious what metric is being referred to, for instance in the `Metrics` object that is keyed with `Metric` URIs.",
                            "meta:xdmType": "string",
                            "meta:xdmField": "@type"
                        },
                        "id": {
                            "title": "Unique Identifier",
                            "type": "string",
                            "minLength": 20,
[...]

PR #524, opened 2018-09-27
Issue 516 - Removing minLength constraint on measure ids

kstreeter:
This addresses issue #516.  The `measure` schema has a constraint on it that requires the ID of the measure to be at least 20 chars.  This has caused some confusion in cases where a shorter identifier was available but could not be used due to the constraint.

I actually don't recall why the constraint was added in the first place.  @trieloff or @cdegroot-adobe do you recall?


cdegroot-adobe:
@jwen-adobe - Looks like you added this originally, do you remember a reason?

jwen-adobe:
@chrisdegroot After digging the change log, I do see I made the change specifically. Frankly, I do not recall any reasons for that though unless @cmathis remembered anything, or maybe someone suggested me do that.

Issue #536, opened 2018-10-16
Add domain and ISP to environment


stormybriggs1:
## What are the schemas that are affected by the issue
context/environment.schema.json

## What are examples of products that are impacted by the issue
Analytics and everything down stream.

PR #537, opened 2018-10-16
Add domain and ISP

stormybriggs1:
Issue: https://github.com/adobe/xdm/issues/536

Add fields for the domain of a service provider and the service provider themselves (as ISP).


lrosenthol:
I don't understand how you would even get that information let alone use it.  Is this really necessary?

stormybriggs1:
@lrosenthol We have been collecting this data for some time in Analytics. I don't know details of usage, but I imagine there are reports today using this data.

cdegroot-adobe:
@lrosenthol - this is currently in use today, there are various data services that provide it, likely based on ip address. 

@kstreeter - I approve this

Issue #538, opened 2018-10-16
Advertising break should not be required


stormybriggs1:
## What are the schemas that are affected by the issue
context/advertising-break.schema.json

## What are examples of products that are impacted by the issue
Analytics.

PR #539, opened 2018-10-16
Remove the requirement for offset in advertising break

stormybriggs1:
Issue: https://github.com/adobe/xdm/issues/538
When Analytics is creating an advertising break object, the only information that it has is the ID. Right now we do not have classification information and therefore cannot get the offset. We would still like to create the advertising break and not put in an inaccurate offset, but instead leave it blank.

### DIFF
--- a/schemas/context/advertising-break.schema.json
+++ b/schemas/context/advertising-break.schema.json
@@ -32,8 +32,7 @@
         }
       },
       "required": [
-        "@id",
-        "xdm:offset"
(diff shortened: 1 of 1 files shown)


cdegroot-adobe:
@gabebarcelos-adobe  can you please review and approve this?

cdegroot-adobe:
I do nto see an issue with this assuming AdCloud do not technically require it

PR #542, opened 2018-10-25
Fix #541. Reduce MutationEvent schema to fields required currently.

### LINKED ISSUE #541: Reduce MutationEvent schema to bare minimum
mutationevent, experienceevent

Campaign, UPS

@kstreeter can fill in the need for this change as I am not exactly aware of the technical limitations at UPS for which this change was requested.

This Issue aims to capture the need to remove extra fields from the mutation event schema and keep only those that are currently being used to produce events by Campaign.

The original idea of mutation event is to enable mutable events to explain a state change. Only the old values of the fields that have changed from the previous event in the new event will be sent as part of the mutation event. This would enable reconstruction of the previous event from the new event and negate the effects of the previous event. if required. Hence, originally, it was kept to be a `$ref` of the original experienceevent schema itself. 
However, this seems to be a problem since the experienceevent schema is huge and "mutation" is also derived from this schema. 

Concerns: This change may make the design non-flexible. Any new field that changes in a state transition would first need to be explicitely added to this schema and then be used.

Informed: @cmathis @cdegroot-adobe @kstreeter

saurabhere:
Fixes #541 

Need suggestions on how to `$ref` individual items into the schema. Currently these are made as copies from the original experienceEvent schema.

I don't suppose something like
```
"xdm:reason": {
    $ref: "https://ns.adobe.com/experience/campaign/message#/properties/xdm:reason" `
}
```
or
```
"allOf":[
{
    "$ref": "https://ns.adobe.com/experience/campaign/experienceevent#/definitions/https://ns.adobe.com/experience/campaign/message#properties/xdm:reason"
},
...
```
would work. 
Or will it?
How can I test by generating expanded schemas?
@trieloff @kstreeter  ?


extensions/adobe/experience/campaign/mutationevent.schema.json
```diff
@@ -13,68 +13,13 @@
   "definitions": {
     "experienceevent": {
```
cfraser: Should this be `mutationevent`?

saurabhere: should be! Fixing it.

cdegroot-adobe:
@kstreeter we will need this merged today as well as it resolves some issues that break semantics UPS requires.

Issue #543, opened 2018-10-25
Add xdm:localTimezoneOffset next to xdm:localTime under placecontext


cluby-adobe:
## What are the schemas that are affected by the issue
context/placecontext

## What are examples of products that are impacted by the issue
Adobe Analytics - Able to preserve the users local timezone

The reason to add the local timezone offset separate from the localTime is to preserve the original local timezone regardless of the serialization used for localTime.  If localTime is stored as a UTC timestamp long or as a Parquet timestamp field the original timezone that may be present in the ISO date-time string would be lost.

The intent is to make the new xdm:localTimezoneOffset field a signed int that will contain the timezone offset in minutes from UTC.

PR #544, opened 2018-10-25
[543] Added xdm:localTimezoneOffset to context/placecontext

cluby-adobe:
https://github.com/adobe/xdm/issues/543

@kstreeter & @cdegroot-adobe

### DIFF
--- a/schemas/context/placecontext.example.1.json
+++ b/schemas/context/placecontext.example.1.json
@@ -1,5 +1,6 @@
 {
   "xdm:localTime": "2001-07-04T12:08:56+01:00",
+  "xdm:localTimezoneOffset": -360,
   "xdm:geo": {
     "@id": "https://data.adobe.io/entities/geo/tokyo",
     "xdm:countryCode": "JP",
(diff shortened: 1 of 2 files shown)


kstreeter:
Hey @cluby-adobe this addition makes sense, but would it make sense to make the offset a string that has the same form of the offset defined in ISO8601?  This definition seems to be what they do in Javascript, but in Java they appear to support the string form.

cluby-adobe:
@cluby-adobe, The issue is we don't know the actual timezone.  We only have the current offset.

kstreeter:
Understood we would not have the actual timezone, just the offset.  ISO8601 has a notation for that (a "-" followed by the HH:MM of the offset).  That notation appears to be supported at least in the Java environment.

cluby-adobe:
@kstreeter, Do you know of anything that tries to use this format for timezone offset on it's own outside of a date-time string?  I want to make sure there is value in using this format to justify the cost of dealing with in at various layers.  If every layer is just going to have to convert it into minutes to use it why don't we just represent it as minutes.

Issue #545, opened 2018-10-30
Need Auxiliary Properties to be part of Experience Event Schema


ankushkd:
Audience Manager accepts auxiliary properties from other solutions and third party providers which are basically key value pairs and do not have fixed set of keys (within a provider or across providers). These properties can be best captured as a map with in an instance of XDM. This will help in providing more context and information to the current instance of XDM.
This need not be limited to Audience Manager requirement but generic requirement as a whole.

## What are the schemas that are affected by the issue
Experience Event (Others later)

## What are examples of products that are impacted by the issue
Adobe Audience Manager
Adobe Cloud Platform

PR #546, opened 2018-10-30
[545] New model for auxiliaryProperties added. Map of String and values

ankushkd:
Issue: https://github.com/adobe/xdm/issues/545


lrosenthol:
We already have a way to do this with XDM - please don't create your own.   @kstreeter can you help here?

ankushkd:
@kstreeter  @lrosenthol  please advise here. We want to be able to define and validate type of value in map.

lrosenthol:
@ankushkd My biggest problem with this approach is shown perfectly in your example - you have numbers represented as strings, which we would not want.  Data needs to be properly typed.

ankushkd:
@lrosenthol The example could be better here, will update. Important point is, it is meant for keys we do not know yet. And these are not always fixed as well. We can not model it on the fly and  even if we try to it, it would end up as huge horizontal schema. Requirement is to model this not currently known key value pairs which are coming though ingestion channels into a Map. Downstream systems and validation services will need to know the value type is a String array.

ankushkd:
Closing this pull request. Will open a new one with latest changes.

ankushkd:
Please refer to https://github.com/adobe/xdm/pull/556 for latest on this.

PR #552, opened 2018-11-06
 Added dcx JSON-LD prefix definition & dcx:type property to the meta schema

ananddaga14:
No Issue filed for this yet.

After we write DCX schemas, before we use one of those schemas we want to know what DCX format the schema applies to, hence the need to introduce the dcx:type property in the meta schema.


meta.schema.json
```diff
@@ -125,6 +125,12 @@
     "title": {
       "type": "string",
       "minLength": 3
+    },
+    "dcx:type": {
```
ogoldman: Suggest we call this "dcx:typeRestriction". It's not a media type in the same sense as dc:format, or the "type" format in the DCX manifest, but simply calling it "dcx:type" suggest that it is.

ananddaga14: Agree, I've made this change now.

kstreeter:
looks like you are good with this one @ogoldman?

PR #554, opened 2018-11-12
XDM composition refactor

cdegroot-adobe:
To better support the Composition semantics in the Experience Platform the Context and Experience Solution schemas are beings structured a little differently. 

This PR covers a number of updates across many schemas but does not change the behaviour of the end schemas other than providing a method of not including all the fields in every case.


schemas/context/identitymap.schema.json
```diff
@@ -0,0 +1,42 @@
...
+  "title": "Profile",
+  "type": "object",
+  "auditable": true,
+  "meta:extensible": true,
+  "meta:intendedToExtend": ["https://ns.adobe.com/xdm/context/profile", "https://ns.adobe.com/xdm/context/pexperienceevent"],
+  "description":
+    "End User identity assertions.",
+  "definitions": {
+    "identitymap": {
+      "properties": {
```
cluby-adobe: @cdegroot-adobe / @kstreeter , I thought this was supposed to use the new map type support.

extensions/adobe/experience/adcloud-experienceevent.schema.json
```diff
@@ -0,0 +1,67 @@
...
+  "title": "Adobe AdCloud ExperienceEvent Base",
+  "type": "object",
+  "description":
+    "This AdCloud ExperienceEvent base schema includes the core/standard ExperienceEvent as well as the Adobe AdCloud ExperienceEvent Extensions.",
+  "meta:extensible": true,
+  "meta:abstract": true,
+  "meta:extends": [
+    "https://ns.adobe.com/xdm/data/time-series",
+    "https://ns.adobe.com/xdm/context/identitymap",
+    "https://ns.adobe.com/xdm/context/experienceevent-advertising",
```
kstreeter: I know we want our tooling to unroll all of the extensions that a leaf schema implements, but here in the repo for authoring purposes I don't think we should duplicate all of these.  It will be very difficult to maintain by hand!

extensions/adobe/experience/adcloud-experienceevent.schema.json
```diff
@@ -0,0 +1,67 @@
...
+    {
+      "$ref": "https://ns.adobe.com/xdm/common/extensible#/definitions/@context"
+    },
+    {
+      "$ref": "https://ns.adobe.com/experience/experienceevent"
+    },
+    {
+      "$ref": "https://ns.adobe.com/experience/adcloud/experienceevent-all"
+    },
+    {
```
kstreeter: I think the XC "shared" stuff should get mixed in once, in the experience namespace shared schema?

extensions/adobe/experience/analytics/experienceevent-all.schema.json
```diff
@@ -5,12 +5,14 @@
     "you may not use this file except in compliance with the License. You may obtain a copy",
     "of the License at https://creativecommons.org/licenses/by/4.0/"
   ],
-  "$id": "https://ns.adobe.com/experience/analytics/experienceevent",
+  "$id": "https://ns.adobe.com/experience/analytics/experienceevent-all",
```
kstreeter: since this is the extension that wraps everything intended to be added to the final schema, should we call it "extension" versus "all"?

cdegroot-adobe: No, the experience/analytics-experienceevent.schema.json is the is the thing "that wraps everything intended to be added to the final schema". it is in-fact a base schema that Analytics will allOf to create the tenant level schema. e.g https://github.com/adobe/xdm/pull/554/files#diff-f452c02ba041146a1fb27a2f8be049bd 

The "-all" in a experience/{solution} is all the solution specific fields that are never shared.

schemas/context/experienceevent-advertising.schema.json
```diff
@@ -0,0 +1,36 @@
+{
+  "meta:license": [
+    "Copyright 2017 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/context/experienceevent-advertising",
```
cdegroot-adobe: @kstreeter - I know you do not like this naming convention I like it because it manifests the `"meta:intendedToExtend": ["https://ns.adobe.com/xdm/context/experienceevent"]` in a way that is easier to see. I'm a simple person... Do you have an alternative recommendation?

schemas/context/profile-person-details.schema.json
```diff
@@ -0,0 +1,36 @@
+{
+  "meta:license": [
+    "Copyright 2018 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/context/profile-person-details",
```
cdegroot-adobe: @kstreeter I am open to changing this to person as you mentioned when we talked.  Because we currently have a complex type of that name `"https://ns.adobe.com/xdm/context/person"` it would need to be person-extension. Is that a pattern you want really? Does it actually help?

PR #556, opened 2018-11-16
XDM#545 Introducing auxiliary properties for dynamic, infinite keys coming with an experience event

ankushkd:
Issue #545 

For historical conversation, please check #546

### DIFF
--- a/schemas/context/experienceevent.example.1.json
+++ b/schemas/context/experienceevent.example.1.json
@@ -165,5 +165,20 @@
         "xdm:status": "realized"
       }
     }
-  ]
+  ],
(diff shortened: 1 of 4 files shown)


ankushkd:
Will update package.json once PR is approved.

kstreeter:
Hi @ankushkd, I don't agree that this is something that should be included in the experience event schema.  This usage is actually one of the reasons folks pushed back on introducing the map concept at all, because it encouraged people to avoid modeling data entirely and just put things into an opaque bucket.
If we have existing application data that is in this form, and it is not possible to model it properly, it should be put in an application extension.

PR #577, opened 2019-01-16
Correcting xdm:eventType in Campaign Extension, it should not appear in the root of EE

cdegroot-adobe:
xdm:eventType included twice, once in the allOf and once as a ref in "https://ns.adobe.com/experience/campaign/orchestration/reportingevent".

I assume the latter is correct. The former is introducing the field 
into the root of context/experienceevent and I am removing that.


cdegroot-adobe:
@kstreeter , we will need this change in today to give UPS their eventType[]. Please review and merge.

Issue #579, opened 2019-01-16
Link Record data with Events related to that record.


cdegroot-adobe:
## What are the schemas that are affected by the issue
'context/profile'

## What are examples of products that are impacted by the issue
Unified Profile

There needs to be a way to link time-series based events with the record based entities that those events belong to. This needs to be a soft relationship because we cannot statically embed the context/experienceevent and context/profile schemas together because they are too large.

PR #580, opened 2019-01-16
WIP - Implement event list for context/record based schemas

cdegroot-adobe:
Please link to the issue #579 
Initial implementation proposal.


schemas/data/record.schema.json
```diff
@@ -21,6 +21,14 @@
           "type": "string",
           "format": "uri-reference",
           "description": "A unique identifier for the record."
+        },
+        "xdm:eventList": {
+          "title": "Event List",
+          "description": "List of time-series based events that relate to schemas based on record.",
+          "type": "array",
```
tcuevas-adobe: Given that this a field for time-series events, what is the expected order desc/asc?

schemas/data/record.schema.json
```diff
@@ -21,6 +21,14 @@
           "type": "string",
           "format": "uri-reference",
           "description": "A unique identifier for the record."
+        },
+        "xdm:eventList": {
```
tcuevas-adobe: can we be explicit and name this "timeSeriesEvents" rather than eventList?

schemas/data/record.schema.json
cdegroot-adobe: @tcuevas-adobe I don't have any particular opinion on the ordering, so please outline what you consider the better option.

I can see the following alternatives:
1. The schema defines one (ascending/descending).
2. There is a descriptor that defines this.
3. The query system takes this as a parameter when it constructs the profile record and the schema may have a property for `eventListOrder` to represent this.

I think the latter is the better option.

schemas/data/record.schema.json
cdegroot-adobe: @tcuevas-adobe I agree that the name can be improved. In practice the implementation will never be an actual `data/time-series` schema, but a union of a class (likely context/experienceevent(~99% of the time)) that has the behaviour `of data/time-series`

I wonder is there a better name, if not I will agree with `timeSeriesEvents`

kstreeter: I'm also good with timeSeriesEvents

Issue #589, opened 2019-01-24
add a new field stitch_id_chosen to Ad Cloud's extension


srikanthrks:
## What are the schemas that are affected by the issue
AdCloud's extension of ExperienceEvents
xdm/extensions/adobe/experience/adcloud/experienceevent-all.schema.json
## What are examples of products that are impacted by the issue
only AdCloud uses this for now

PR #591, opened 2019-01-25
Issue-589 - add a new field stitch_id_chosen to Ad Cloud's extension

skannan-adobe:
add a new field stitch_id_chosen to Ad Cloud's extension
https://github.com/adobe/xdm/issues/589


skannan-adobe:
@kstreeter @lrosenthol Can you please review this.

extensions/adobe/experience/adcloud/experienceevent-all.schema.json
```diff
@@ -35,6 +35,11 @@
           "title": "Ad Asset Details",
           "$ref": "https://ns.adobe.com/experience/adcloud/advertisement",
           "description": "Digital advertisement details"
+        },
+        "https://ns.adobe.com/experience/adcloud/stitchIdChosen": {
+          "title": "Stitch ID Chosen Type",
+          "type": "string",
+          "description": "ID from the Ad Servers through AdCloud STATS to track click-through conversion on browsers that clock third party cookies"
```
cdegroot-adobe: What does "clock third party cookies" mean?

extensions/adobe/experience/adcloud/experienceevent-all.schema.json
```diff
@@ -35,6 +35,11 @@
           "title": "Ad Asset Details",
           "$ref": "https://ns.adobe.com/experience/adcloud/advertisement",
           "description": "Digital advertisement details"
+        },
+        "https://ns.adobe.com/experience/adcloud/stitchIdChosen": {
```
cdegroot-adobe: Naming conventions require the following.
"https://ns.adobe.com/experience/adcloud/stitchIDType"
"title": "Stitch Identifier Type"
"description": "Identifier type from the Ad Servers collected through AdCloud STATS to track click-through conversion on browsers that ?clock? third party cookies."

extensions/adobe/experience/adcloud/experienceevent-all.schema.json
skannan-adobe: @cdegroot-adobe : ah, 🤦‍♂️ That's a typo. Should have been block. Will fix it in my next commit

extensions/adobe/experience/adcloud/experienceevent-all.schema.json
gabebarcelos-adobe: We should change this to stichId. The chosen is implied.

extensions/adobe/experience/adcloud/experienceevent-all.schema.json
```diff
@@ -35,6 +35,11 @@
...
+          "title": "Stitch ID Chosen Type",
```
gabebarcelos-adobe: Take "Type off the title"

lrosenthol:
Why is this one single value in the middle of a schema that points to all sorts of others?  Why not put stitchID in one of them?

skannan-adobe:
@gabebarcelos-adobe : based on @lrosenthol 's comment do you think `adcloud-experienceevent.schema.json` would be better place to add this property ?

kstreeter:
Hi @lrosenthol, this field doesn't appear to be related to the other sub-schemas ("campaign", etc.)  Looks like it is appropriate to keep it at the root?

skannan-adobe:
@lrosenthol : I concur with @kstreeter. The `stitchId`is specific for AdCloud experience and does not belong to any of the definitions it includes, `campaign`, `inventory`, `advertisement` or the root `experienceevent`

lrosenthol:
Ah, I see where this is going.  This is a uncommon way to create a new "experience event" just for AdCloud.  Is this really the right way?

skannan-adobe:
@lrosenthol : I am not sure I understand. My goal is not to create a `experienceevent`, but to include a new field (stitchId) to the existing schema, `extensions/adobe/experience/adcloud/experienceevent-all.schema.json`, which extends the `schemas/context/experienceevent.schema.json`

lrosenthol:
My point is that `experienceevent-all` is simply an extension that (as the name implies) copies all the others *and* now will add something else.  While valid, it just seems a strange approach.  Why do we need an `all` flavor?  What is it gaining us to start with?  Trying to understand the use here...

kstreeter:
Hi @lrosenthol, the "-all" notation is just being used to make the extension distinct from the "-shared" extension we are creating for some services (like target).  It's a little confusing here because AdCloud doesn't have both flavors, but for target we needed to carve out the extensions that also get pulled in from Analytics.  We have tried to keep the naming convention consistent in case this distinction becomes important for other applications.

skannan-adobe:
@kstreeter Thanks for addressing that. I am fairly new to XDM and I am trying  get understand this better. @lrosenthol : to add more color to why`experienceevent-all` was chosen to add this new field . The `stich_id_chosen` was recent addition (Jan-2019) to the `StatsEvent` which maps to the experience-event and going forward will be part of all the StatsEvent. Hence I thought he best place to include this was the existing schema instead of creating a new extension or adding it to the one of the definition this extension imports or the root.

skannan-adobe:
@lrosenthol : a gentle reminder. Let me know if you still have concern about this ?

extensions/adobe/experience/adcloud/experienceevent-all.schema.json
kstreeter: Hey @cdegroot-adobe, is this still a concern?

cdegroot-adobe:
@kstreeter - I am good with this one thanks

skannan-adobe:
thanks @kstreeter @cdegroot-adobe  :)

PR #593, opened 2019-01-30
 Changes to change profile segments to segmentMap

meetraviagg:
Issue #595


rkpandey:
cc @chrisdegroot

harleensahni:
One issue with modeling segment membership as a map versus a list is that the list gave us an avenue to support storing historic states of segment membership for a segment by allowing you to have multiple segment memberships for the same segment but with different timestamps.  This could let you see when the segment was entered, requalified, or even see previous values for payloads. Granted this method of relisting was verbose. If we'd want to introduce a similar concept here, the map should be a map of lists. Or we could modify the membership object itself to track this. 

The other comment I have is that whatever we do to model segment membership here, we may want to explore doing also for the segment membership that exists in Experience Event.

kstreeter:
@rkpandey, do you have a response to @harleensahni 's question? 

I am actually unclear on why we need this.  There is no detail in the referenced issue #595.

meetraviagg:
Hi @kstreeter,

I have updated the issue with more details. Please let me know if any more information is required.

cc @rkpandey

kstreeter:
@meetraviagg, are you okay with @harleensahni 's suggestion to make this a map of lists?  If no we will need to resolve the issues he raised before we can move forward with this.

rkpandey:
@harleensahni and @kstreeter Array of segment was never meant for storing historical segment information. We have used arrays in last PR because we didn't have map support at that time... 
Can we handle that requirement to track historical status using a different PR.

harleensahni:
@rkpandey works for me. We did talk about using arrays to store historical information a while ago (I think as a counter to adding each of  the historic states in each segment membership item itself), but I don't want to hold up this PR on this, or go for the suggestion of map of lists which is probably not natural for your use case. 

Here is a suggestion if we want to track historic membership going forward: introduce a new field called historicSegmentMap and let that be a map of arrays. We live this field as-is to track the current state. @kstreeter, does that work for you? I don't think it needs to be done as this PR.

kstreeter:
Okay sounds like we have a path forward.  @meetraviagg, to complete the PR the same change needs to be applied to Experience Event.  Since segmentmembershipitem will no longer be referenced, this will should be removed.

meetraviagg:
@kstreeter, actually from UPS point of view UPS does not use/recommend segment-membership at experience-event level, we actually segment profiles and not experience events. In our opinion segment membership should be removed from experience event schema. Since i am not sure, which team/solution is actually using segment membership at experience event level, we did not want to touch that, as that might hamper their understanding.
cc @rkpandey

kstreeter:
@meetraviagg, AAM team uses segment membership in ExperienceEvent.  @harleensahni has reviewed and is okay with this change.  We need to apply it consistently.

meetraviagg:
[As per my conversation with Raghavendra and Vineet]
In that case @kstreeter can we track the experience-event change in a separate PR.
If not, @harleensahni can you or someone from team please pitch in to make the relevant changes.
cc @rkpandey

meetraviagg:
Hi @kstreeter, in interest of time for now i have introduced segmentMap in experience-event as well. I have done it in same way as profile i.e. created a new field instead of changing the old one to not break anything. That's why segmentmembershipitem still remains.
cc @rkpandey

kstreeter:
@meetraviagg, @harleensahni please see PR #602.  There was some cleanup required on the original PR, and I did not have direct access to the branch so I opened a new one.

kstreeter:
As this PR is now superseded by #602 I am closing

PR #598, opened 2019-02-02
AdditionalParams map for dynamic everchanging data

ankushkd:
PR allows to add additional params map which will help solutions capturing dynamic ever changing data on an experience event.


kstreeter:
Hi @ankushkd, this topic has come up before.  We are not going to include a generic bucket into ExperienceEvent.  If you need to capture semi-structured data in a map you can, but it needs to be in your extension namespace.  I am going to close this PR.

Issue #599, opened 2019-02-08
Check breaking changes for updated schemas


jwen-adobe:
## What are the schemas that are affected by the issue
All schemas modified in the PR.

## What are examples of products that are impacted by the issue
platform xdm and tooling.

PR #600, opened 2019-02-08
Add script to check breaking changes of modified schemas [Plat 27179]

jwen-adobe:
This PR links to the issue #599. 

The incompatibility check script will check breaking changes as mentioned below and throw out errors to fail the build start.

-when an existing schema file got removed (rename a file will be breaking change in this case)
-when an field got removed from the schema
-when a field's data type got changed
-when the value of $ref is changed
-when a $ref got removed from allOf

The script will generate warning messages for cases below.

-when the meta:extends does not match allOf

@cmathis @chrisdegroot


kstreeter:
This is awesome @jwen-adobe thank you!

PR #601, opened 2019-02-14
Refactored adCloud Profile extension for schema composition.

cdegroot-adobe:
AdCloud has a `context/profile` extension that was not converted to the new schema composition pattern. This refactors that solution extension to bring it inline with the other Adobe Solution extension.

- Created `extensions/adobe/experience/adcloud/profile-all.schema.json`
- Created `extensions/adobe/experience/adcloud-profile.schema.json`
- Updated examples
- Removed legacy schema.


cdegroot-adobe:
@kstreeter - can you have a look at this, I will do one more review with the adcloud team than then we should be ready to merge.

cdegroot-adobe:
@kstreeter - This is failing because of the "Incompatibility Check". I though they were just supposed to be warnings. Yes I am breaking stuff here. But that is the intent in this case.

Issue #595, opened 2019-01-31
Introduce segmentMap in Profile schemas


meetraviagg:
## What are the schemas that are affected by the issue
Profile
UPS only understands/uses segment map at profile level. That's why this PR is to target segment membership only in profile schema and not in any other schema. Right thing to do is to approach the team who owns/uses segment memberships in other schemas for any changes in schemas other than Profile.

## What are examples of products that are impacted by the issue
ACP

## Reason for change
Map's are usually easy to edit (as they act as json objects in cosmos record). But with arrays one need to read all elements first and then make any changes and then write the array back. Also indexes can be created easily when dealing with maps/json-objects.

PR #602, opened 2019-02-15
Issue 595 - Moving segment membership from array to map

kstreeter:
Addresses issue #595.  This PR supercedes #593.

### DIFF
--- a/schemas/context/experienceevent-segmentmembership.example.1.json
+++ b/schemas/context/experienceevent-segmentmembership.example.1.json
@@ -1,18 +1,26 @@
 {
-  "xdm:segmentMemberships": [{
-  	"xdm:segmentID": {
-  	    "@id": "https://data.adobe.io/entities/segmentIdentity/92312748749128",
-  	    "xdm:namespace": {
-  	      "xdm:code": "AAM"
-  	    }
+  "xdm:segmentMembership": {
+    "AAM": {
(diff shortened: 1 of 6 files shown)


kstreeter:
The structure in this PR is basically the same as the one proposed in #593, but re-uses the original mixins so that the definitions are properly pulled into all of the solution extensions.  I also added the (optional) profile stitch ID to the map item, so that profile and experienceevent now have matching definitions for segment membership.

meetraviagg:
@kstreeter since we are changing existing definitions, will this be released as new xdm version. Because if xdm version is not changing wont it break the existing clients who are still expecting segment-array instead of segment-map ?

kstreeter:
Hi @meetraviagg yes this is a breaking change, but we have a number of those planned for 09.9 as we address issues based on feedback from users.

meetraviagg:
Hi @kstreeter, so will it be released as xdm version 0.9.9.9 itself ??
If that will be the case, it will be a problem for us.
@rkpandey - please confirm ...

meetraviagg:
Hi @kstreeter i talked to Raghav, and we are ok if these changes are going in 0.9.9.9.
But i am hoping that before any deployment happens, the timelines for that will be synced with profile team, so that clients are not affected. Thanks.
cc @rkpandey

kstreeter:
Okay, this one is in.  Note that after some discussion with @cdegroot-adobe we decided to keep the original definitions temporarily to give folks a chance to absorb the change.

PR #604, opened 2019-02-19
Added back endUserIDs for ExperienceEvent, enabled it for solution extensions.

cdegroot-adobe:
This PR brings back endUserIDs for Experience Event and clarifies that it contains the user ids asserted during capture of the EE.

LEFT UNTOUCHED:
context/experienceevent-enduserids-deprecated.schema.json, that will 
need to be removed later.

NON BREAKING CHANGE:
to solution template mixins to use new non 
deprecated endUserIDs


cdegroot-adobe:
@kstreeter - This is the PR to re-instate end user identities in ee. The approach I take is to add them as a mixin. I also considered adding endUserIDs to the EE class and could go that way if you think it is better.

PR #607, opened 2019-02-20
Add Profile Directmarketing Addresses

cdegroot-adobe:
The context/profile has standardized fields for personal(home) and work addressing (postal addresses, email addresses, phone numbers etc) and these work well for their intended purposes.

The issue arises where a set of addresses are to be used for direct marketing to a profile. In the B2B case work addresses make sense, and in the consumer b2c case the home* addressing makes sense. But there has further been issues with joint b2b and b2c cases in the same implementation as well as where a system needs to maintain both but use a mix for direct communication. 

Standard customer implementations use a dedicated direct marketing set of addresses that are distinct from the work* and home* addresses.

The recommendation from implementors of Direct Marketing solutions implementors is what has been executed here and uses the enhanced directmarketing-* data types which include quality measures for the address interactions.

### DIFF
--- a/schemas/context/profile-directmarketing.schema.json
+++ b/schemas/context/profile-directmarketing.schema.json
@@ -16,17 +16,17 @@
   "definitions": {
     "profile-directmarketing": {
       "properties": {
-        "xdm:address": {
+        "xdm:directMarketingAddress": {
           "title": "Direct Marketing Address",
           "$ref": "https://ns.adobe.com/xdm/context/directmarketing-address",
           "description": "Direct Marketing postal address."
         },
-        "xdm:email": {
+        "xdm:directMarketingEmail": {
           "title": "Direct Marketing Email",
           "$ref": "https://ns.adobe.com/xdm/context/directmarketing-emailaddress",
           "description": "Direct Marketing email address."
         },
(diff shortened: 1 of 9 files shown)


schemas/context/profile-directmarketing.schema.json
```diff
@@ -0,0 +1,46 @@
...
+  "title": "Profile Direct Marketing",
+  "type": "object",
+  "meta:extensible": true,
+  "meta:abstract": true,
+  "meta:intendedToExtend": ["https://ns.adobe.com/xdm/context/profile"],
+  "description": "Direct Marketing addressing for profiles.",
+  "definitions": {
+    "profile-directmarketing": {
+      "properties": {
+        "xdm:address": {
```
aarjain: Keeping the field name as 'address' may be a little ambiguous for the marketer. If I am not wrong, this field would be accessible to the marketer through the path /address/*.
Can we have something on the lines of directMarketingAddress/targetedAddress? What do you think?

cdegroot-adobe: Good catch, there is an issue here, let me fix it.

eknee:
Looks good!

PR #610, opened 2019-02-22
Modifying the directmarketing-emailaddress ref to point to the emailaddress datatype

aarjain:
The ref was initially pointing to phonenumber. Modified that.


cdegroot-adobe:
I have created https://github.com/adobe/xdm/pull/611 to replace this one.

PR #612, opened 2019-02-25
Created descriptorOneToOne Schema

polavishnu4444:
Adding the definitions for the specific descriptors.

- Added incomplete descriptorOneToOne schema definitions.
- Added the required fields and the optional fields constraints. 
- Made sourceSchema required in the schemadescriptor file.
- Need to see how to use display name reference during created of descriptors.

- Any relationship descriptor needs to have either sourceProperty or a label defined during relationship definition.

Other callout: validation of destination schema to have identity field is not modelled or cannot be created in this schema definitions.

Any ideas on how to handle this would help.


schemas/common/descriptors/relationshipdescriptor.schema.json
```diff
@@ -10,7 +10,7 @@
   "title": "Relationship Descriptor",
   "meta:extensible": true,
   "meta:extends": [
-    "https://ns.adobe.com/xdm/common/desciptors/schemadescriptor#/definitions/descriptor"
+    "https://ns.adobe.com/xdm/common/descriptors/schemadescriptor#/definitions/descriptor"
```
polavishnu4444: This is being considered as remove action in allOf, because of which build is failing with incompatibility changes.

schemas/common/descriptors/relationshipdescriptor.schema.json
```diff
@@ -40,7 +40,7 @@
   },
   "allOf": [
     {
-      "$ref": "https://ns.adobe.com/xdm/common/desciptors/schemadescriptor#/definitions/descriptor"
+      "$ref": "https://ns.adobe.com/xdm/common/descriptors/schemadescriptor#/definitions/descriptor"
```
polavishnu4444: This is being considered as remove action in allOf, because of which build is failing with incompatibility changes.

schemas/common/descriptors/relationship/descriptorOneToOne.schema.json
```diff
@@ -0,0 +1,68 @@
...
+      "$ref": "https://ns.adobe.com/xdm/common/desciptors/schemadescriptor#/definitions/descriptor"
+    },
+    {
+      "$ref": "https://ns.adobe.com/xdm/common/descriptors/relationshipdescriptor#/definitions/relationshipdescriptor"
+    },
+    {
+      "$ref": "#/definitions/descriptorOneToOne"
+    },
+    {
+      "oneOf": [
```
r-prasanth: why not add this to the base descriptor itself? requiring a label or a src-property should be relevant to all relationships.

schemas/common/descriptors/relationshipdescriptor.schema.json
```diff
@@ -35,6 +35,19 @@
           "$ref": "https://ns.adobe.com/xdm/common/descriptors/itemselector#/definitions/selector"
         }
       },
+      "oneOf": [
```
polavishnu4444: should this be anyOf or are we restricting it to oneOf?

kstreeter: Hey @polavishnu4444, I'm not sure the registry tooling will handle this oneOf, @cmathis can you comment?

schemas/common/descriptors/relationship/descriptorOneToOne.schema.json
```diff
@@ -0,0 +1,41 @@
...
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/common/descriptors/descriptorOneToOne",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "One to One Relationship Descriptor",
+  "meta:extensible": false,
+  "meta:extends": [
+    "https://ns.adobe.com/xdm/common/desciptors/schemadescriptor#/definitions/descriptor",
```
kstreeter: Don't need the #/definitions here, just the schema URI

polavishnu4444:
@adobe/open-source-office
This PR is blocking for CLA sign. My adobe LDAP: pola

schemas/common/descriptors/relationshipdescriptor.schema.json
polavishnu4444: @kstreeter 
This I assumed since we will be using it with some standard nodejs validator it would work.
Also today the all of merged result is having the oneOf values also at the top parallel to properties section.

kstreeter: Hi @polavishnu4444 it's not so much the validation that's a problem, it's tooling like protobuf and POJO generation that breaks down.  We should probably just the properties optional at this point.  @cmathis please correct me if this won't be an issue

r-prasanth: @kstreeter if we make the properties optional, then there is not much of a value-add in this PR. The main intend of the change was to try and make sure the constraints between the various properties in a relationship descriptor are visible to clients and not hidden in the server-side code.

@cmathis do we filter-out 'oneof' usages when loading schemas from git? I do see some existing schemas that contain 'oneof' in this repo - https://github.com/adobe/xdm/search?p=1&q=oneof&unscoped_q=oneof

polavishnu4444: Yeah,
I verified these in the local XDMR and also verified data payload with the schema generated (by combining the base descriptors fields) by XDMR with these definitions in global.

polavishnu4444: Yeah Craig,
I verified these in the local XDMR and also verified data payload with the schema generated (by combining the base descriptors fields) by XDMR with these definitions in global.

PR #613, opened 2019-02-26
Typo fix schema descriptor

polavishnu4444:
Schema descriptor ID typo fix.


schemas/common/descriptors/relationshipdescriptor.schema.json
```diff
@@ -40,7 +40,7 @@
   },
   "allOf": [
     {
-      "$ref": "https://ns.adobe.com/xdm/common/desciptors/schemadescriptor#/definitions/descriptor"
+      "$ref": "https://ns.adobe.com/xdm/common/descriptors/schemadescriptor#/definitions/descriptor"
```
polavishnu4444: This is being considered as remove action in allOf, because of which build is failing with incompatibility changes.

PR #620, opened 2019-03-08
Implemented the Adhoc Behaviour.

cdegroot-adobe:
Implemented the Adhoc Behaviour.

When included (meta:extends+allOf) in a schema this indicates to processors that the schema is not normalized with any other schemas and that the data must be assumed not to be semantically interoperable with any other data defined with an XDM schema.


schemas/data/adhoc.schema.json
```diff
@@ -0,0 +1,30 @@
...
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/data/adhoc",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Adhoc Schema",
+  "type": "object",
+  "meta:extensible": true,
+  "meta:abstract": true,
+  "description":
+    "Used to indicate the behaviour of Adhoc data semantic when composed into data Schemas. The Adhoc data semantic scope is limited to just the local schema, and no compatibility can be assumed between schemas.",
```
rkpandey: just wondering if it is a good idea to be specific and mention that field values can not be assumed to be standardized/normalized with existing schemas.

cdegroot-adobe: @rkpandey  The description currently says ""The Adhoc data semantic scope is limited to just the local schema, and no compatibility can be assumed between schemas."" Is that ok?

kstreeter:
Hi @adobe/open-source-office can you sign off my CLA please?  Thank you!

PR #628, opened 2019-03-20
Add segment code to Campaign ExperienceEvent.

### LINKED ISSUE #627: Add segmentCode in Campaign Experience Event
extensions/adobe/experience/campaign/experienceevent-all.schema.json

Adobe Campaign

saurabhere:
Fixes #627 .


saurabhere:
@kstreeter @trieloff Review/Merge this please?

extensions/adobe/experience/campaign/experienceevent-all.schema.json
```diff
@@ -126,6 +126,11 @@
               "title": "External ID",
               "type": "string",
               "description": "An ID generated by the parent activity invoking service to check later if the incoming experienceEvent was generated using this ID."
+            },
+            "xdm:segmentCode": {
```
kstreeter: Hey @saurabhere, I think elsewhere we have referred to "the thing that refers to a definition of a segment" using "xdm:segmentID".  Would that be appropriate here?

kstreeter: Hey @saurabhere can you comment on this?

saurabhere: yes, by the looks of it, i can probably use `segmentMembership/payload/payloadStringValue`. It does not seem to be exposed as a mixin. I will have to add a $ref to `https://ns.adobe.com/xdm/context/segmentmembership` here.
Is that how you see it too?

kstreeter: @saurabhere, I was actually just thinking of using the same property name, `xdm:segmentID`.  But I am assuming that the code here is semantically the same thing, ie "the ID or name of the segment definition, as provided by the system that own the definition".  If that's right, let's just change the name, rather than include the entire membership definition.

saurabhere:
@kstreeter does this looks good?

PR #634, opened 2019-03-29
eventID should be usable without orchestrationDetails: created a sepa…

reivaxy:
…rate mixin

Right now, in the https://ns.adobe.com/experience/campaign/experienceevent-all mixin, where the eventID field is defined, there is also the orchestrationDetailselement, which has required fields.

Both elements do not have the same goal. EventID needs to be in events processed by Voyager to make journeys works, whilst orchestrationDetails is a set of values that records what happens in Voyager.

We need to be able to use eventID without orchestrationDetails.

I have made changes consisting of:
-	Creating a new mixin that defines the eventID field (with the same json hierarchy as before)
-	Using this mixin in the experiencevent-all mixin.

I have two concerns:
-	Is it the right way to do it? I thought it might require creating a json schema defining the eventID field, and then create a mixin using this json schema. And also reference this json schema in experienceevent-all mixin
-	What about existing schemas (possibly for which data was ingested) that are using the experienceevent-all mixin ? I think it should be fine since the final schema definition is not changed?


extensions/adobe/experience/campaign/experienceevent-all.schema.json
```diff
@@ -244,8 +244,8 @@
           "properties": {
             "xdm:eventID": {
               "title": "Event ID",
-              "type": "string",
-              "description": "ID that identifies the event sent by the data source. This is used by Campaign orchestration to identify the event without inspecting its payload to determine which action should be triggered when the event is received. The value is automatically generated in Campaign at event creation time, it is globally unique across organizations."
+              "description": "ID that identifies the event sent by the data source. This is used by Campaign orchestration to identify the event without inspecting its payload to determine which action should be triggered when the event is received. The value is automatically generated in Campaign at event creation time, it is globally unique across organizations.",
+              "$ref": "https://ns.adobe.com/experience/campaign/orchestration/eventid"
```
kstreeter: Hi @reivaxy, we need to change this in order to avoid a double-nesting of the `xdm:eventID` property.  The best way to do this is simply to re-define the eventID right here, rather than reference it.  This will compose as expected.

reivaxy: Thanks @kstreeter, I did it, please check it's good now.

PR #639, opened 2019-04-03
Annotation XDM data model

adipan:
1. Implementation for annotation data model along with examples.
2. Introducing principalprofile schema that can be useful in collaborative workflows besides review/annotation use-cases.

This PR should be related to issue #7 and #8


adipan:
@lrosenthol Thanks for providing feedback earlier. It would be great if you could please review this PR once.

PR #640, opened 2019-04-03
Annotation XDM data model

adipan:
1. Implementation for annotation data model along with examples.
2. Introducing principalprofile schema that can be useful in collaborative workflows besides review/annotation use-cases.

This PR should be related to issue #7 and #8


adipan:
@lrosenthol could you please review this PR related to annotations data model?

schemas/assets/annotation.schema.json
```diff
@@ -0,0 +1,360 @@
...
+        },
+        "adbcollab:strokeWidth": {
+          "title": "Stroke width",
+          "description": "stroke width for drawing the annotation",
+          "type": "number",
+          "default": 1
+        },
+        "adbcollab:opacity": {
+          "title": "Opacity",
+          "description": "Fill and stroke opacity of the annotation",
```
adipan: This is similar to CA entry in Table 170 in PDF 32000-1:2008 
(Optional; PDF 1.4) The constant opacity value that shall be used in painting the annotation (see Sections 11.2, “Overview of Transparency,”and 11.3.7, “Shape and Opacity Computations”). This value shall apply to all visible elements of the annotation in its closed state.
--

schemas/common/principalprofile.example.1.json
```diff
@@ -0,0 +1,9 @@
+{
+  "xdm:provider": {
```
adipan: I saw the example of https://github.com/adobe/xdm/blob/master/schemas/common/principal.example.1.json to arrive at this. I think xdm:provider should be just a URL but I am not sure if I can decide such a value on my own.

PR #648, opened 2019-04-10
Added page slot and position of listing in search result.

### LINKED ISSUE #647: Search Schema - Add slot & position properties for Paid Search
`schemas/context/search.schema.json` has added ways to post experience event tracking hits from search results page. 

## Questions
1. does `pageDepth` property mean the page number?
1. we are using the term `keywords`, does this mean the `query` given by the surfer in the search bar? Most of the search engines refer this as `query` 

## Suggestions
Search results page also has the following attributes which are missing here,
1. Slot => Is the listing appearing in top or side? There could be other slots as well, which depends on the search engine or search partner. For reference check [AdWords Value Tracking Documentation](https://developers.google.com/adwords/api/docs/guides/valuetrack-mapping), look for `{adposition}`
1. Position => the position of the listing, first in the page? fifth in the page?

sathyz:
Fixes #647


schemas/context/search.schema.json
```diff
@@ -39,6 +39,16 @@
           "title": "Page Depth",
           "type": "integer",
           "description": "The page depth in the search results."
+        },
+        "xdm:slot": {
+          "title": "Page Slot",
+          "type": "string",
+          "description": "The section of the page where the search result appeared. For eg. Top, Side"
```
cdegroot-adobe: Is it possible to mane this an enumeration, even just a soft enum (just meta:enum) that acts like a set of recommendations?

PR #655, opened 2019-04-17
Refactoring, corrections - Complex and buggy syntax caused issues to processors.

cdegroot-adobe:
Refactored the syntax to simplify processing and resolve errors that 
seem to be in the syntax that cause processors to fail.
Implemented the full list of channels, some were omitted before.
Simplified syntax also resolve critical issues such as not having titles 
and descriptions.

Non breaking changes.

### DIFF
--- a/schemas/context/optinout.schema.json
+++ b/schemas/context/optinout.schema.json
@@ -1,118 +1,307 @@
 {
-  "meta:license": [
+   "meta:license": [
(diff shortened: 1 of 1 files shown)


kstreeter:
Hey @cdegroot-adobe, is there any way to avoid copying all of the enumeration options, etc and still have the tooling work?  This is a good change, but it's definitely unfortunate to have all of the copy/paste.

PR #667, opened 2019-05-01
[AMO-128509] - Ad Cloud Search Metadata Schema

jishah-adobe:
Please link to the issue #…


extensions/adobe/experience/adcloud/search/account-all.schema.json
```diff
@@ -0,0 +1,65 @@
...
+  "$id": "https://ns.adobe.com/experience/adcloud/search/account-all",
+  "title": "Adobe AdCloud Search Account Entity Full Extension",
+  "type": "object",
+  "meta:extensible": true,
+  "meta:abstract": true,
+  "meta:intendedToExtend": [
+    "https://ns.adobe.com/xdm/context/profile-adcloud"
+  ],
+  "description": "Adobe AdCloud Search Account Entity Full Extension. Contains all Solution added fields.",
+  "definitions": {
```
lrosenthol: All these fields need a proper namespacing, such as `xdm`.  (eg. `xdm:account`, `xdm:currency`)

extensions/adobe/experience/adcloud/search/account-all.schema.json
```diff
@@ -0,0 +1,65 @@
...
+      "properties": {
+        "account": {
+          "title": "Adobe Ad Cloud Search Account Top Level Fields",
+          "$ref": "https://ns.adobe.com/experience/adcloud/search/advertisingattributes",
+          "description": "Adobe Ad Cloud Top Level Entity Fields."
+        },
+        "currency": {
+          "title": "Currency the account is operated in",
+          "type": "string",
+          "description": "Currency associated with the account"
```
lrosenthol: should give examples?  For example, do I put "dollars" or "USD" or what??

extensions/adobe/experience/adcloud/search/account-all.schema.json
```diff
@@ -0,0 +1,65 @@
...
+        },
+        "managed": {
+          "title": "Is account managed or self serve",
```
lrosenthol: Why is this a string and not a boolean?

extensions/adobe/experience/adcloud/search/adgroup-all.schema.json
```diff
@@ -0,0 +1,70 @@
...
+    "title": "Adobe AdCloud Search Adgroup Full Extension",
+    "type": "object",
+    "meta:extensible": true,
+    "meta:abstract": true,
+    "meta:intendedToExtend": [
+        "https://ns.adobe.com/xdm/context/profile-adcloud"
+    ],
+    "description": "Adobe AdCloud Search Adgroup Full Extension. Contains all Solution added fields.",
+    "definitions": {
+        "adgroup-all":{
```
lrosenthol: Again, all these fields need to be in a namespace

extensions/adobe/experience/adcloud/search/adgroup-all.schema.json
```diff
@@ -0,0 +1,70 @@
...
+            "properties": {
+                "adgroup": {
+                    "title": "Adobe Ad Cloud Search Adgroup Top Level Fields",
+                    "$ref": "https://ns.adobe.com/experience/adcloud/search/advertisingattributes",
+                    "description": "Adobe Ad Cloud Top Level Attributes for Adgroups."
+                },
+                "campaignInternalId": {
+                    "title": "Campaign Unique Identifier",
+                    "type": "integer",
```
lrosenthol: is the UID really just an integer?

extensions/adobe/experience/adcloud/search/adgroup-all.schema.json
```diff
@@ -0,0 +1,70 @@
...
+                        "vertical": "Vertical",
+                        "age_range": "Age Range",
+                        "keyword": "Keyword",
+                        "adgroup": "Adgroup",
+                        "user_interest_and_list": "User Interest and List"
+                    }
+                },
+                "mobileBidAdjustment": {
+                    "title": "Mobile Bid Adjustment",
+                    "type": "integer",
```
lrosenthol: if it's a percentage, shouldn't this be a number and not an integer?

extensions/adobe/experience/adcloud/search/adgroup-all.schema.json
```diff
@@ -0,0 +1,70 @@
...
+                    "description": "Percentage determining the change in bid for mobile ads."
+                },
+                "searchMaxCPC": {
+                    "title": "Seach Max Cost per Click",
+                    "type": "integer",
```
lrosenthol: cost in what units?  Should this be a number as well?

extensions/adobe/experience/adcloud/search/advertisingattributes.schema.json
```diff
@@ -0,0 +1,61 @@
...
+    ],
+    "$schema": "http://json-schema.org/draft-06/schema#",
+    "$id": "https://ns.adobe.com/experience/adcloud/search/advertisingattributes",
+    "title": "Ad Cloud Top Level Entity Attributes DataTypes",
+    "type": "object",
+    "description": "Ad Cloud Top Level Entity Attributes DataTypes",
+    "definitions": {
+        "advertisingattributes": {
+            "properties": {
+                "externalId": {
```
lrosenthol: again, these need to be namespaced...

extensions/adobe/experience/adcloud/search/advertisingattributes.schema.json
```diff
@@ -0,0 +1,61 @@
...
+                    "description": "Name of the entity."
+                },
+                "description": {
+                    "title": "Description",
+                    "type": "string",
+                    "description": "Description of the entity."
+                },
+                "status": {
+                    "title": "Status",
+                    "type": "string",
```
lrosenthol: are there a fixed set of status?  could this be an enum?

extensions/adobe/experience/adcloud/search/campaign-all.schema.json
```diff
@@ -0,0 +1,107 @@
...
+    "meta:extensible": true,
+    "meta:abstract": true,
+    "meta:intendedToExtend": [
+        "https://ns.adobe.com/xdm/context/profile-adcloud"
+    ],
+    "description": "Adobe AdCloud Search Campaign Entity Full Extension. Contains all Solution added fields.",
+    "definitions": {
+        "campaign-all": {
+            "properties": {
+                "campaign": {
```
lrosenthol: same comment - namespacing

extensions/adobe/experience/adcloud/search/portfolio-all.schema.json
```diff
@@ -0,0 +1,55 @@
...
+    "type": "object",
+    "meta:extensible": true,
+    "meta:abstract": true,
+    "meta:intendedToExtend": [
+        "https://ns.adobe.com/xdm/context/profile-adcloud"
+    ],
+    "description": "Adobe AdCloud Search Portfolio Full Extension ",
+    "portfolio-all": {
+        "properties": {
+            "portfolio": {
```
lrosenthol: namespacing...

extensions/adobe/experience/adcloud/search/account-all.example.1.json
```diff
@@ -0,0 +1,15 @@
+{
+    "@id": "100001061",
```
lrosenthol: `@id` is special in JSON-LD.  You probably want to be using `xdm:id`

extensions/adobe/experience/adcloud/search/adgroup-all.example.1.json
```diff
@@ -0,0 +1,16 @@
+{
+    "@id": 1792062465,
```
lrosenthol: `@id` is special in JSON-LD.  You probably want to be using `xdm:id`

extensions/adobe/experience/adcloud/search/campaign-all.example.1.json
```diff
@@ -0,0 +1,22 @@
+{
+    "@id": "100000123",
```
lrosenthol: `@id` is special in JSON-LD.  You probably want to be using `xdm:id`

extensions/adobe/experience/adcloud/search/portfolio-all.example.1.json
```diff
@@ -0,0 +1,13 @@
+{
+    "@id": "100000123",
```
lrosenthol: `@id` is special in JSON-LD.  You probably want to be using `xdm:id`

schemas/context/admeasuressearch-adcloud.example.1.json
```diff
@@ -0,0 +1,4 @@
+{
+  "@id": "2419",
```
lrosenthol: `@id` is special in JSON-LD.  You probably want to be using `xdm:id`

schemas/context/profile-adcloud.example.1.json
```diff
@@ -0,0 +1,3 @@
+{
+    "@id": "1792062465"
```
lrosenthol: `@id` is special in JSON-LD.  You probably want to be using `xdm:id`

extensions/adobe/experience/adcloud/search/adgroup-all.schema.json
jishah-adobe: In the source datastore (Postgres) the field is defined as integer which fits well into XDM integer acceptable value ranges.

extensions/adobe/experience/adcloud/search/account-all.schema.json
jishah-adobe: This can be a boolean flag as there are only 2 possible values - managed and not managed. However many internal systems filter data based on these values rather than boolean flag. We will make this as an enum

extensions/adobe/experience/adcloud/search/account-all.schema.json
jishah-adobe: Added proper namespacing

extensions/adobe/experience/adcloud/search/account-all.schema.json
jishah-adobe: Added example for the currency field

extensions/adobe/experience/adcloud/search/adgroup-all.schema.json
jishah-adobe: Added fields under proper namespace

extensions/adobe/experience/adcloud/search/advertisingattributes.schema.json
jishah-adobe: Added under proper namespace

extensions/adobe/experience/adcloud/search/campaign-all.schema.json
jishah-adobe: Added fields under proper namespace

extensions/adobe/experience/adcloud/search/account-all.example.1.json
jishah-adobe: Fixed to use xdm:id

extensions/adobe/experience/adcloud/search/campaign-all.example.1.json
jishah-adobe: Resolved by replacing with xdm:id

extensions/adobe/experience/adcloud/search/portfolio-all.example.1.json
jishah-adobe: Resolved by replacing with xdm:id

schemas/context/profile-adcloud.example.1.json
jishah-adobe: Resolved by replacing with xdm:id

schemas/context/admeasuressearch-adcloud.example.1.json
jishah-adobe: Replaced with xdm:id

extensions/adobe/experience/adcloud/search/portfolio-all.schema.json
jishah-adobe: Added namespace to the fields

extensions/adobe/experience/adcloud/search/adgroup-all.example.1.json
jishah-adobe: Replaced example with xdm:id

jishah-adobe:
@lrosenthol @prabhum2

extensions/adobe/experience/adcloud/search/adgroup-all.schema.json
jishah-adobe: Mobile Bid Adjustment is a deprecated field and is not required in schema.

extensions/adobe/experience/adcloud/search/adgroup-all.schema.json
jishah-adobe: Search Max CPC is no longer required to be exposed and stored in platform.

extensions/adobe/experience/adcloud/search/advertisingattributes.schema.json
jishah-adobe: Converted status as an enum.

schemas/context/profile-adcloud.schema.json
```diff
@@ -0,0 +1,34 @@
+{
+  "meta:license": [
+    "Copyright 2017 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/context/profile-adcloud",
```
cmathis: In talking with the ad-cloud group, the purpose of this new class is because they didn't like the current Profile includes the identityMap be default. 
There has been talks to remove the identityMap from the class, so I would suggest we go that route and avoid an adcloud specific class within standard XDM.

schemas/context/admeasuressearch-adcloud.schema.json
```diff
@@ -0,0 +1,36 @@
+{
+  "meta:license": [
+    "Copyright 2017 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/context/admeasuressearch-adcloud",
```
cmathis: Classes are intended to have universal meaning across the platform. Is this class for an adcloud purpose only? If so, should it be moved under the adcloud namespace instead? 
What does this class do (or not do) that the existing experienceevent class does not provide?

schemas/context/profile-adcloud.schema.json
kstreeter: Agreed.  The identity map is optional anyway, so this should not be an issue.

schemas/context/admeasuressearch-adcloud.schema.json
kstreeter: Agreed, this should be in ad cloud namespace.  Maybe there was some mechanical reason this couldn't be done that we need to address?

extensions/adobe/experience/adcloud/search/account-all.schema.json
```diff
@@ -0,0 +1,75 @@
...
+    "meta:extensible": true,
+    "meta:abstract": true,
+    "meta:intendedToExtend": [
+        "https://ns.adobe.com/xdm/context/profile-adcloud"
+    ],
+    "description": "Adobe AdCloud Search Account Entity Full Extension. Contains all Solution added fields.",
+    "definitions": {
+        "account-all": {
+            "properties": {
+                "xdm:account": {
```
cmathis: Because this mixin (has the intendedToExtend) exists inside the adcloud namespace, all top level fields must be under an adcloud namespace rather than xdm. 
This can be solved two different ways:
Option 1: Instead of xdm:account the field should be https://ns.adobe.com/experience/adcloud/account
Option 2: You create a top level field named https://ns.adobe.com/experience/adcloud/account and then nest all the xdm: fields under it. 

See an example here - https://github.com/adobe/xdm/blob/master/extensions/adobe/experience/campaign/experienceevent-profile-push-details.schema.json#L19
2. You create a top level field named https://ns.adobe.com/experience/adcloud/account

schemas/context/admeasuressearch-adcloud.schema.json
jishah-adobe: Moved Ad Cloud Specific class under ad-cloud namespace.  The underlying data-set doesn't fall under experience event category (data is not time-series events rather aggregated data of number of clicks within a day). Hence we had to model as a ad-cloud search specific class.

extensions/adobe/experience/adcloud/search/account-all.schema.json
jishah-adobe: Thanks @cmathis. We chose option 1 and modified all top level fields under an ad-cloud namespace

ktukker-adobe:
@adcloud team - @jishah-adobe I would like to udnertand the use-cases and why this separate record - time-series schema is created ; where can I find out more about this?

schemas/context/profile-adcloud.schema.json
jishah-adobe: We can live with identity map for a short term unless it is removed from the class. Would removal of identity map be a breaking change for us - what I mean is do we have to re-patch our mix-in and schema at a later state? If not, we can live with what you have proposed above.

extensions/adobe/experience/adcloud/search/profile-adcloud.schema.json
```diff
@@ -0,0 +1,34 @@
+{
+  "meta:license": [
+    "Copyright 2017 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/experience/adcloud/search/profile-adcloud",
```
cfraser: Proifile is loaded jargon. It shouldn't be part of your $id especially since it doesn't have anything to do with the existing profile class.

extensions/adobe/experience/adcloud/search/profile-adcloud.schema.json
```diff
@@ -0,0 +1,34 @@
...
+  "title": "Adobe Ad Cloud Record",
+  "type": "object",
+  "auditable": false,
+  "meta:extensible": true,
+  "meta:abstract": true,
+  "meta:extends": [
+    "https://ns.adobe.com/xdm/data/record"
+  ],
+  "description":
+    "An Adobe Ad Cloud Record is a custom class to represent and define Ad Cloud Entity Level Attributes.",
```
cfraser: This description should be improved. When should I use this class? What are `Ad Cloud Entity Level Attributes`?

schemas/context/profile-adcloud.schema.json
cmathis: We are planning to remove it from the class so you will not have it long. Should be going in about the same time as this PR. As long as you do not use that field for ingestion it will not be a breaking change.

What you would need to do is change the meta:intendedToExtend attribute to reference the core context/profile instead. Let's do that now as part of this PR.

extensions/adobe/experience/adcloud/search/account-all.schema.json
```diff
@@ -0,0 +1,75 @@
...
+        "of the License at https://creativecommons.org/licenses/by/4.0/"
+    ],
+    "$schema": "http://json-schema.org/draft-06/schema#",
+    "$id": "https://ns.adobe.com/experience/adcloud/search/account-all",
+    "title": "Adobe AdCloud Search Account Entity Full Extension",
+    "type": "object",
+    "meta:extensible": true,
+    "meta:abstract": true,
+    "meta:intendedToExtend": [
+        "https://ns.adobe.com/experience/adcloud/search/profile-adcloud"
```
cmathis: Would suggest change this to https://ns.adobe.com/xdm/context/profile. Will be removing the identityMap from this core class in another PR.

schemas/context/profile-adcloud.schema.json
jishah-adobe: Removed Ad-Cloud Specific Record class and instead used base profile class to define this mix-in.

extensions/adobe/experience/adcloud/search/profile-adcloud.schema.json
jishah-adobe: Description improved and made more meaningful across all other mix-in and data-types.

extensions/adobe/experience/adcloud/search/account-all.schema.json
jishah-adobe: Modified this across all objects to use base profile class pre-defined in XDM registry.

extensions/adobe/experience/adcloud/search/profile-adcloud.schema.json
jishah-adobe: Removed ad-cloud defined class

jishah-adobe:
> @adcloud team - @jishah-adobe I would like to udnertand the use-cases and why this separate record - time-series schema is created ; where can I find out more about this?

@ktukker-adobe - We are moving our reporting data from relational databases to Ad Cloud Platform. The aggregated metrics data falls under two categories - performance (cost/clicks/impressions) and conversions (registrations, orders, revenue).  Performance Metrics are pulled from ad platforms (search/social) and data we receive is aggregated at daily level across collection of attributes (e.g ad, device and match types).  The data we receive doesn't fall under time-series events.  However, data platform currently lacks mechanism to define custom defined partitions - in our case assume by day).  Also this data gets reconciled and needs to be re-processed at a bare minimum 3 times.  To facilitate some level of partitioning scheme the proposal from XDM team was to define our own time-series class, mix-in and data-types under ad-cloud namespace and then use them to define schemas per tenant.  The same theory holds true for attribute/entity tables.  They do not fall under profile a.k.a record class. We were able to define them as a custom/ad-hoc schemas in earlier version of XDM but since that restriction was lifted and schemas were forced to extend from a class we are building our attribute schemas using profile class though none of the fields will be populated and left empty. We can schedule a quick meeting on this if it helps to move faster.

extensions/adobe/experience/adcloud/search/admeasuresearch-adcloud.schema.json
```diff
@@ -0,0 +1,36 @@
+{
+  "meta:license": [
+    "Copyright 2017 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/experience/adcloud/search/admeasuresearch-adcloud",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Adobe Ad Cloud Aggregated Measures Time Series Class",
```
cmathis: @cfraser - Since this class will appear int he UI along side the existing "XDM ExperienceEvent" and "XDM Profile". What would you like them to name the class here?

jishah-adobe: Hi @cmathis @cfraser 
I updated the title and description to reflect the purpose and usage restriction of this ad-cloud solution specific class. Let me know if new title and description makes sense.

cmathis:
This PR looks good to me. Only waiting on a formal OK from PM around this new class which will appear in the UI along side the other two core classes.

extensions/adobe/experience/adcloud/search/admeasuresearch-adcloud.schema.json
```diff
@@ -0,0 +1,36 @@
+{
+  "meta:license": [
+    "Copyright 2019 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/experience/adcloud/search/admeasuresearch-adcloud",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Adobe Ad Cloud Search Aggregated Measure",
```
cfraser: Why benefits does this empty class bring you over using ExperienceEvent?

jishah-adobe: The sole purpose of having this class is to get data based partitioning functionality for ingesting daily aggregated measures/metrics.  The schema and underlying data-sets does not classify as an experience event.  The best outcome in this scenario was to create a new class under ad-cloud namespace.

ktukker-adobe:
My Open questions:

- In the IRD on the wiki (https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=EfficientFrontier&title=XDM+Schema+Review), the baseschema (class) is listed on the wiki as : “profile-adcloud” – I don’t see this defined in this PR. If it would be defined, what does a “Profile AdCloud” describe? Is is the consumer who clicks on an add? Is it the Adobe Account (Customer) who has accounts with google and other ad spaces? (“Profile” is too generic of a description).

- In Experience Platform “context/profile” soon will be renamed to “context/individual” to better represent the definition: A consumer, it’s attributes, behavior, identities and segment membership. In future, definitions for “Enterprise”, “Household” and “Neighborhood” might be defined. These all represent "Profiles" of different entities. Where does this object fit in?

- Current definition of “adgroup-all.schema.json” extends “context/profile” – is this the actual “profile-adcloud” mentioned above? Is this modeling correct? If I look at the definitions, these look more “account” level (Adobe customer) than “Individual” (Consumer level) definitions

PR #670, opened 2019-05-03
Added properties related to In-App message required for dynamic Reporting.

ayush-pr:


extensions/adobe/experience/campaign/experienceevent-all.schema.json
```diff
@@ -240,6 +240,11 @@
               "title": "URL Category",
               "type": "string",
               "description": "The category of the link. It may be `subscription`, or a user-defined category."
+            },
+            "xdm:inAppMessageAction": {
+              "title": "In-App Message Action",
+              "type": "string",
+              "description": "Actions performed on the In-App message. Actions can be Button 1 click, Button 2 click or Dismissals."
```
saurabhere: This could be an enum for tight bound on the proposed values:
```
Button1 click
Button2 click
Dismissal
``` 

Also. is it not duplicate information if you have this field in addition to the metrics `xdm:inAppDismissal` ?
That is, can it happen that the metric is `xdminAppDismissal` and the value of this is `Button2 Click`?

schemas/context/direct-marketing.schema.json
```diff
@@ -60,6 +60,21 @@
...
+        },
+        "xdm:impressions": {
+          "$ref": "https://ns.adobe.com/xdm/data/measure",
+          "description":
+            "The direct-marketing message ( Push Notifications, In-App etc) has been seen by the end user."
+        },
+        "xdm:inAppClicks": {
+          "$ref": "https://ns.adobe.com/xdm/data/measure",
+          "description":
+            "Click(s) actions on an In-App message."
```
saurabhere: Does this need its own clicks metric? There is already a `clicks` metric at 
https://github.com/adobe/xdm/blob/99643dace9e262773327dd570fdfd14561d416ce/schemas/context/direct-marketing.schema.json#L22

extensions/adobe/experience/campaign/experienceevent-all.schema.json
ayush-pr: The values are not tight bound, Values could be the label on the buttons or dismissal. I will update the description.
To answer the second question, for metric inAppDismissal inAppMessageAction will always be dismissal, but for metric inAppClicks inAppMessageAction is the label on the button that was clicked

schemas/context/direct-marketing.schema.json
ayush-pr: For InApp channel we calculated InAppClickThroughRate as ( unique InAppClicks ) / ( Impressions ).
In addition to this new clickThroughRate formula we also show inAppClick metric on the workspace. So this is required to distinguish between email clicks and inApp clicks.

extensions/adobe/experience/campaign/experienceevent-all.schema.json
saurabhere: Okay.
If that is the case, we can remove the `inAppDismissal` metric and use the clicks metric with the `inAppMessageAction` being `Dismissal`

schemas/context/direct-marketing.schema.json
saurabhere: Going purely by the xdm(so as to not be influenced by application logic), is it not enough to send `channel=inApp` and `metric=clicks` to infer what you need?

ayush-pr: Yes that should be enough

extensions/adobe/experience/campaign/experienceevent-all.schema.json
ayush-pr: It should be treated like some string similar to URL Label.

Issue #671, opened 2019-05-06
Request to add profile details to user


adipan:
## What are the schemas that are affected by the issue
https://github.com/adobe/xdm/blob/master/extensions/ims/user.schema.json

## What are examples of products that are impacted by the issue
CloudTech/ACP collaboration services. We need public profile details for various collaborative workflows - sharing, comment etc.

adipan:
Closing the issue as PR #672 addresses the same.

PR #672, opened 2019-05-06
Add public profile properties to user schema

adipan:
Adding public profile properties to user.schema.json so that it can be used in collaborative workflows.

Please link to the issue #…
https://github.com/adobe/xdm/issues/671


extensions/ims/user.schema.json
```diff
@@ -26,6 +26,23 @@
         "type": {
           "type": "string",
           "const": "https://ns.adobe.com/adobecloudplatform/ims/user"
+        },
```
lrosenthol: Can you also please fix id->xdm:id and type->xdm:type?  Since this is still an experimental schema, we can change anything...

extensions/ims/user.schema.json
```diff
@@ -36,6 +55,11 @@
...
     {
       "$ref": "#/definitions/user"
+    },
+    {
+      "$ref": "#/definitions/publicprofile"
     }
-  ]
+  ],
+  "required" : ["xdm:identityProvider", "id", "type"],
+  "meta:status": "experimental"
```
adipan: Hi @lrosenthol 
I wish to mark the schema as experimental and specify some required fields. I think the schema is not used at least in this repo. But, it is not marked experimental, unfortunately. Since it is not marked experimental, I was hesitant to change existing properties like changing id to xdm:id. Should I make the change?

PR #673, opened 2019-05-07
Campaign offer response detail

jbeckert:
Please link to the issue #…


jbeckert:
The incompatibility check fails because the data type schema for the property "https://ns.adobe.com/experience/campaign/offerOpened" changed from "https://ns.adobe.com/experience/offer-management/proposition-response-detail" to "https://ns.adobe.com/experience/campaign/offer-response-detail".
The latter is however derived from the former so it is type and structure extension.

extensions/adobe/experience/campaign/experienceevent-all.schema.json
```diff
@@ -283,7 +283,7 @@
         "https://ns.adobe.com/experience/campaign/offerOpened": {
           "title": "Offer Opened",
           "description": "The offer in the message for an 'offer opened' event.",
-          "$ref": "https://ns.adobe.com/experience/offer-management/proposition-response-detail"
+          "$ref": "https://ns.adobe.com/experience/campaign/offer-response-detail"
```
cmathis: Is this going to be a breaking change to the existing schema and data?

jbeckert: No, the Campaign team isn't able to generate data just yet. The needed the more specialized schema to finish their part. What this change does is to take a mere general schema offer-management/proposition-response-detail extend it to campaign/offer-response-detail and use the new, more specialized schema in the place where the old one was used.

jwen-adobe:
@cmathis @prabhum2 It seems the offer-response-detail.schema.json defined xdm:offer which already in the offer-proposition-detail.schema.json schema. Both of them are referenced by experienceevent-all.schema.json which is causing conflict now.

cmathis:
@jbeckert  - We should have looked at this PR more closely. It appears you are extending one mixin, but simply re-adding the same fields again to the new mixin. Is the whole goal here to simply create a new mixin under the campaign namespace that looks identical to the one you had under offer-managment?

jbeckert:
the new PR #681 may remove the conflict between the properties but then the semantic that https://ns.adobe.com/experience/campaign/offer-response-detail IS_A https://ns.adobe.com/experience/offer-management/proposition-response-detail gets lost
The schema definition for campaign/offer-response-detail is a compatible with schema definition for offer-management/proposition-response-detail because the rule applied to xdm:offer for the campaign/offer-response-detail is stricter (from a type relationship perspective) than the rule applied to xdm:offer for the offer-management/proposition-response-detail. All JSON objects that satisfy offer-management/proposition-response-detail also satisfy campaign/offer-response-detail

Now, an application that is looking to handle offer-management/proposition-response-detail can only hope -- but not infer -- that a JSON object from campaign is meant to be an offer-management proposition-response-detail.

cmathis:
Why did we change the reference to be a schema under the campaign namespace and not offer-management then? 

What you had before is not really valid JSON Schema or in XDM.  Your allOf was saying to include all the fields from offer-management/proposition-response-detail plus the fields from an internal definition inside of the new campaign/offer-response-detail. In doing this you were defining the same field at the same level (xdm:offer) two different ways. With an allOf this produces a collision in it's meaning and not a redefinition or override of that field.

jbeckert:
You may have established different rules than the JSON Schema standard but in the JSON Schema semantic allOf simply means all **rules** must be satisfied. The rules brought in via one item in the allOf list must hold true independently from the rules brought in via a different item from that list. It does not matter if the rules test repeatedly the same thing or if one rule set is a superset of the other. Obviously in an allOf list the rules cannot contradict each other because that schema wouldn't validate against anything.

Regarding your question "Why did we change the reference to be a schema under the campaign namespace and not offer-management then?" The Campaign team wanted the **datatype** for the xdm:offers property (which is embedded in the datatype https://ns.adobe.com/experience/campaign/offer-response-detail) to be extended by an additional property. If there is any why to do that without violating the schema assembler then I'd be interested to know that. However I can say that the JSON schema validator (ajv) I used did not have any issues with the schema definitions, specifically with the allOf for $ref https://ns.adobe.com/experience/offer-management/proposition-response-detail in addition to the definition given inline

cmathis:
You are right, it's likely valid JSON Schema and the AGV validator supports it for JSON validation, but its more an issue in the tooling we use to merge those different field definitions together and provide a concrete data ingestion and physical storage schema:

In the original PR you defined the xdm:offer two ways:

In offer-managment/offer-detail
xdm:offer { //An object that contains one attribute
    xdm:tags array[string]
}

in campaign/offer-detail
xdm:offer { //An object that contains two attributes
    xdm:eligibilityRule string
    xdm:tags array[string] . //included from the offer-management/offer-detail
}
In the new campaign/offer-response-detail you then try to say the final schema should contain both of those fields. The allOf resolver we use considers this two different definitions of the same xdm:offer field (and eventual storage layer struct) and doesn't try to automatically resolve the differences into a new object definition for that field.

jwen-adobe:
Another way to do this which will keep the semantic that https://ns.adobe.com/experience/campaign/offer-response-detail IS_A https://ns.adobe.com/experience/offer-management/proposition-response-detail is to keep the $ref in the allOf but not defining the fields again in its own. We do not want to have both as the de-reference process will see duplicate keys as conflict, we do want to alert these out though.

jbeckert:
For the offer-detail in the campaign namespace we are not repeating the properties. We only add another property.

>   "definitions": {
>     "offer-detail": {
>       "properties": {
>         "xdm:eligibilityRule": {
>           "type": "string",
>           "format": "uri",
>           "title": "Eligibility Rule",
>           "description": "The reference to an eligibility rule that tests a condition using the profile and/or other contextual XDM instances. The value is the URI (@id) of the eligibility rule that is referenced. See schema https://ns.adobe.com/experience/offer-management/eligibility-rule"
>         }
>       }
>     }
>   }

The issue seemed to be when that derived type is **_used_** in the two datatypes for response-detail that also derive from another.
What I just realized is the following: Why is there an issue in https://ns.adobe.com/experience/campaign/offer-response-detail that was fixed by removing the "extends" relationship to https://ns.adobe.com/experience/offer-management/proposition-response-detail when there is the schema for https://ns.adobe.com/experience/campaign/offer-proposition-detail (note proposition not response) that uses the exact same pattern. The tooling didn't flag that datatype schema!

jbeckert:
We ended up with a variant where only the additional attributes are added to the definition in the derived datatype https://ns.adobe.com/experience/campaign/offer-response-detail 
See PR #684 
This worked even with adding the additional property ("xdm:eligibilityRule") inside another property ("xdm:offer") that the base datatype schema https://ns.adobe.com/experience/offer-management/proposition-response-detail already had.

Issue #669, opened 2019-05-03
Request to add more properties in Principal, and Principal clean-up


adipan:
## What are the schemas that are affected by the issue
- [ ] https://github.com/adobe/xdm/blob/master/schemas/common/principal.schema.json
- [ ] https://github.com/adobe/xdm/blob/master/extensions/ims/user.schema.json

## What are examples of products that are impacted by the issue
ACP Collaboration services (sharing service, comment service, task service) and products that rely on ACP collaboration services.

## Details -
We are primarily interested in the following schema -
  https://github.com/adobe/xdm/blob/master/schemas/common/principal.schema.json

There is a related file that references the above instead of extending it, but appears to redefine similar properties (some without namespaces).
   https://github.com/adobe/xdm/blob/master/extensions/ims/user.schema.json

I think this is what Leonard suggested in principal.schema.json -

- [ ] If I got it right, Leonard suggested that @id and @type as used in principal.schema.json is wrong usage of @ in schema. It should have been xdm:id and xdm:type instead. The @id and @type are required properties in the existing principal.schema.json. I can not submit a change to fix this in principal.schema.json as npm run incompatibility-check fails (part of build job in github.com/adobe/xdm). I think the reason for that build step is that schemas can only be extended but not retrospectively changed. 
- [ ] This schema is marked non-extensible. Perhaps, that is purposely done?
 
If so, is it possible to add following optional properties to principal.schema.json itself -

- xdm:name,  
- profileImage (avatar image URI), 
- profileLink (like to your behance public profile homepage), 
- isUserAnonymous (boolean - since guest commenting could be there, we wish to identify anonymous users so they could be identified by a client app to avoid impersonation cases).
 
Oliver suggested that it may be better to make changes in the original principal schema to cover collaboration use-cases, instead of me trying to create a slightly different clone.

Would it be possible to make the requested changes in principal.schema.json (I would be happy to create a pull request), and if some properties do not make sense for your use-case, let me extend it?

PR #674, opened 2019-05-08
Changes to user, organization and principal schema to fix issue 669

adipan:
Please link to the issue #669 
**Summary**: Earlier, user and organization were not really principals. They both defined similar looking overlapping properties. This PR fixes the problem.

1. Principal schema is made extensible.
2. User and Organization schema is extended from Principal. 
3. User and Organization schemas are not really tied to IMS anymore. If `"xdm:identityProvider": "https://ns.adobe.com/adobecloudplatform/ims"` is set in entity-instance/example, then it is an IMS user otherwise, it could have been a user for another identity provider. 
4. Earlier, type property in user schema was not being used to denote 2 separate things - that the principal is a user and that the user belongs to IMS. So, there was really no need for `xdm:identityProvider`.
5. Earlier, there was no schema validation for any schemas that used user/organization. For example, please type `npm run test` and create an invalid schema referencing a user or a organization.
5. In principal schema, xdm:id and xdm:type is used instead of @id, @type. @id is supposed to be a uri-reference. In samples for principal, @id is an IMS guid/string and not a URI. So, @id was not being used for its purpose. Replacing that with xdm:id.
6. In principals schema, xdm:identityProvider is picked from user and organization schema. It is just a URI instead of being an object. identity-provider schema is deleted.

https://github.com/adobe/xdm/blob/master/schemas/common/principal.schema.json,
https://github.com/adobe/xdm/blob/master/extensions/ims/user.schema.json and https://github.com/adobe/xdm/blob/master/extensions/ims/organization.schema.json all have meta:status as experimental. So, I think breaking changes should be allowed.

Some dependent examples are affected too. Due to this, the number of files in this pull request is high. If I run `npm run package`, then there would be changes in `docs` folder too. I have avoid updating docs folder as part of this pull request.


lrosenthol:
Looks good to me - thanks!

prabhum2:
@lrosenthol - are you looking to merge this PR

PR #679, opened 2019-05-13
Format for loading global descriptors

r-prasanth:
[For review only, not for merging]

The XDM Registry team is planning to support loading global schema descriptors that are defined in the XDM git. This is a dummy PR to highlight the proposed conventions so that we can discuss and finalize the same. I will close the PR once the conventions are accepted.

Proposed Conventions:

- Global descriptors will be co-located with the schemas they are referring to
- Descriptors will be distinguished via the suffix ".descriptor.json"
- Like schemas, descriptors will contain a user defined id
- any referenced schema or property ids in the descriptor will be validated by the registry before its loaded

This is a sample PR for a user trying to create a link between two asset schemas by creating a "linked-asset-relationship.descriptor.json" file that resides along with the "asset.schema.json" file


r-prasanth:
@kstreeter @cdegroot-adobe please review

r-prasanth:
@prabhum2 please review

cmathis:
This needs to go along with how we define a "global" schema in this github repo as well. For the registry today, everything in this repo is considered a component (mixin, class, datatype, or behavior) with the actual "schema" being defined in the registry only through the composition of these components. Today the descriptors are always in context of the final "schema" and NOT the components themselves. 

If we introduce global descriptors, then it makes sense we would have global "schemas" as well, or we need to modify the registry logic to somehow inherit the descriptor information that is associated with one of the mixins, classes, etc.

schemas/external/repo/linked-asset-relationship.descriptor.json
```diff
@@ -0,0 +1,7 @@
+{
+  "@id": "https://ns.adobe.com/descriptors/sample-asset-relationship-descriptor",
```
lrosenthol: That is not a valid JSON-LD ID.  You probably want this to be `xdm:id` instead.

schemas/external/repo/linked-asset-relationship.descriptor.json
```diff
@@ -0,0 +1,7 @@
...
+  "@type": "xdm:descriptorOneToOne",
```
lrosenthol: That is not a valid JSON-LD type.  You probably want this to be `xdm:type` instead.

schemas/external/repo/linked-asset-relationship.descriptor.json
cmathis: Descriptors use an @id and @type - https://github.com/adobe/xdm/blob/master/schemas/common/descriptors/relationshipdescriptor.example.1.json

What makes this URI not valid?

r-prasanth:
@cmathis I

listing out the global-schema conventions for review:

- Any path/dir that is loaded into registry can contain global schemas - we don't have to separate out the definition of global schemas from any schema types.
- We currently distinguish between mixin/class etc based on the inheritance flags. Sticking to that model, we can set "meta:extensible=false" property to indicate that the file is a global schema. 
- Unlike tenant schemas, global schemas will not be composed exclusively via mixins. They can have contain refs to inline definitions. 

if we are good with the conventions I will add a sample global-schema file also to this PR

cmathis:
Let's not add a global schema concept until we have PM on the same page. As of now, they like the idea of having all the actual dataset schemas being tenant specific (made up of the small components defined in this repo) rather than a list of schemas unrelated to their usage of the platform. If we introduce a global schema, then we immediately show these across all tenants.

r-prasanth:
@cmathis I had a discussion with chris on the need for registry to support schemas that are available across tenants or in a tenant-less context (non-enterprise users). I think we are aligned on the overall concept and just need to explore ways to model those schemas and the means to load them into the registry.

/cc @cdegroot-adobe

PR #687, opened 2019-05-20
Amo 128508

jishah-adobe:
@cmathis @cfraser @kstreeter @ktukker-adobe @prabhum2


extensions/adobe/experience/adcloud/searchadvertisingcampaign.schema.json
```diff
@@ -0,0 +1,166 @@
...
+          "title": "Campaign ID",
+          "type": "string",
+          "description": "Campaign ID on the search advertising platform."
+        },
+        "xdm:campaignName": {
+          "title": "Campaign Name",
+          "type": "string",
+          "description": "Name of the Campaign."
+        },
+        "xdm:campaginDescription": {
```
cmathis: Campaign is misspelled

extensions/adobe/experience/adcloud/searchadvertisingcampaign.schema.json
```diff
@@ -0,0 +1,166 @@
...
+  "meta:extensible": true,
+  "meta:abstract": true,
+  "meta:auditable": true,
+  "meta:extends": ["https://ns.adobe.com/xdm/data/record"],
+  "description": "Search Advertising Campaigns is a set of ad groups (ads, keywords, and bids) that share a budget, location targeting, and other settings.",
+  "definitions": {
+    "searchadvertisingcampaign": {
+      "properties": {
+        "xdm:campaignId": {
+          "title": "Campaign ID",
```
cmathis: Given this field is defined as a core "xdm" field, does it need to have the exact same title as other locations? I see in the adcloud/campaign schema this field has a title of "Campaign Identifier".

extensions/adobe/experience/adcloud/searchadvertisingplatform.schema.json
```diff
@@ -0,0 +1,38 @@
...
+  "title": "Search Advertising Platform",
+  "type": "object",
+  "meta:extensible": true,
+  "meta:abstract": true,
+  "meta:intendedToExtend": ["https://ns.adobe.com/xdm/context/profile"],
+  "description": "Information about Search Advertising Platform.",
+  "definitions": {
+    "searchadvertisingplatform": {
+      "properties": {
+        "https://ns.adobe.com/experience/adcloud/adPlatform": {
```
cmathis: Why did we use "xdm" namespaced fields in the classes above, but adcloud name spaced fields here? Not saying one or the other is incorrect, just wanting to understand the reasoning.

extensions/adobe/experience/adcloud/searchadvertisingcampaign.schema.json
jishah-adobe: Fixed the spelling

extensions/adobe/experience/adcloud/searchadvertisingcampaign.schema.json
jishah-adobe: Modified all ID fields title to have identifier fully spelled rather than ID.

extensions/adobe/experience/adcloud/searchadvertisingplatform.schema.json
jishah-adobe: Initial thought was this class will be specific to ad-cloud as it is not a generic entity in search advertising market.  But it can leave in xdm namespace too.  I have modified it to be in xdm namespace

ktukker-adobe:
Looks like these schemas focus on the "search" use-case ; there are also display and video use-cases. I assume these would have their own classes. As such, instead of "namespacing" these with a name, can they live in their own namespace?

https://ns.adobe.com/experience/adcloud/search/
- advertising account

(move the "search" from the name to the namespace)

jishah-adobe:
> Looks like these schemas focus on the "search" use-case ; there are also display and video use-cases. I assume these would have their own classes. As such, instead of "namespacing" these with a name, can they live in their own namespace?
> 
> https://ns.adobe.com/experience/adcloud/search/
> 
> * advertising account
> 
> (move the "search" from the name to the namespace)

Yes they can live in their own namespace and I also understand from earlier discussion that the idea is to build a class that is generic to Search Advertising and can be used to build data-sets outside of a specific product - Ad Cloud.  If that is the approach then it makes more sense to have it as https://ns.adobe.com/experience/searchadvertising/. The reason being Search is much broader term and may create confusion with concept of Search Engines like Google, Bing.  However, we mentioned earlier that any class/schemas we create are exclusively for Ad Cloud use cases.  The reason being attributes vary across search advertising platforms.

cmathis:
I'm fine with this PR as is. But just something to think about:

Rather than make these classes work for Ad Cloud only, what about making these more generic searchadvertising classes with only the common fields that exist in all cases and then create Search Engine specific mixins based on what is required when that Search Engine is used. The Ad Cloud implementation of the class will always include every Search Engine mixin, but a non Ad Cloud user could use this to only manage their Google searchadvertising if thats all they had.

jishah-adobe:
> I'm fine with this PR as is. But just something to think about:
> 
> Rather than make these classes make sense for Ad Cloud only, what about making these more generic searchadvertising classes with only the common fields that exist in all cases and then create Search Engine specific mixins based on what is required when that Search Engine is used. The Ad Cloud implementation of the class will always include every Search Engine mixin, but a non Ad Cloud user could use this to only manage their Google searchadvertising if thats all they had.
> 
> The problem I see is that I don't know who makes this formal decision. Is it @ktukker, is it @prabhum2 , is it me??? This is the weird state that XDM is in.

We do not foresee anyone outside of Ad Cloud or at the best internal Adobe solutions to use this schema.  We selectively choose fields from search engines to build Ad Cloud Campaign.  In my opinion we should stay away from an idea of constructing shared schema within an advertising space (search or display).

cmathis:
Sounds good. I'm leaning on you from the search space to know if was a good idea or not. If @prabhum2 is good with these new classes and understands how they fit along side the current Profile and ExperienceEvent class then I'm good.

extensions/adobe/experience/adcloud/searchadvertisingaccount.schema.json
```diff
@@ -0,0 +1,91 @@
...
+  "type": "object",
+  "meta:extensible": true,
+  "meta:abstract": true,
+  "meta:auditable": true,
+  "meta:extends": ["https://ns.adobe.com/xdm/data/record"],
+  "description": "Search Advertising Account defines Customer Configuration on Search Advertising Platform.",
+  "definitions": {
+    "searchadvertisingaccount": {
+      "properties": {
+        "xdm:accountId": {
```
prabhum2: Why cant we use id that comes out of common record class here ?

extensions/adobe/experience/adcloud/searchadvertisingadgroup.schema.json
```diff
@@ -0,0 +1,121 @@
...
+  "meta:extensible": true,
+  "meta:abstract": true,
+  "meta:auditable": true,
+  "meta:extends": ["https://ns.adobe.com/xdm/data/record"],
+  "description": "Search Advertising Ad Group is a logical collection or grouping of ads.",
+  "definitions": {
+    "searchadvertisingadgroup": {
+      "properties": {
+        "xdm:adgroupId": {
+          "title": "Ad Group Identifier",
```
prabhum2: Why cant you use same _id that comes out of record here ?

jishah-adobe: Most of the entities (search) have two identifiers - one is ad-cloud identifier which uniquely identifies the record across search engines and other is external identifier (unique per search engine).  _id that comes out of record is used as an internal identifier.

extensions/adobe/experience/adcloud/searchadvertisingaccount.schema.json
jishah-adobe: Most of the entities (search) have two identifiers - one is ad-cloud identifier which uniquely identifies the record across search engines and other is external identifier (unique per search engine). _id that comes out of record is used as an internal identifier.

cmathis:
@jishah-adobe  - We need to make it clear that these new classes are not standard XDM but something under the adcloud namespace. Once we address the id fields we will merge this in, but @prabhum2 will continue to look into a better way to model this without solution specific classes (classes normally don't do this). Once this is available we would expect you to migrate to the official XDM classes and mixins. 

@cfraser and @ktukker   - These classes will appear in the UI along side the standard Profile and ExperienceEvents and there will be nothing today that makes it clear they are for AdCloud only. I can easily see users using the Advertising Campaign class simply because it has the fields they want.

extensions/adobe/experience/adcloud/searchadvertisingadgroup.schema.json
cmathis: What value would you populate in the id that comes from the record?

udayeyunni: @cmathis Internal Identifier comes from our product (AMO) databases. It is an unique Id across all search engine data that we work with.

Issue #689, opened 2019-05-28
Add an Optional Correlation ID to Experience Event


adobe-activation:
## What are the schemas that are affected by the issue
https://github.com/adobe/xdm/blob/master/schemas/context/experienceevent.schema.json

## What are examples of products that are impacted by the issue
Target, Analytics, ECID, Ad Cloud, Platform

**Summary**
The larger Experience Edge Project has a couple work streams which include Unified JavaScript and Unified Data Collection on a common Domain.  ( For details of that project see: https://wiki.corp.adobe.com/display/DMSArchitecture/Experience+Edge ) As this new unified JS library is sending data to Adobe rather than it's acquired solutions, it has an opportunity to fix a long standing issue of not being able to tie events (such as hits or page views) together into a single grouping.  For example in SPAs and mobile the concept of page loads or page views are fluid and there may not be a single point in time page load in the traditional sence. Another example is the A4T case where we do crazy gymanstics both client side and server side to tie a Personalization view to an Analytics Page View.  This this partcilar case the JS library is likely to send mutiple commuincations and mutiple events and yet the reporting is likley to want to group these together.  Ex: Top of page an ECID is requested and and authenticated state of the user is recorded, subsequently as personalization request for content is made, lastly a page view or page bottom call is made to record the visit and record which personalized experience or activity was visible.  Likely 3 commuincations or beacons were sent and yet there may be a desire to "stitch" these or group them together.   THis is just one of many examples. 

To address these and other use cases, we are proposing an optional grouping or Correlation ID be added to the Experience Event that consumers of the ID can use to stitch or group events together both with Adobe and Non-Adobe products whenever two or more events should be correlated together. 

**For more Details see:** https://wiki.corp.adobe.com/display/Activation/Universal+Event+Correlation

littlebutty:
After receiving alpha customer feedback, we have decided to rename this field to something less confusing and more specific.  "Stitch ID" primarily to drive users toward it's original purpose of stitching hits together similar to the A4T type use cases and less about session or attribution type use cases.

PR #690, opened 2019-05-28
Issue#689 add correlation id in experience events

adobe-activation:
Link to Issue: https://github.com/adobe/xdm/issues/689


schemas/context/experienceevent.schema.json
```diff
@@ -18,7 +18,13 @@
   "description": "An ExperienceEvent captures observations, including the point in time and identity of the subject involved. Experience Events are fact records of what occurred, representing what happened without aggregation or interpretation. They are critical for time-domain analytics as they allow for observation and analysis of changes that occur in a given window of time and the comparison between multiple windows of time to track trends. Experience Events can be either explicit (directly observable human actions) or implicit (raised without a direct human action).",
   "definitions": {
     "experienceevent": {
-      "properties": { }
+      "properties": {
+        "xdm:correlationID": {
```
cmathis: This should be named xdm:corrleationId (lower case d) to match the naming format of other ID fields. 

Also, I'm assuming this field is always populated by the data producer before Platform ingestion and not something that happens internally by the Platform so we may want to specify that in the description.

schemas/context/experienceevent.schema.json
```diff
@@ -18,7 +18,13 @@
   "description": "An ExperienceEvent captures observations, including the point in time and identity of the subject involved. Experience Events are fact records of what occurred, representing what happened without aggregation or interpretation. They are critical for time-domain analytics as they allow for observation and analysis of changes that occur in a given window of time and the comparison between multiple windows of time to track trends. Experience Events can be either explicit (directly observable human actions) or implicit (raised without a direct human action).",
   "definitions": {
     "experienceevent": {
-      "properties": { }
+      "properties": {
+        "xdm:correlationId": {
+          "title": "ExperienceEvent Correlation ID",
+          "type": "string",
```
clubycoder: I would suggest adding `format: "uuid"` to constrain the format, guide users to actually use a UUID, and reduce the chance of users from accidentally sending in junk like "1" or "test".

prabhum2: Brandon Pack - Do you agree with this recommendation ?

littlebutty: Yes, I agree with this.  Do we need to be more explicit and say UUID v4?

adobe-activation: added

schemas/context/experienceevent.example.2.json
```diff
@@ -1,5 +1,6 @@
 {
   "@id": "https://data.adobe.io/experienceid-2345678",
+  "correlationID": "9fb4e78d-b0bf-4d99-b987-cb37abdbcd15",
```
Aaronius: Should this be `correlationId`?

littlebutty:
We asked our external alpha users how they would implement and use this ID and they are very excited about it.  However we've decided to rename it to avoid some confusion.  For now I think wel'' close the PR and submit a new one after we've validated the changes with external users again.

schemas/context/experienceevent.schema.json
raresvla: @cmathis Is this naming convention enforced right now? I ask this because there are quite a few inconsistencies in the base schemas:
* [context/device](https://github.com/adobe/xdm/blob/master/docs/reference/context/device.schema.md): `xdm:typeID` and `xdm:typeIDService`
* [context/search](https://github.com/adobe/xdm/blob/master/docs/reference/context/search.schema.md): `xdm:searchEngineID`
* [context/segmentMembershipItem](https://github.com/adobe/xdm/blob/master/docs/reference/context/segmentmembershipitem.schema.md): `xdm:profileStitchID`, `xdm:segmentID`
* [assets/asset](https://github.com/adobe/xdm/blob/master/docs/reference/assets/asset.schema.md): `xdm:documentID`, `xdm:aliasIDs`
* [data/order](https://github.com/adobe/xdm/blob/master/docs/reference/data/order.schema.md): `xdm:purchaseID`

To sum up, I'd like to understand:
1. Is the `Id` (lower case "d") suffix preferred going forward? Is it also enforced?
2. Is there a plan to make base schemas consistent with this naming convention?

cfraser:
@prabhum2 Please review and merge if it is acceptable.

schemas/context/experienceevent.schema.json
cfraser: The style guide uses `ID` so this field is named correctly.

prabhum2:
Closed this PR as requested by @littlebutty

PR #691, opened 2019-05-30
2019.05.29 Campaign EE Update

eknee:
--update to campaign template mixin to included only the necessary fields
--deletion of the legacy /campaign/experienceevent-all schema
--update to mutationevent schema to drop web object and rename message object


eknee:
@prabhum2 / @cmathis can you review this change.  Before we commit I also need campaign engineering to review.

jwen-adobe:
@eknee you may want to change the property name to something else as it caused validation error.
https://github.com/adobe/xdm/blob/56d3d3279022c25dde9a55c366fe910a113ec22c/extensions/adobe/experience/campaign-experienceevent.schema.json#L220

eknee:
Working on addressing this now.

cmathis:
Removing fields is an issue if you still have ETL or ingestion files that are expecting to use those. I don't know of a way to verify this on all Campaign based Datasets that have recent data.

eknee:
Campaign is reviewing this and will provide approval.  We are tracking this in the following JIRA -> https://jira.corp.adobe.com/browse/PLATIR-4513

extensions/adobe/experience/campaign-experienceevent.schema.json
```diff
@@ -17,91 +17,414 @@
...
   ],
   "definitions": {
     "campaign-experienceevent": {
-      "properties": {}
+      "properties": {
+        "https://ns.adobe.com/experience/campaign/message": {
+          "title": "A single message sent to a recipient.",
+          "type": "object",
+          "properties": {
+            "xdm:id": {
```
aarjain: When we removed experienceevent-all mixin from this, I think we have missed adding profileSnapshot object from there to this mixin. We should include it too.

extensions/adobe/experience/campaign/mutationevent.schema.json
```diff
@@ -19,17 +19,12 @@
...
-          "title": "Web",
-          "$ref": "https://ns.adobe.com/xdm/context/webinfo",
-          "description": "The information related to web page and link of the ExperienceEvent."
-        },
         "xdm:directMarketing": {
           "title": "Direct Marketing",
           "$ref": "https://ns.adobe.com/xdm/context/direct-marketing",
           "description": "The events and properties related to direct/outbound marketing such as email, direct mail, texts and in-app notifications."
         },
-        "https://ns.adobe.com/experience/campaign/message": {
```
aarjain: As far as I know, this should not change the XDM path to which we currently send our data, is this understanding correct?

extensions/adobe/experience/campaign-experienceevent.schema.json
```diff
@@ -17,91 +17,414 @@
...
+              "description": "The recipient of this message is a member of a control group for this campaign activity. A control group is a population which will not receive the delivery; it is used to track post-delivery behavior and campaign impact by making a comparison between the control group (or hold out group) with the behavior of target population, which has received the delivery.",
+              "default": false
+            },
+            "xdm:testMember": {
+              "title": "A/B Test Member",
+              "type": "boolean",
+              "description": "The recipient of this message is a member of a test group for this campaign activity. This is primarily used for identifying the initial population of an A/B Test. After sending different variants to the testMembers, the winner Variant is decided to be sent to the remaining population based on behavioural changes in testMembers.",
+              "default": false
+            },
+            "xdm:size": {
```
saurabhere: we dont use this today. Since the aim is to make the schema compact, we can remote this and re-introduce this whenever we start sending this detail

extensions/adobe/experience/campaign-experienceevent.schema.json
eknee: @aarjain what was the need for profileSnapshot?  I had in my notes this could be removed as its not ever in your payload.  Can you confirm?

extensions/adobe/experience/campaign-experienceevent.schema.json
eknee: @saurabhere what don't you use?  xdm:size isn't used?  Sorry wasn't clear from the comment.

extensions/adobe/experience/campaign/mutationevent.schema.json
eknee: correct this should not change the path.

extensions/adobe/experience/campaign-experienceevent.schema.json
saurabhere: Yes `xdm:size` isn't used today.

extensions/adobe/experience/campaign-experienceevent.schema.json
sawhneyh: It is used for passing the profile fields at the time of creation of experience event for the profile.

extensions/adobe/experience/campaign/mutationevent.schema.json
```diff
@@ -19,17 +19,12 @@
...
+        "xdm:message": {
```
sawhneyh: Is adding a namespace backward compatible.

extensions/adobe/experience/campaign-experienceevent.schema.json
aarjain: @eknee : Adding on to what @sawhneyh said, you can check the presence of these fields in the wiki titled 'Experience Event Payload from Campaign'.

cmathis:
Because it is producing a breaking change we will need this to be in the form of new mixins. We have no way in AEP today to know who else may have used this mixin and if a field is removed we could potentially break their ingestion process. You can rename the old mixins as "deprecated" and give the new mixin the old name.

cmathis:
We need to do this in a way in which the original file stays around (with the title changed to 'Deprecated'), and we then introduce a new file with v2 in the $id path. 

The goal here is not have this appear as a breaking change. The build process cannot see that a file has been removed.

extensions/adobe/experience/campaign-experienceevent-v2.schema.json
```diff
@@ -0,0 +1,172 @@
...
+      "$ref": "https://ns.adobe.com/xdm/common/extensible#/definitions/@context"
+    },
+    {
+      "ref": "https://ns.adobe.com/xdm/experience/campaign/experienceevent-all-v2"
+    },
+    {
+      "$ref": "https://ns.adobe.com/experience/decisioning/proposition"
+    },
+    {
+      "$ref": "#/definitions/campaign-experienceevent-v2"
```
cmathis: The local definition from above is named "campaign-experienceevent". This is what you need to reference here.

cmathis:
Also, lets add "meta:status": "deprecated" to the deprecated files

extensions/adobe/experience/campaign-experienceevent-v2.schema.json
eknee: fixed

eknee:
added "meta:status": "deprecated" to old versions

eknee:
waiting for campaign to approve via jira ticket -> https://jira.corp.adobe.com/browse/PLATIR-4513

extensions/adobe/experience/campaign-experienceevent-v2.schema.json
```diff
@@ -0,0 +1,172 @@
+{
+  "meta:license": [
+    "Copyright 2018 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/experience/campaign-experienceevent-v2",
```
aarjain: The change from our side would be to just update the mixin ref with the new one (containing -v2), is that correct?

eknee: Correct but there are additional changes in the v2 versions.  I modified the mutuation event schema so that you lose the hierarchical mess when trying to reference the message field (see pic)
image

I also dropped the web object within mutation event.

aarjain: Yes got it thanks

cmathis:
@eknee and @aarjain - We are trying to move these solution mixins towards defining their custom fields with no use of the "xdm:" namespace. Here is an alternative PR that does this and will result in the exact same hierarchical XED format that is exposed in AEP. Take and look and let us know if you see a problem.

cmathis:
@saurabhere and @aarjain, we need to know if you are still wanting to go with these changes proposed by Eric? If not, we will close this PR.

PR #693, opened 2019-06-03
Adding additional attributes to support intelligent services Schema

prabhum2:
This is work in progress - DONT MERGE

As a part of intelligent service schema onboarding added more values to channel types .


schemas/context/normalized-experianceevent.json
```diff
@@ -0,0 +1,135 @@
...
+    "https://ns.adobe.com/xdm/context/experienceevent-media",
+    "https://ns.adobe.com/xdm/context/experienceevent-profile-stitch",
+    "https://ns.adobe.com/xdm/context/experienceevent-search",
+    "https://ns.adobe.com/xdm/context/experienceevent-segmentmembership",
+    "https://ns.adobe.com/xdm/context/experienceevent-technical-details",
+    "https://ns.adobe.com/xdm/context/experienceevent-web"],
+  "definitions": {
+    "normalized-experienceevent": {
+      "properties": {
+          "xdm:eventType": {
```
cmathis: A context/experienceevent (which this mixin is meta:intendedToExtend) already include the data/time-series schema which has an eventType field already.

cmathis:
It appears that this normalized-experienceevent is a new mixin that simply wraps all other available standard mixins. Basically takes us back to the original uber experienceevent we had before. 

Who is using this schema, and how will the Dataset it's associated with be populated?

forpankil:
Why aren't we re-using existing mix-ins like "environment","search", "web", "application","commerce","marketing" ?

cmathis:
@prabhum2  - I do not understand what you are doing here. You appear to be duplicating all the existing objects but giving them a new $id that includes "std" in the value. Is that the purpose?

Issue #700, opened 2019-06-24
Update XDM with latest Media Analytics changes


cosminci:
## What are the schemas that are affected by the issue
- `media-timed-asset-reference.schema.json`
- `media-timed-asset-view-details.schema.json`
- `media-timed-qoe.schema.json`
- `media-timed.schema.json`
- `unique-time-played.schema.json` [NEW]

## What are examples of products that are impacted by the issue
- Adobe Media Analytics

Internal JIRA issue: https://jira.corp.adobe.com/browse/VA-2129

High-level changes:
- most of the PR relates to the addition of audio streams and the new properties they come with (e.g. `xmpDM:artist`);
- support for downloaded content; note that **downloaded does not mean offline**, as downloaded content can be played while the device is online;
- support for the new `uniqueTimePlayed` metrics;
- support for error code lists in the QoE schema;

midomir:
## What are the schemas that are affected by the issue
* advertising-timed-asset-reference.schema.json
* media-timed-asset-reference.schema.json
* media-timed-asset-view-details.schema.json
* media-timed-qoe.schema.json
* media-timed.schema.json
* average-minute-audience.schema.json [NEW]
* unique-time-played.schema.json [NEW]
* ad-count.schema.json [NEW]
* chapter-count.schema.json [NEW]

## What are examples of products that are impacted by the issue
* Adobe Media Analytics

Internal JIRA issue: https://jira.corp.adobe.com/browse/VA-2888

High-level changes:
* addition of audio streams and the new properties they come with (e.g. xmpDM:artist);
* addition of ad properites (advertiser, campaign)
* addition of video properties (segment, path)
* support for authorization properties (mvpd, auth)
* support for downloaded content; note that downloaded does not mean offline, as downloaded content can be played while the device is online;
* support for the new uniqueTimePlayed, averageMinutesAudience, adCount, and chapterCount metrics
* support for error code lists in the QoE schema;

lidiaist:
## What are the schemas that are affected by the issue

* media.schema.json

## What are examples of products that are impacted by the issue

*  Adobe Media Analytics
Internal JIRA issue: https://jira.corp.adobe.com/browse/VA-2129

High-level changes:
* support for Streaming Media Parameters (station, label, publisher)

lidiaist:
## What are the schemas that are affected by the issue

* media.schema.json

## What are examples of products that are impacted by the issue

*  Adobe Media Analytics
[...]

PR #701, opened 2019-06-24
[VA-2129] Update XDM with latest Media Analytics changes

cosminci:
Issue #700


cosminci:
Adobe GitHub access request submitted.

PR #703, opened 2019-06-24
[AMO-131288] - Added Schema Definitions for Ad Cloud Search MetaData

jishah-adobe:
Please link to the issue #…

### DIFF
--- a/extensions/adobe/experience/adcloud/searchadvertising/account.example.1.json
+++ b/extensions/adobe/experience/adcloud/searchadvertising/account.example.1.json
@@ -1,0 +1,11 @@
+{
+  "xdm:id": "100001061",
+  "xdm:accountId": "10001234",
+  "xdm:accountName": "My First Account",
+  "xdm:currencyCode": "USD",
+  "xdm:accountStatus": "Enabled",
+  "xdm:budget": 123.45,
+  "xdm:budgetType": "Monthly",
+  "repo:createDate": "2019-04-26T14:00:00+00:00",
+  "repo:modifyDate": "2019-04-26T14:00:00+00:00"
+}
--- a/extensions/adobe/experience/adcloud/searchadvertising/account.schema.json
+++ b/extensions/adobe/experience/adcloud/searchadvertising/account.schema.json
@@ -1,0 +1,91 @@
+{
+  "meta:license": [
+    "Copyright 2019 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/experience/adcloud/searchadvertising/account",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Search Advertising Account",
+  "type": "object",
+  "meta:extensible": true,
+  "meta:abstract": true,
+  "meta:auditable": true,
+  "meta:extends": ["https://ns.adobe.com/xdm/data/record"],
+  "description": "Search Advertising Account defines Customer Configuration on Search Advertising Platform.",
+  "definitions": {
+    "searchadvertising-account": {
+      "properties": {
+        "xdm:accountId": {
+          "title": "Account Identifier",
+          "type": "string",
+          "description": "Identifier that defines Customer/Client ID setup on the Search Advertising Platform."
+        },
+        "xdm:accountName": {
+          "title": "Account Name",
+          "type": "string",
+          "description": "Name of the Account."
+        },
+        "xdm:currencyCode": {
+          "title": "Currency Code",
+          "type": "string",
+          "examples": ["USD", "EUR", "JPY"],
+          "pattern": "^[A-Z]{3}$",
+          "description": "Currency in which this account operates."
+        },
+        "xdm:accountStatus": {
+          "title": "Account Status",
+          "type": "string",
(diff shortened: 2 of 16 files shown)


jishah-adobe:
@ktukker @ktukker-adobe @prabhum2 @cmathis @cfraser

cmathis:
We've gone through these before so there is not a lot to review. 

@ktukker and @cfraser  - This will introduce 8 new global classes that will show up along side the current XDM Profile and XDM ExperienceEvent. These will appear for every tenant. Waiting for your confirmation and approval before we merge this in.

sathyz:
## Campaigns, Accounts
Are we expecting each Adobe Marketing Solutions (Adobe Campaign, CMP, DSP and Search) create their version of campaign and accounts? We should have one generic version of campaign that fits all solutions (like campaign id, name, goal, budget, spend strategy etc). 

In aggregateperformancebykeyword.schema.json, do we need to add attributes like campaignId, keywordId, termId, accountId. when they are already defined in adcloud/campaign.schema.json. 

## Match type
When we worked on publishing experience events for AdCloud (Search) we have added matchType to AdCloud's Campaign, we now have matchType in aggregateperformancebyadbykeyword.schema.json  & aggregateperformancebykeyword.schema.json and we had defined enums all in small cases and here it in capital cases (broad vs Broad). This will create a state of disjoint. Match Types enum should be defined in one place and other schemas should use it same as how we are doing for device.

jishah-adobe:
> ## Campaigns, Accounts
> Are we expecting each Adobe Marketing Solutions (Adobe Campaign, CMP, DSP and Search) create their version of campaign and accounts? We should have one generic version of campaign that fits all solutions (like campaign id, name, goal, budget, spend strategy etc).
> 
> In aggregateperformancebykeyword.schema.json, do we need to add attributes like campaignId, keywordId, termId, accountId. when they are already defined in adcloud/campaign.schema.json.
> 
> ## Match type
> When we worked on publishing experience events for AdCloud (Search) we have added matchType to AdCloud's Campaign, we now have matchType in aggregateperformancebyadbykeyword.schema.json & aggregateperformancebykeyword.schema.json and we had defined enums all in small cases and here it in capital cases (broad vs Broad). This will create a state of disjoint. Match Types enum should be defined in one place and other schemas should use it same as how we are doing for device.

Yes each Adobe solution (Search, DSP, Social) are expected to define their own version of campaign. The definition and semantic of campaign is different for each of these solutions hence different schema. Unfortunately there can be no single version of campaign that fits all solutions.  In fact during the exercise we noticed that search partners (Google, Yahoo, Bing) also have different flavors of campaign definition.  We also started with creating mix-ins and data-types but unfortunately that didn't fit well with XDM semantics.  We were recommended to create class for each entity and define fields within the class for that entity (e.g campaign, ad group and so on).  As far as adcloud/campaign.schema.json is concerned, the schema should not have been called as campaign in first place. It is is just a hierarchy of ID's. 

As far as match types are concerned, I believe it is an attribute of a keyword and should be associated with keywords only.  I agree on part of case sensitivity.  We need to decide on who bears the responsibility of case conversions - Readers of the data (query time) or writers of the data (ingestion time).  On a second thought, I am also rethinking on defining hard enumerations.  I will discuss them with XDM team and adjust schemas based on the guidance.

cmathis:
I have a hard time believing there are no common fields between the different instances of a Campaign no matter what solution it is for. I would expect every Campaign to have an ID, name, and description at least. I created the following PR as an alternative approach to give an example of how this could work  - https://github.com/adobe/xdm/pull/705

This same concept could be used to create a generic Account and solution specific mixins for that as well. 

I will let @ktukker @cfraser  and Srini decide what is the best approach for AEP.

lohariwa-adobe:
> I have a hard time believing there are no common fields between the different instances of a Campaign no matter what solution it is for. I would expect every Campaign to have an ID, name, and description at least. I created the following PR as an alternative approach to give an example of how this could work - #705

@cmathis @ktukker @sathyz: I would say that such a "Campaign" would be a _mixin_ and not a parent class. This is because it won't be a clear real world parent representation of search campaigns and demand-side-platform (DSP) campaigns together. A user of DSP will take it to mean one thing which would not hold true for Search Advertising and vice-versa. I would recommend that we should not attempt to merge these concepts just because they are both under the AdCloud umbrella. If we want to model these concepts in a user/persona Experience driven way, it will be best to represent them clearly in separate namespaces. This will allow marketers and advertisers to look at objects created using such schema in the correct context and make sense of them.

jwen-adobe:
@jishah-adobe need to get Adobe CLA Signed for passing the check.

ktukker-adobe:
After meeting with the AdCloud team, Srinivas and I support this PR. Additional context on the naming of the entities (namespace in the adcloud/searchadvertising):

In the Search Advertising domain:

Campaign: Definition
A set of ad groups (ads, keywords, and bids) that share a budget, location targeting, and other settings. Campaigns are often used to organize categories of products or services that you offer.
•	Advertising Search account can have one or many ad campaigns running.
•	Each campaign consists of one or more ad groups.
•	Settings that you can set at the campaign level include budget, language, location, distribution for the Network, and more.
•	You can create separate ad campaigns to run ads in different locations or using different budgets.

In the Display advertising domain:

Display Placement: Definition
A set of parameters for a specific video ad or set of video ads or  Image ads that define where and when the ads should run. You can create unlimited placements for each campaign. Placement have Ads Units and Budgets and Targeting criteria (Geo, Inventory, Site, Audience

ktukker-adobe:
@kstreeter , @prabhum2 can you merge this PR?

udayeyunni:
@prabhum2 / @kstreeter - Can this PR be merged this week? Do you need anything else from us before you merge the PR?

PR #706, opened 2019-07-08
Meta schema changes

polavishnu4444:
Please link to the issue #…

Allowing for ability to accept type to be multi type as supported in #draft-06 of JSON-LD.


polavishnu4444:
@cdegroot-adobe @cmathis

meta.schema.json
```diff
@@ -112,7 +112,7 @@
       ]
     },
     "type": {
-      "type": "string",
+      "type": ["string", "array"],
```
cdegroot-adobe: There is no way I can approve the use of the type [] semantics. This adds a level of complexity to all users of XDM that requires major consideration and investment. Please propose an alternative implementation using simple types. Reach out to me if you need assistance.

Thanks C.

adipan: I am not really if this change is actually needed. I wonder if the following approach could be taken (which is present in this file itself) -
```
"anyOf": [
        {
          "properties": {
            "meta:extends": {
              "type": "string",
              "format": "uri"
            },
            "allOf": { "$ref": "#/definitions/extending-all-of" }
          },
          "required": ["meta:extends", "allOf"]
        },
        {
          "properties": {
            "meta:extends": {
              "type": "array",
              "minItems": 1,
              "items": {
                "format": "uri",
                "type": "string"
              }
            },
            "allOf": { "$ref": "#/definitions/extending-all-of" }
          },
          "required": ["meta:extends", "allOf"]
        }
```

PR #708, opened 2019-07-10
Segment Definition Schema

ankushkd:
Segment definition schema will help defined segment metadata from profile service and any other external segment producer like Adobe Audience Manager, Adobe Analytics etc.


schemas/context/segmentdefinition-meta-base.schema.json
```diff
@@ -0,0 +1,72 @@
...
+  "title": "Segments basic metadata",
+  "type": "object",
+  "meta:extensible": true,
+  "meta:abstract": true,
+  "meta:intendedToExtend": ["https://ns.adobe.com/xdm/context/segmentdefinition"],
+  "description": "Common fields required for any segment definition",
+  "definitions": {
+    "segment-meta-base": {
+      "properties": {
+        "xdm:entityIdentity": {
```
cfraser: Is this a segment identity or an entity identity? The title and property name would be less confusing if they were aligned.

schemas/context/segmentdefinition-meta-base.schema.json
```diff
@@ -0,0 +1,72 @@
...
+        "xdm:source": {
+          "title": "Source",
+          "type": "string",
+          "description": "Segment Owner. For example AAM, UPS etc"
+        },
+        "xdm:creationTime": {
+          "title": "Creation Time",
+          "type": "string",
+          "format": "date-time",
+          "description": ""
```
cfraser: This should have a proper description, or the property should be omitted.

schemas/context/segmentdefinition-meta-base.schema.json
```diff
@@ -0,0 +1,72 @@
...
+        },
+       "xdm:updateTime": {
+          "title": "Update Time",
+          "type": "string",
+          "format": "date-time",
+          "description": ""
```
cfraser: This should have a proper description, or the property should be omitted.

schemas/context/segmentdefinition-meta-base.schema.json
```diff
@@ -0,0 +1,72 @@
...
+        },
+        "xdm:status": {
+          "type": "string",
+          "title": "Status",
+          "description": "Current status of segment from external system normalized to ACTIVE/INACTIVE",
+          "enum" : [
```
cfraser: The `enum` field is generally used for validation. If you don't wish for the data to be validated explicitly this should move to a `meta:enum`

schemas/context/segmentdefinition-meta-base.schema.json
```diff
@@ -0,0 +1,72 @@
...
+      "properties": {
+        "xdm:entityIdentity": {
+          "title": "Segment Identity",
+          "$ref": "https://ns.adobe.com/xdm/context/segmentidentity",
+          "description": "Identity of the Segment"
+        },
+        "xdm:name": {
+          "title": "Name",
+          "type": "string",
+          "description": "Name of the Segment"
```
cfraser: Should segment be capitalized in this context?

schemas/context/segmentdefinition-meta-base.schema.json
```diff
@@ -0,0 +1,72 @@
+{
+  "meta:license": [
+    "Copyright 2018 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/context/segment-meta-base",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Segments basic metadata",
```
cfraser: Is this metadata or the required fields for a segment? This is the first time I've seen a schema definition described as metadata.

schemas/context/segmentdefinition.schema.json
```diff
@@ -0,0 +1,32 @@
...
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/context/segmentdefinition",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Segment Definition",
+  "type": "object",
+  "meta:extensible": true,
+  "meta:abstract": true,
+  "description": "A Segment Definition schema represents the metadata associated with a segment. For example id, name, description etc",
+  "definitions": {
```
cfraser: Given that the definition block is empty it can be omitted. Of course that begs the question what the purpose of `segmentdefinition` is if it doesn't define additional fields from `segement-meta-base`

schemas/context/segmentdefinition.schema.json
```diff
@@ -0,0 +1,32 @@
...
+    "segmentdefinition": {
+      "properties": { }
+    }
+  },
+  "required": [
```
cfraser: Shouldn't this required block be defined in `segment-meta-base` instead?

schemas/context/segmentdefinition-meta-base.schema.json
ankushkd: Validation here is intended.

schemas/context/segmentdefinition.schema.json
```diff
@@ -0,0 +1,74 @@
...
+  ],
+  "$id": "https://ns.adobe.com/xdm/context/segmentdefinition",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Segment Definition",
+  "type": "object",
+  "meta:extensible": true,
+  "meta:abstract": true,
+  "description": "Segment Definition class to describe a segment from any system",
+  "definitions": {
+    "segmentdefintion": {
```
neerajgoel82: Can you please fix the spelling here?

schemas/context/segmentdefinition.schema.json
```diff
@@ -0,0 +1,74 @@
...
+      "properties": {
```
neerajgoel82: This seems to be lacking some important properties like:
-Segment definition expression
-Info about the ims org/sandbox id for which it is defined
-Version
-Things related to how the segment is evaluated like streaming/batch etc
-ttl
-Merge policy

schemas/context/segmentdefinition-expression.schema.json
```diff
@@ -0,0 +1,54 @@
...
+          "description": "Type of expression. For example PQL",
+          "meta:enum" : [
+            "PQL",
+            "AAM"
+          ]
+        },
+        "xdm:format": {
+          "title": "Format",
+          "type": "string",
+          "description": "Format of expression. For example pql/text"
```
neerajgoel82: Do you think this should also be enum?

schemas/context/segmentdefinition-expression.schema.json
```diff
@@ -0,0 +1,54 @@
...
+  ],
+  "$id": "https://ns.adobe.com/xdm/context/segmentdefinition-expression",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Segment Expression",
+  "type": "object",
+  "meta:extensible": true,
+  "meta:abstract": true,
+  "description": "Segment expression",
+  "definitions": {
+    "segmentdefinition-expression": {
```
neerajgoel82: Can you please check if we follow any convention to have these in camel case? Reason I am asking is that most of the places I have seen these in camel case. But I am not sure if there is a convention followed or is it by chance.

schemas/context/segmentdefinition.schema.json
```diff
@@ -0,0 +1,74 @@
...
+          "title": "Name",
+          "type": "string",
+          "description": "Name of the segment"
+        },
+        "xdm:description": {
+          "title": "Description",
+          "type": "string",
+          "description": "Description of the segment"
+        },
+        "xdm:source": {
```
neerajgoel82: I am not sure what the difference between a namespace in segmentIdentity and source is. Can you please help me understand what this source is for?

schemas/context/segmentdefinition-expression.schema.json
ankushkd: "type" is enough to determine in most cases. Hence left it for specific use cases with values on user's discretion.

schemas/context/segmentdefinition.schema.json
ankushkd: Source is contributing system where as namespace is sub type within the source. Thought segment realization's identity currently has just identity and namespace and no source. That means namespace has to be unique through out all contributing systems. Which makes source redundant. Only benefit it offers is giving information on provider of the segment definition. Having said that, if you think we have/should have other means to relate namespace to source, we can remove source from here.

schemas/context/segmentdefinition.schema.json
ankushkd: These fields will be part of solution sepcific mixin.
That mixin with base class segmentdefinition, will help in creating schema specific to that solution.

schemas/context/segmentdefinition-expression.schema.json
ankushkd: I got inspired from experienceevent, phonenumber, identityitem etc.
<class all lower>-<mixin>

schemas/context/segmentdefinition-expression.example.1.json
```diff
@@ -0,0 +1,6 @@
+{
+    "xdm:type":"PQL",
```
hiteshs: this seems like a completely new type that is Adobe specific? Is there a reason why it is not namespaced?

schemas/context/segmentdefinition-expression.example.1.json
```diff
@@ -0,0 +1,6 @@
...
+    "xdm:format":"pql/text",
```
hiteshs: Namespacing?

schemas/context/segmentdefinition-expression.schema.json
```diff
@@ -0,0 +1,54 @@
...
+      "properties": {
+        "xdm:type": {
+          "title": "Type",
+          "type": "string",
+          "description": "Type of expression. For example PQL",
+          "meta:enum" : [
```
hiteshs: These again seem to be Adobe specific? Should this be in an adobe specific extension?

schemas/context/segmentdefinition.schema.json
hiteshs: Does this explanation need to be put into the field descriptions?

schemas/context/segmentdefinition-expression.schema.json
cfraser: Camelcase is the preferred approach. This is outlined in https://github.com/adobe/xdm/blob/master/CONTRIBUTING.md#coding-styleguides

schemas/context/segmentdefinition-expression.example.1.json
cfraser: XDM fields should have strong semantic meaning. `xdm:type` is already in use in the Experience Event schema. Do you feel your usage aligns with the semantic meaning of the field as it is used in the Experience Event schema?

ankushkd: Hey @cfraser , you mean this - https://github.com/adobe/xdm/blob/master/schemas/data/time-series.schema.json#L31 ? I can change it to expressionType to make it specific to expressions.

schemas/context/segmentdefinition-expression.schema.json
ankushkd: @cfraser  this is not property name and it also be same as URL specifying $id. Should it still be came case?

cfraser: I don't think this definition properties -must- be camel case. But field properties should follow the above guidelines.

ankushkd:
@cfraser  Added expression data type. Please review.

cfraser:
@prabhum2 please review and comment

cfraser:
@ankushkd Please resolve the build failures.

cmathis:
Given this is intended to be the formal object definition used to represent a segment within the Platform, we need to see approval from all parties that will be using it. 

@cluby-adobe

schemas/context/expression.schema.json
```diff
@@ -0,0 +1,46 @@
...
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/context/expression",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Expression",
+  "type": "object",
+  "meta:extensible": true,
+  "meta:abstract": true,
+  "description": "Any expression which can be used for defining a condition.",
```
clubycoder: Since this is the more generic layer I would change the description to talk about an expression for any result type not just a boolean.  Alternatively if you want to restrict it to just boolean expressions I would change the naming to represent that instead of just calling it "expression".

schemas/context/expression.schema.json
```diff
@@ -0,0 +1,46 @@
...
+        "xdm:format": {
+          "title": "Format",
+          "type": "string",
+          "description": "Format of expression. For example pql/text, application/sql",
+          "pattern": "\\w+/[-.\\w]+(?:\\+[-.\\w]+)?"
+        },
+        "xdm:value": {
+          "title": "Value",
+          "type": "string",
+          "description": "Actual expression string. The syntax of expression will depend upon "
```
clubycoder: Looks like part of your description was left out.

rmunix:
What does supporting multiple formats actually means compatibility-wise, is the aim for the segment definitions to be compatible between the different formats? In other words, given the same data set, should I expect to get the same output regardless of the segment definition format?

schemas/context/expression.schema.json
ankushkd: You are right, an expression can have multiple types can be very generic. I am removing this data type as it needs more thought. Moving expression fields into segmentdefinition-expression mixin.

ankushkd:
> @ankushkd , How is the expression linked to the definition?

The link will be defined in solution specific mixin in extension. As some solution/provider might want to submit just definitions and not expressions.

clubycoder:
> > @ankushkd , How is the expression linked to the definition?
> 
> The link will be defined in solution specific mixin in extension. As some solution/provider might want to submit just definitions and not expressions.

To simplify that linkage would it be simples to add the expression as an optional field in the definition?

ankushkd:
> > > @ankushkd , How is the expression linked to the definition?
> > 
> > 
> > The link will be defined in solution specific mixin in extension. As some solution/provider might want to submit just definitions and not expressions.
> 
> To simplify that linkage would it be simples to add the expression as an optional field in the definition?

By keeping expression in a seperate mixin (which can only extend segment definition), I have left control on schema creator if for certain segments we dont want to keep expressions. This is same optional paradigm but enforced via mixins. This will keep the schema size in check.
Also, since additive schema evolution is allowed, it can be added anytime in future from UI.
Example: I am assuming if one system is sharing segment metadata with other systems but it still owns the entity and realizations, it is optional for it to include expressions while sending.

rmunix:
👍 lgtm!

schemas/context/segmentdefinition.schema.json
ktukker-adobe: @ankushkd can youupdated the descriptions with an edited version of this explanation?

cmathis:
For this new segmentdefinition to be treated as a class it needs to extend the data/record behavior.

schemas/context/segmentdefinition.schema.json
ankushkd: @ktukker  Modified. Please review.

ankushkd:
> For this new segmentdefinition to be treated as a class it needs to extend the data/record behavior. Without this, this object will be treated as a datatype within AEP.
> 
> Also, to get the createdDate, etc. fields you can simply extend and include the
> common/auditable object. Look at the Profile class for an example.
> Also @ktukker-adobe and @cfraser this will introduce another class in the UI so be ready for that.

Done

schemas/context/segmentdefinition.schema.json
neerajgoel82: I think at least following fields should be part of  
- Segment definition expression
- Version
- Info about the IMS org/sandbox id for which it is defined

ankushkd: Hey Neeraj, Thanks!

- Segment definition expression is added as mixin and will be added with solution specific extension or schema creation as this is optional for external segments which is owned by other systems.
- Good idea about version, will add it.
- IMS Org Id and Sandbox Id are system level details and not part of schema by default. If they are still required, can be added in solution/system level extensions.

schemas/context/segmentdefinition.schema.json
neerajgoel82: @ankushkd It would be good to think about this in context of information present in segment membership. Reason for the ask is that the mapping from namespace to other contextual information may be required to make sense of segment membership info as well.

ankushkd:
> Approved once the meta:extends attribute is added to the schemadefinition. This should include the $id values of the data/record and common/auditable objects.

Added meta:extends.

PR #717, opened 2019-08-02
[AMO-132268] - Modified title and description for the aggregated perf…

jishah-adobe:
…ormane by Keyword Object

@prabhum2


jishah-adobe:
For CLA certificate - here is the response
David Crompton, a member of the UIM group, has resolved your case and provided the following details: account jkshah12 deprovisioned, provisioned basic access to jishah-adobe, user will receive an invite.

PR #719, opened 2019-08-06
Update meta.schema.json pattern regex

polavishnu4444:
Added `adobecloudplatform` in the `$id` pattern section to be allowed for `user.schema.json`

### DIFF
--- a/meta.schema.json
+++ b/meta.schema.json
@@ -40,7 +40,7 @@
     "$id": {
       "type": "string",
       "format": "uri",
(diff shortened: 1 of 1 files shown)


polavishnu4444:
`$id` used in the file --> `https://github.com/adobe/xdm/blob/master/extensions/ims/user.schema.json#L9` is not in the allowed pattern list of the meta.schema.json

polavishnu4444:
@chrisdegroot - Please find the change required for this schema used in here --> https://git.corp.adobe.com/experience-platform/acp-cs-xdm-dev/pull/31
Please do let us know if this works.

polavishnu4444:
CC: @cmathis

Issue #729, opened 2019-08-24
DecisionEvents must be countable, group-able and aggregate-able by channel


jbeckert:
## What are the schemas that are affected by the issue
https://ns.adobe.com/experience/decisioning/decisionevent ->
https://ns.adobe.com/experience/decisioning/proposition-details ->
https://ns.adobe.com/experience/decisioning/proposition-detail

## What are examples of products that are impacted by the issue
Platform Decisioning, Campaign

DecisionEvents must be countable, group-able and aggregate-able by channel.
The channel is related to the activity and option selections and therefore the logical place for recording the channel context for the decision events is proposition detail.
Note that there can be multiple propositions made with a single decision and those propositions don't have to be for the same channel. E.g. fatigue management decisions will span multiple channels.

PR #730, opened 2019-08-24
DecisionEvents must be countable, group-able and aggregate-able by channel

jbeckert:
Issue https://github.com/adobe/xdm/issues/729

The channel is related to the activity and option selections and therefore the logical place for recording the channel context for the decision events is proposition detail.
Note that there can be multiple propositions made with a single decision and those propositions don't have to be for the same channel. E.g. fatigue management decisions will span multiple channels.


jbeckert:
Two example files failed validation with ajv. The identity items in the xdm:identityMap must have namespaces properties.

jbeckert:
@cmathis Could you please merge this PR? We depend on it for the 19.9 iterations.
Thank you!

PR #732, opened 2019-08-28
Experience decisioning split model domain into decisioning and offer management

jbeckert:
This PR mainly changes the way offer management schemas are composed.
A number of class and datatype schemas are used to assemble the offer management schemas.
A few new concepts to describe interfaces of decision rules and function invocations were added.
xdm-expression, xdm-function-context-definition, xdm-function-invocation


jbeckert:
The incompatibility-check fails due to data type schema files moved to another location but the effective class and data set schemas constructed with them remain generally the same. I.e. the resulting property & property type model is the same.

prabhum2:
@jbeckert is this PR still needed ?

jbeckert:
@prabhum2 Yes, I'm taking it up again and would like you to put yourself as a reviewer on this. I will fix the issues with circle and make a few additions

Issue #734, opened 2019-08-29
Add Schema Definitions for Ad Cloud DSP Advertising Feed related MetaData


raileanu-adobe:
What are the schemas that are affected by the issue
We add 3 new classes for the AdCloud DSP Metadata , related to an advertising Feed:

Feed
Ad Server
Media Property

What are examples of products that are impacted by the issue
Adobe Advertising Cloud

raileanu-adobe:
Closing - Feed schema is no longer needed.

PR #736, opened 2019-08-29
DSP Feed-related Metadata Hierarchy

raileanu-adobe:
Please link to the issue #734 

This will introduce 3 new global classes : Feed, Ad Server and Media Property. These will model the support for feed-based advertising.
Inventory (bid requests, auctions, ad calls) are the means by with we buy media and are the key input to the DSP. Bid requests are fed to via media placements (feeds).  

Each feed belongs to a media property, which tries to represent a) the media owner (publisher) and b) the entity that will bill us for the media (service provider, which often maps to the SSP).
Additionally each feed has an ad server, which indicates the technology used to convey the feed.  Often this maps to the SSP.

A more extended documentation regarding the part of the DSP Advertising model we want to expose can be found in our [wiki page](https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=EfficientFrontier&title=%5BXDM+Mapping%5D+Feed-related+hierarchy).

@prabhum2 , @cfraser Can you please help us with the review?

### DIFF
--- a/extensions/adobe/experience/adcloud/dsp/adserver.example.1.json
+++ b/extensions/adobe/experience/adcloud/dsp/adserver.example.1.json
@@ -1,0 +1,8 @@
+{
+  "xdm:id": "12",
+  "xdm:adServerName": "LiveRail",
+  "xdm:adServerStatus": "Active",
+  "xdm:dealIdSupported": true,
+  "xdm:partnerId": 567,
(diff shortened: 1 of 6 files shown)


raileanu-adobe:
This is no longer needed for Cross Channel Reporting or AAM integration.

PR #745, opened 2019-09-06
added consumer-expevent consolidated mixin

prabhum2:
Please link to the issue #…


extensions/adobe/experience/consumer-experienceevent.schema.json
```diff
@@ -0,0 +1,70 @@
...
+      "$ref": "https://ns.adobe.com/xdm/context/experienceevent-segmentmembership"
+    },
+    {
+      "$ref": "https://ns.adobe.com/xdm/context/experienceevent-technical-details"
+    },
+    {
+      "$ref": "https://ns.adobe.com/xdm/context/experienceevent-web"
+    },
+    {
+      "$ref": "https://ns.adobe.com/experience/analytics/commerce"
```
cmathis: Should this be the standard commerce and not the analytics form?

prabhum2: sorry my bad , i overlooked

cmathis:
@Khanna-vikas and @smcodes, 

This is the PR which introduces a new Mixin that is used to build an ExperienceEvent that includes all of the standard mixins/fields minus any of the solution specific fields. This is intended to be the "Consumer" mixin, or some may say this is the mixin you use to generate the normalized "Adobe Experience Event".

PR #749, opened 2019-09-07
Segment definition and Segment folder to have identity in identity map

ankushkd:
New Audience Manager extension specific segment folder schema added.
Segment definition and segment folder will use identityMap to pass primary identity.


cmathis:
On the new SegmentFolder object, is the segmentParentFolderIdentity field expected to be processed by UPS or would they only care about the identityMap value(s)?

ankushkd:
@cmathis  this will only work with identityMap and need not work with segmentParentFolderIdentity.

cmathis:
@ankushkd  - If the Identity Service makes a change where they automatically consume any field of type "context/identity", will that be a problem? I don't think that logic is in place today, but may be part of their future roadmap.

ankushkd:
@cmathis  Thats true. How can we handle this? We need folder identity here to tell which folder this segment belongs to.

cmathis:
Is this "folder identity" associated with the core identity service at all?  The current context/identity is intended to represent an identity from that core service. If they are not associated, we should look into representing the folder identity in a different way.

ankushkd:
no, its not to be used by identity service for this schema, its just reference to identity in identity map of segment folder schema. Since its a reference, structure of id and namespace required.

cmathis:
I hate to see up duplicating too much, but we may be safer to have a new context/identityReference object which can be used for this. It can be much smaller that the current context/identity and only those fields needed in this case.

The description would be something along the lines of: "Identity information that is not intended to be the identity of **this** schema, but used to reference the identity of another schema.

cc @prabhum2 and Vikas to agree or disagree with this approach.

ankushkd:
@cmathis  Is using Segment Identity https://github.com/adobe/xdm/blob/master/schemas/context/segmentidentity.schema.json an option?

cmathis:
@ankushkd  - Yes, that should work. Anything that the Identity Service would not be watching for and treating as the identity.

cmathis:
As part of the PR we are introducing a new context/identityReference data type to avoid any confusion with the current context/identity. This new data type is used when you want the field to reference the identity field of a foreign schema/object. This along with the existing relationshipIdentity descriptor would allow for the following within XDM.

1. A field can be defined as a string and use the referenceIdentity descriptor to assign the associated foreign namespace 
2. A field can be defined as an identityReference where the namespace is supplied in the ingested data

ktukker:
Hello Craig, team, The PR as provided with "context/referenceIdentity" is fine by me. We should clean this up in future by looking at this holistically.

cmathis:
retest

cmathis:
Have updated the PR to introduce a new boolean refefenceIdentity attribute on the context/identity datatype. This attribute is set to true when the field is referring to the identity of a foreign entity, and should **NOT** be be consumed by Identity Service as one of its own identifiers. 

cc Vikas and Srini

cmathis:
@khanna-vikas and @smcodes please take a look at this and approve. Was decided that because UIS will not process non-person identities, the introduction of a new flag is not necessary at the moment. This will give UIS some time to process this proposal of new flag, or come up with a better alternative.

PR #750, opened 2019-09-10
[AMO-133102] - Updated namespace for search advertising campaign object

jishah-adobe:
Please link to the issue #…


jishah-adobe:
@prabhum2 @cmathis

extensions/adobe/experience/adcloud/searchads/campaign.example.1.json
```diff
@@ -0,0 +1,14 @@
+{
+  "_adcloud.searchads:id": "100001061",
+  "_adcloud.searchads:campaignId": "10001234",
```
cmathis: This should be "searchads:campaignId": "10001234". 
Use the exact field name from the XDM schema here.

jishah-adobe: Fixed example for all fields

extensions/adobe/experience/adcloud/searchads/campaign.example.1.json
```diff
@@ -0,0 +1,14 @@
+{
+  "_adcloud.searchads:id": "100001061",
```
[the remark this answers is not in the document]
jishah-adobe: Fixed

extensions/adobe/experience/adcloud/searchads/campaign.schema.json
```diff
@@ -0,0 +1,166 @@
...
+          "title": "Campaign Name",
+          "type": "string",
+          "description": "Name of the Campaign."
+        },
+        "searchads:campaignDescription": {
+          "title": "Description",
+          "type": "string",
+          "description": "Description of the Campaign."
+        },
+        "searchads:campaignStatus": {
```
sathyz: A keyword could be disapproved[1], but a campaign isn't [2].

[1] https://support.google.com/google-ads/answer/2453978?hl=en 
[2] https://support.google.com/google-ads/answer/2549115?hl=en

extensions/adobe/experience/adcloud/searchads/campaign.schema.json
```diff
@@ -0,0 +1,166 @@
...
+            "On Hold (Other)": "On Hold (Other)",
+            "Duplicate": "Duplicate",
+            "Orphan": "Orphan",
+            "Optimize": "Optimize",
+            "Disabled": "Disabled",
+            "Unknown": "Unknown",
+            "Dummy": "Dummy"
+          }
+        },
+        "searchads:channelType": {
```
sathyz: What are possible channel types?

extensions/adobe/experience/adcloud/searchads/campaign.schema.json
```diff
@@ -0,0 +1,166 @@
...
+          ],
+          "meta:enum": {
+            "daily": "Daily",
+            "Monthly": "Monthly",
+            "Lifetime": "Lifetime",
+            "Daily Spend Until Depleted": "Daily spend until depleted",
+            "Day of week": "Day of week"
+          }
+        },
+        "searchads:adDeliveryType": {
```
sathyz: If there are only two states of adDelivery, should this be boolean?

extensions/adobe/experience/adcloud/searchads/campaign.schema.json
```diff
@@ -0,0 +1,166 @@
...
+          "description": "Campaign Objective indicates the primary objective of the campaign - e.g. Promotions, Brands, App Installs.",
+          "enum": ["visit_web", "promote_brand", "install_app", "visit_offer"],
+          "meta:enum": {
+            "visit_web": "visit_web",
+            "promote_brand": "promote_brand",
+            "install_app": "install_app",
+            "visit_offer": "visit_offer"
+          }
+        },
+        "searchads:campaignStartDate": {
```
sathyz: Should we have generic startDate and endDate instead of campaignStartDate and campaignEndDate?

extensions/adobe/experience/adcloud/searchads/campaign.schema.json
jishah-adobe: (1) - Accounts missing in campaign seems to be a good catch. I missed that earlier. Thanks for catching
(2) - Budget > AUD 150 is not a query criteria for reporting use cases, fields can be concatenated
(3) xdm: searchEngineId has a different semantics that our case, for us both searchEngine and SearchEngineId is actually the advertising platform name (google adwords, MSN adcenter and so on), nothing to do with the search.
(4) Campaign Description is actually a field in user db.

extensions/adobe/experience/adcloud/searchads/campaign.schema.json
jishah-adobe: Boolean doesn't signify anything. I would rather put it as a string. For now enum should suffice.

extensions/adobe/experience/adcloud/searchads/campaign.schema.json
jishah-adobe: Create a generic object startDate and endDate and share it across schemas that is what you meant?  If so, the idea here was to define a complete object in its entirety rather than adding mix-ins or data-types. If that was not the case let me know what your thoughts were

extensions/adobe/experience/adcloud/searchads/campaign.schema.json
jishah-adobe: As of now I see only search as a channel type. The definition of channel type is heavily overloaded

jishah-adobe:
> @jishah-adobe Just verify whether the date is sufficient for campaign start date and end date. OR a better format date-time is required.

The fields are stored in following format in our OLTP system - YYYY-MM-DD so it should suffice for us

khanna-vikas:
LGTM

PR #754, opened 2019-09-12
Define Search Advertising Metadata, Performance and Attributed Model Conversions Schema

jishah-adobe:
Please link to the issue #…
@prabhum2 @cmathis


extensions/adobe/experience/adcloud/searchads/adgroup.schema.json
```diff
@@ -5,39 +5,39 @@
...
-  "meta:extensible": true,
-  "meta:abstract": true,
+  "meta:extensible": false,
+  "meta:abstract": false,
   "meta:auditable": true,
   "meta:extends": ["https://ns.adobe.com/xdm/data/record"],
   "description": "Search Advertising Ad Group is a logical collection or grouping of ads.",
   "definitions": {
-    "searchadvertising-adgroup": {
+    "searchads-adgroup": {
```
prabhum2: @jishah-adobe can you please justify the reasoning for adding adgroup name and id under searchad namespace Vs adcloud name space so that these becomes reusable between searchad and dsp

jishah-adobe: DSP and Search have semantically different meanings thought entity/schema names may be alike.  In this specific case Ad Group is relevant to Search Advertising only hence searchads. Entity Ad Group would not be defined for dsp

extensions/adobe/experience/adcloud/searchads/adgroup.schema.json
```diff
@@ -5,39 +5,39 @@
...
           "description": "Ad Group ID on the search advertising platform."
         },
-        "xdm:campaignId": {
+        "searchads:campaignId": {
           "title": "Campaign Identifier",
           "type": "string",
           "description": "Campaign ID for the Ad Group on the search advertising platform."
         },
-        "xdm:internalCampaignId": {
+        "searchads:internalCampaignId": {
```
sathyz: What is the purpose of exposing `internalCampaignId`. None of the performance report uses `internalCampaignId` and neither the https://ns.adobe.com/xdm/adcloud/searchads/campaign schema has the `internalCampaignId`.

extensions/adobe/experience/adcloud/searchads/adgroup.schema.json
```diff
@@ -80,7 +80,7 @@
             "Dummy": "Dummy"
```
sathyz: What does status `Dummy` mean?

extensions/adobe/experience/adcloud/searchads/adgroup.schema.json
jishah-adobe: The schemas are not defined only for reporting use cases. They are meant to be consumed by other data sources or internal solutions team.  As far as campaign is concerned we store internalCampaignId in the id field of the record object.  We are also storing internal ad group id in the id field.

extensions/adobe/experience/adcloud/searchads/account.example.1.json
```diff
@@ -0,0 +1,11 @@
+{
+  "@id": "100001061",
+  "searchads:accountId": "10001234",
+  "searchads:accountName": "My First Account",
+  "searchads:currencyCode": "USD",
+  "searchads:accountStatus": "Enabled",
+  "searchads:budget": 123.45,
+  "searchads:budgetType": "Monthly",
+  "repo:createDate": "2019-04-26T14:00:00+00:00",
+  "repo:modifyDate": "2019-04-26T14:00:00+00:00"
```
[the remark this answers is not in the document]
jishah-adobe: Example will be added

extensions/adobe/experience/adcloud/searchads/adgroup.schema.json
jishah-adobe: This is one of the possible status defined in the source database (PG).  It is possible that none of the ad-groups may have that status but application logic allows to have that status.

extensions/adobe/experience/adcloud/searchads/aggregateperformancebyad.schema.json
```diff
@@ -6,66 +6,66 @@
...
           "$ref": "https://ns.adobe.com/xdm/data/measure",
           "description": "Top of the Page Bid Estimator. The measure is computed by the network and is not an additive measure."
         },
-        "xdm:qualityScore": {
+        "searchads:qualityScore": {
           "$ref": "https://ns.adobe.com/xdm/data/measure",
           "description": "Quality Score of the Keyword assigned by the network - range 1 to 10. This is not an additive measure."
         },
-        "xdm:device": {
+        "searchads:device": {
```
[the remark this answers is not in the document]
jishah-adobe: Can you share where is xdm:device defined in xdm namespace?  I see device schema and field underneath that schema is called type. Ideally the field should be called deviceType if we want to namespace this in xdm. The same holds true for all other measure fields such as impressions, clicks and so on.  I think we better keep all these fields under searchads namespace so we can ajdust them as per our needs in future if required

extensions/adobe/experience/adcloud/searchads/account.schema.json
```diff
@@ -5,36 +5,36 @@
...
           "description": "Identifier that defines Customer/Client ID setup on the Search Advertising Platform."
         },
-        "xdm:accountName": {
+        "searchads:accountName": {
           "title": "Account Name",
           "type": "string",
           "description": "Name of the Account."
         },
-        "xdm:currencyCode": {
+        "searchads:currencyCode": {
```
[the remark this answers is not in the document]
jishah-adobe: As mentioned earlier there will be similar fields defined under xdm namespace by other solutions -> e.g campaign or campaignId will be used by many solutions but meaning may be different across solutions.  Hence, we left all our fields under searchads namespace.

schemas/context/device.schema.json
```diff
@@ -44,7 +44,9 @@
             "gaming": "Gaming Console",
             "television": "Television",
             "settop": "Set-top Box",
-            "mediaplayer": "Media Player"
+            "mediaplayer": "Media Player",
+            "computers": "Computers",
+            "tv screens": "TV Screens"
```
cmathis: @khanna-vikas - Note this field has a meta:enum with no enum so technically it has not constrained values. Also note that value on "tv screens" has a space. This may be okay, but just something to be aware of.

khanna-vikas: meta:enums are fine, unless they are spelled wrongly.

PR #764, opened 2019-09-27
Alternative Campaign Mixin

cmathis:
Provide an alternative to PR https://github.com/adobe/xdm/pull/691

The schemas in this PR use full URI's for the custom campaign specific fields. The XED that is exposed in AEP comes out the same. This allows the campaign extensions to not define fields using the "xdm" namespace.


cmathis:
@aarjain  - This is the PR you should look at instead of PR #691

cmathis:
This PR is on hold until Campaign gets back with an approval.

cmathis:
@aarjain  - You are right, feel free to take over ownership of this branch and mark any of those deprecations that you wanted. I don't know what you would still use or expect this to replace.

PR #778, opened 2019-10-09
Adding Schemas and Example Json files for graph data

mjasrotia:
Please link to the issue #…


schemas/context/graph-edge.schema.json
```diff
@@ -0,0 +1,127 @@
...
+    ],
+    "$id": "https://ns.adobe.com/xdm/context/graph-edge",
+	"$schema": "http://json-schema.org/draft-06/schema#",
+	"title": "XDM Graph Edge",
+	"type": "object",
+	"description": "XDM Graph Schema for Edge/Relationship between two nodes/vertices.",
+	"auditable": true,
+	"meta:extensible": true,
+	"meta:abstract": true,
+	"meta:intendedToExtend": ["https://ns.adobe.com/xdm/context/profile"],
```
cmathis: Is this correct, that these objects are representations of a Profile? 

Or, should this become it's own type of class?

schemas/context/graph-node.schema.json
```diff
@@ -0,0 +1,83 @@
...
+      "of the License at https://creativecommons.org/licenses/by/4.0/"
+    ],
+    "$id": "https://ns.adobe.com/xdm/context/graph-node",
+	"$schema": "http://json-schema.org/draft-06/schema#",
+	"title": "XDM Graph Node/Vertex Schema",
+	"type": "object",
+	"description": "XDM Schema for a graph Node/Vertex.",
+	"meta:extensible": true,
+	"meta:abstract": true,
+	"meta:intendedToExtend": ["https://ns.adobe.com/xdm/context/profile"],
```
cmathis: Is this correct, that these objects are representations of a Profile?

Or, should this become it's own type of class?

cmathis:
@mjasrotia - Am I reading this correctly that these graph concepts (nodes, edges, cluster) are all intended to extend a Profile class (aka Profile schema). Or should these graph components be their own type of business entities (their own classes).

schemas/context/graph-edge.schema.json
mjasrotia: @cmathis , Initially the intent was to have the graphs exist as its own class and have the Node and Edge mixins extend this, we later made this extend profile class assuming that it would be used along with the profile data. We will get more clarity once we have feedback from the Identity services team.

schemas/context/graph-node.schema.json
mjasrotia: @cmathis , Initially the intent was to have the graphs exist as its own class and have the Node and Edge mixins extend this, we later made this extend profile class assuming that it would be used along with the profile data. We will get more clarity once we have feedback from the Identity services team.

mjasrotia:
Creating another pull request for the XDM Graph Schemas.

PR #792, opened 2019-11-07
#791 - Add environment field to implementationDetails

jonsnyder:
#791


schemas/context/implementationdetails.schema.json
```diff
@@ -34,6 +34,18 @@
           "title": "Version",
           "type": "string",
           "description": "The version identifier of the API, e.g h.18."
+        },
+        "xdm:environment": {
+          "title": "Environment",
+          "type": "string",
+          "description": "The environment of the implementation",
+          "enum": ["web", "mobile", "server", "iot"],
```
dancingcactus: I wonder if we should go 'browser', 'app', 'server', 'iot'. 

It takes some of the ambiguity out since web could be on mobile or desktop. And App could be more than just mobile. It could be a native app windows or an app on Roku. 

I am on the fence for iOT since most of the time is is just an app (like a set-top box). Basically I think there is a place for a device sending data as part of the operating system (and not part of the app) which is why I think there is a place for it.

jonsnyder: I'll make the change.

jfkhoury:
@cmathis @khanna-vikas Can we please get reviews on this PR? Thanks!

schemas/context/implementationdetails.schema.json
```diff
@@ -33,7 +33,19 @@
         "xdm:version": {
           "title": "Version",
           "type": "string",
-          "description": "The version identifier of the API, for example, 'h.18'."
+          "description": "The version identifier of the API, e.g h.18."
+        },
+        "xdm:environment": {
```
emdobrin: How is this addition different than https://github.com/adobe/xdm/blob/master/schemas/context/environment.schema.json ?

jonsnyder: `implementationDetails.environment` is where the event was captured, but `environment.type` is where the event occurred.  For example, for an event that originated from a page load on a website, `environment.type=browser`.  If this event was added to the xdm dataset via a server to server call, `implementationDetails.environment=server`.

prabhum2:
@jonsnyder is this PR still needed ?

jfkhoury:
Hi @prabhum2 ,

We still need this PR, thank you so much for reaching out. Can we get it merged?

cc: @jonsnyder @dancingcactus

jfkhoury:
Thank you @prabhum2 . Any idea when it will be released?

jfkhoury:
Hi @prabhum2 Any idea when this PR will be released in Prod? Thank you.

prabhum2:
it should be in prod already . Please let me know if you do not see

PR #795, opened 2019-11-07
Support for record timestamp based profile overrides in a dataset

patnaiku:
Please link to the issue #…


patnaiku:
retest

cmathis:
FYI - the existing audit fields today are not automatically being populated by the AEP ingestion like everyone expects. I would first touch based with the AEP architects and make sure they agree that this was an oversight and something they will implement. If they all come back and say it will **not** automatically be populated and that the existing fields are intended to represent the "source", then there is not reason for another field to be added.

prabhum2:
@patnaiku - is this PR still needed ?

PR #797, opened 2019-11-07
Update multiple class and mixin description strings.

djwendo:
Update "description" fields for multiple Classes and Mixins.


schemas/context/segmentdefinition.schema.json
```diff
@@ -11,7 +11,7 @@
   "type": "object",
   "meta:extensible": true,
   "meta:abstract": true,
-  "description": "Segment Definition class to describe a segment from any system",
+  "description": "This is a base class for a segment definition which includes required fields like identity of a segment and name of segment. Segment definitions coming from various system can use this class to build Segment Definition XDM schema.",
```
cfraser: `a base class` or `the base class`?

amy-durocher:
These strings are great!  Approved on my side!

schemas/context/segmentdefinition.schema.json
djwendo: "a" was provided, but believe "the" would be more accurate as there is no other base class.

amy-durocher:
We are ready to push out changes - no more changes or PRs expected

PR #802, opened 2019-11-15
Added meta flag to auditable in profile schema.

mjasrotia:
Fixed the meta:auditable flag in the profile schema.


prabhum2:
not needed

PR #807, opened 2019-11-28
[AdCloud] Change adClass type from enum to string

larisaad:
This is an addendum to the DSP name spaced schemas. 

- Switch data type of field `adClass` (`advertisement` schema) from enum to string.


udayeyunni:
@cmathis / @prabhum2  /  @mjasrotia  - PR is ready for your review. Thx

cmathis:
There is a potential for other users out there also using this advertisement schema who are expecting the enum. We don't really have a way to know if that's the case and to get their feedback.

We can do one of two things:
1. We mark the existing dsp:adClass as deprecated and create a new dsp:adClassification (or something like that) allowing the current dsp:adClass to stay around as is for current users.
2. You reach out to existing XDM users and find out who may be using this schema and make sure they are okay with the change. If no problems are found then we go ahead with it.

larisaad:
@cmathis we are the only ones using that schema. The schema was created by the AdCloud DSP team under the DSP namespace. Unfortunately, we made a wrong assumption when creating the adClass field and we now want to make it less restrictive.

PR #810, opened 2019-12-06
[AMO-134906] - Create Schema for AMO Search Ads

danisbet-adobe:
Please link to the issue #…


danisbet-adobe:
Thanks for reviewing. After review on requirements from product, we're pausing development on ingestion of this data until a later date. Thank you for your patience.

PR #817, opened 2019-12-17
Fields for capturing region information for Experience Events and Profiles

ankushkd:
Region information field to help capture the region where experience event or profile was seen. Timestamp of capture on that particular region can also be defined using this.
This includes addition of a new datatype - regioninfo which is being used in Experience event technical details mixin and profiles technical details mixin (newly added)


rkpandey:
I suggest using map in profile object if intention is to capture last timestamp for a region. As array will keep on growing.

harleensahni:
Looks good to me

extensions/adobe/experience/experienceevent-edgeregion.schema.json
```diff
@@ -0,0 +1,192 @@
...
+  "description": "",
+  "definitions": {
+    "experienceevent-edgeregion": {
+      "properties": {
+        "https://ns.adobe.com/experience/userActivityRegion": {
+          "title": "User activity region",
+          "description": "Region info where event is captured.",
+          "type": "object",
+          "properties": {
+            "https://ns.adobe.com/experience/dataCenterLocation": {
```
khanna-vikas: @ankushkd you can avoid unnecessary nesting **_experience**.dataCentreLocation by the XED converter, by just naming the property by "dataCentreLocation" instead of the URI. Check if you want to make this change?

extensions/adobe/experience/experienceevent-edgeregion.schema.json
```diff
@@ -0,0 +1,192 @@
...
+                "VA8",
+                "VA9",
+                "WA1",
+                "WA2",
+                "WY1",
+                "ZAF1",
+                "ZAF2"
+              ]
+            },
+            "https://ns.adobe.com/experience/captureTimestamp": {
```
khanna-vikas: @ankushkd Same here "captureTimestamp" instead of the URI.

extensions/adobe/experience/aam-experienceevent.example.1.json
```diff
@@ -22,6 +22,10 @@
       }
     ]
   },
+  "xdm:userActivityRegion": {
+    "xdm:dataCenterLocation": "VA6",
+    "xdm:captureTimestamp": "2017-09-26T15:52:20+00:00"
```
rkpandey: Why this timestamp will be different from event timestamp?

ankushkd: there can be a difference between when action actually happened and when it was captured on edge.

shagreel: A prime example is mobile hits. It is possible to send events to the system in batch (mobile phone comes online again) where the event timestamps might be in the past.

PR #818, opened 2019-12-17
B2B Account Schemas

mjasrotia:
Adding Schemas for following:
1. account (Class) - Have immutable attributes of an account.
2. account-sf (MIXIN) - Have attributes specific to Sales Force CRM system.
3. account-crm (MIXIN) - Have common account standard attributes.
4. account-sync (MIXIN) - Have sync attributes


schemas/context/account-marketo.schema.json
```diff
@@ -0,0 +1,184 @@
...
+  ],
+  "$id": "https://ns.adobe.com/xdm/context/account-marketo",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Marketo Account",
+  "type": "object",
+  "meta:extensible": true,
+  "meta:abstract": true,
+  "meta:intendedToExtend": [
+    "https://ns.adobe.com/xdm/context/account",
+    "https://ns.adobe.com/xdm/context/account-sf"
```
cmathis: https://ns.adobe.com/xdm/context/account-sf is not a class. This is intended to be an array of classes that this mixin can be added to

mjasrotia: The account-marketo MIXIN needs to extend the account-sf mixin. Can we just put the mixin in the intendtoExtend array ? It should automatically be added to the account class as well.

mjasrotia: Added the account-sf in the allOf section of account-marketo and also added the "meta:extends" in the account-marketo mixin. 
Removed the account-sf reference from the "meta:intendToExtend"

mjasrotia:
Creating another PR

PR #822, opened 2020-01-17
Update experienceevent-edgeregion.schema.json

mjasrotia:
Fixing the Meta Enums for this Schema - It needs to be an object with a key:value pairs and not an array of strings

Please link to the issue #…


cfraser:
LGTM 👍

PR #830, opened 2020-02-18
Delivery Feedback Events XDM

sauagarw:
Please link to the issue #…


extensions/adobe/experience/campaign/feedbackevent.schema.json
```diff
@@ -0,0 +1,244 @@
...
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Adobe Delivery Feedback Events",
+  "type": "object",
+  "meta:extensible": true,
+  "meta:intendedToExtend": ["https://ns.adobe.com/xdm/context/experienceevent"],
+  "description": "Contains all delivery event fields.",
+  "definitions": {
+    "feedbackevent": {
+      "properties": {
+        "xdm:feedbackType": {
```
mjasrotia: Should this be "The type of feedback" ? The event type is already part of the "time-series" schema which is extended by "experienceevent" schema.

extensions/adobe/experience/campaign/feedbackevent.schema.json
```diff
@@ -0,0 +1,244 @@
...
+            "xdm:size": {
+              "title": "Size (Bytes)",
+              "type": "integer",
+              "description": "Size in bytes of the message.\n* For email, the `size` reflects the all MIME envelope of the message, including the encoded attachments if any.\n* For SMS, the `size` reflects the number of bytes necessary to deliver the message.\nIf that number is over 160 the message will be delivered over more than one SMS.",
+              "minimum": 0,
+              "maximum": 2147483647
+            },
+            "https://ns.adobe.com/experience/campaign/recipient": {
+              "title": "The recipient of this message",
+              "type": "object",
```
mjasrotia: Should this be an array as the same message could have multiple recipients for to/CC/BCC at the same time.

extensions/adobe/experience/campaign/feedbackevent.schema.json
```diff
@@ -0,0 +1,244 @@
...
+          "title": "Delivery channel",
+          "type": "object",
+          "properties": {
+            "xdm:channelType": {
+              "title": "Channel Type",
+              "type": "string",
+              "description": "Classifies whether the channel is email, push, sms"
+            },
+            "https://ns.adobe.com/experience/campaign/email": {
+              "title": "E-mail channel",
```
mjasrotia: Capitalize "Channel"

extensions/adobe/experience/campaign/feedbackevent.schema.json
```diff
@@ -0,0 +1,244 @@
...
+              "type": "object",
+              "properties": {
+                "https://ns.adobe.com/experience/campaign/interactions": {
+                  "title": "User interations",
```
mjasrotia: Capitalize "Interactions"

extensions/adobe/experience/campaign/feedbackevent.schema.json
```diff
@@ -0,0 +1,244 @@
...
+            "https://ns.adobe.com/experience/campaign/recipient": {
+              "title": "The recipient of this message",
+              "type": "object",
+              "properties": {
+                "xdm:to": {
+                  "title": "To",
+                  "type": "string",
+                  "description": "Address used as a to/callee address. Depending on the communication channel, the `address` has a different format.\n\n  * `email`: an email address.\n  * `textMessage` or `phone`: a shortcode or phone number."
+                },
+                "xdm:Type": {
```
mjasrotia: lower case "type"

extensions/adobe/experience/campaign/feedbackevent.schema.json
```diff
@@ -0,0 +1,244 @@
...
+  "meta:extensible": true,
+  "meta:intendedToExtend": ["https://ns.adobe.com/xdm/context/experienceevent"],
+  "description": "Contains all delivery event fields.",
+  "definitions": {
+    "feedbackevent": {
+      "properties": {
+        "xdm:feedbackType": {
+          "title": "Type",
+          "type": "string",
+          "description": "The type of message event"
```
mjasrotia: Pleaese add (.) to end of statement

extensions/adobe/experience/campaign/feedbackevent.schema.json
```diff
@@ -0,0 +1,244 @@
...
+        },
+        "xdm:retryCount": {
+          "title": "retryCount",
+          "type": "integer",
+          "description": "Number of failed attempts before the current event"
```
mjasrotia: Pleaese add (.) to end of statement

extensions/adobe/experience/campaign/feedbackevent.schema.json
```diff
@@ -0,0 +1,244 @@
...
+              "properties": {
+                "xdm:to": {
+                  "title": "To",
+                  "type": "string",
+                  "description": "Address used as a to/callee address. Depending on the communication channel, the `address` has a different format.\n\n  * `email`: an email address.\n  * `textMessage` or `phone`: a shortcode or phone number."
+                },
+                "xdm:Type": {
+                  "title": "Type",
+                  "type": "string",
+                  "description": "Whether this message message event describes the BCC, CC or To recipient"
```
mjasrotia: Pleaese add (.) to end of statement

extensions/adobe/experience/campaign/feedbackevent.schema.json
```diff
@@ -0,0 +1,244 @@
...
+            },
+            "xdm:errorCode": {
+              "title": "Error Code",
+              "type": "string",
+              "description": "Error code by which the remote server described a failed delivery attempt."
+            },
+            "xdm:bounceType": {
+              "title": "Bounce Type",
+              "type": "string",
+              "description": "Classifies whether it is a soft bounce or hard bounce"
```
mjasrotia: Pleaese add (.) to end of statement

extensions/adobe/experience/campaign/feedbackevent.schema.json
```diff
@@ -0,0 +1,244 @@
...
+            },
+            "xdm:bounceClass": {
+              "title": "Bounce Class",
+              "type": "string",
+              "description": "Classification code for a given message"
```
mjasrotia: Pleaese add (.) to end of statement

extensions/adobe/experience/campaign/feedbackevent.schema.json
```diff
@@ -0,0 +1,244 @@
...
+            }
+          }
+        },
+        "xdm:solutionContext": {
+          "title": "Solution Context",
+          "type": "object",
+          "description": "Solution specific data",
```
mjasrotia: Pleaese add (.) to end of statement

extensions/adobe/experience/campaign/feedbackevent.schema.json
```diff
@@ -0,0 +1,244 @@
...
+          "$ref": "https://ns.adobe.com/experience/analytics/keyedlist"
+        },
+        "https://ns.adobe.com/experience/campaign/channel": {
+          "title": "Delivery channel",
+          "type": "object",
+          "properties": {
+            "xdm:channelType": {
+              "title": "Channel Type",
+              "type": "string",
+              "description": "Classifies whether the channel is email, push, sms"
```
mjasrotia: Pleaese add (.) to end of statement

extensions/adobe/experience/campaign/feedbackevent.schema.json
```diff
@@ -0,0 +1,244 @@
...
+                    },
+                    "xdm:reportBy": {
+                      "title": "Report By",
+                      "type": "string",
+                      "description": "E-mail Address of the entity reporting this message as spam."
+                    },
+                    "xdm:reportTo": {
+                      "title": "Report To",
+                      "type": "string",
+                      "description": "E-mail Address to which this spam report is to be delivered"
```
mjasrotia: Pleaese add (.) to end of statement

extensions/adobe/experience/campaign/feedbackevent.schema.json
```diff
@@ -0,0 +1,244 @@
...
+                    },
+                    "xdm:userAgent": {
+                      "title": "User Agent",
+                      "type": "string",
+                      "description": "Value of the browser's User-Agent header"
```
mjasrotia: Pleaese add (.) to end of statement

extensions/adobe/experience/campaign/feedbackevent.schema.json
```diff
@@ -0,0 +1,244 @@
...
+                  }
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  },
+  "allOf": [
```
mjasrotia: Please add 
`{
      "$ref": "https://ns.adobe.com/xdm/common/extensible#/definitions/@context"
    },
`

extensions/adobe/experience/campaign/feedbackevent.schema.json
sauagarw: We will be receiving different feedback events for each one of the to/CC/BCC. So, this will be one at a time.

extensions/adobe/experience/campaign/feedbackevent.schema.json
sauagarw: Yes, this is the feedback type. Updated the title.

sauagarw:
@mjasrotia Thanks for the review. Incorporated the comments. Please review.

sauagarw:
@khanna-vikas : We have defined objects in the xdm with namespaces E.g. "https://ns.adobe.com/experience/campaign/messageFailure". Can we define these fields as standard fields instead like "xdm:messageFailure".

khanna-vikas:
@sauagarw : We need to understand the semantic meaning of messageFailure, not just in the context of the campaign. If we can apply the same property elsewhere that doesn't change the meaning we are all good to standardize the property in the xdm namespace.

sauagarw:
@khanna-vikas Thanks for clarification. A couple of questions:

1. Keeping the schema as is, when we would fetch this xdm from the xdm registry, we would get "messageFailure" node instead of "https://ns.adobe.com/experience/campaign/messageFailure" node. Am I right?
I have looked at the following references which talks about xed format. 
Ref: https://wiki.corp.adobe.com/display/DMSArchitecture/XDM+Registry+-+Composition+API#XDMRegistry-CompositionAPI-XDMCompositionAPI , https://github.com/adobe/xdm/blob/master/docs/extensions.md

2. Also, we have defined  "https://ns.adobe.com/experience/campaign/recipient" object inside of "https://ns.adobe.com/experience/campaign/message" object. How will the path be mapped when we fetch the xdm(in xed form) given both message & recipient have the same namespace path but recipient is a child of message. 

Thanks!

khanna-vikas:
1. You are right. When we prefix the core namespace (xdm) along with the property name the namespace by default is stripped off when preserving in the XED format.
2. the things become ugly, as the property under the object will also carry the namespace initials in the XED format. so to access the recipient it would be _experience.campaign.message._experience.campaign.recipient.

We will take this feedback and try to overcome this issue in the next upcoming sprints.

sauagarw:
@khanna-vikas Thanks a lot. 
Please correct me if I am wrong. This means at the time of publishing data corresponding to this xdm to pipeline/platform, we would have to pass on the full path as per the XED format "_experience.campaign.message._experience.campaign.recipient."  i.e. passing value with "recipient" key would not suffice.
If this is right, then we would want to keep our objects in xdm namespace("xdm:") like we have kept for all other fields E.g. "xdm:retryCount"

khanna-vikas:
@sauagarw : xdm is the core namespace where we standardize the data types/mixins and schemas for our tenants to use and apply in the real world.When we define any property under xdm that gets the same semantic meaning across the board for all the services to understand and acknowledge. This is where the governance to data modelling happens where each property is very closely monitored and discussed for its full enriched semantic meaning. 

tenants, partners can, of course, refer to this xdm: property while building their compositions but are guided to create new properties under the guiding consultation with the DM team.

Please reattribute your definitions under your namespace and not under the xdm namespace. If you wish to get them under xdm namespace, they have to be more legitimate in describing meaning of that property so that others should be able to reuse it.

sauagarw:
@khanna-vikas Thanks for the detailed explanation. I have made the changes. Please review.

extensions/adobe/experience/campaign/feedbackevent.schema.json
```diff
@@ -0,0 +1,244 @@
...
+        "https://ns.adobe.com/experience/campaign/message": {
+          "title": "A single message sent to a recipient.",
+          "type": "object",
+          "properties": {
+            "xdm:id": {
+              "title": "Message Identifier",
+              "type": "string",
+              "description": "Identifier of the message."
+            },
+            "xdm:subject": {
```
khanna-vikas: the "subject" is a very open and wider concept. It could be the study of anyThing. In this case, it is semantically "subject of the email message" sent to recipients. Use xdm:emailSubject to make it more meaningful or use Subject under your namespace.

extensions/adobe/experience/campaign/feedbackevent.schema.json
```diff
@@ -0,0 +1,244 @@
...
+              "description": "A human-friendly name of the campaign activity which is originating this message."
+            },
+            "xdm:name": {
+              "title": "Delivery Internal Name",
+              "type": "string",
+              "description": "A human-friendly identifier of the campaign activity which is originating this message."
+            }
+          }
+        },
+        "https://ns.adobe.com/experience/campaign/message": {
```
khanna-vikas: Maybe instead of the "message", say it "email/message", then you can use xdm: subject and xdm: size more as properties as they semantically make the properties much richer.

extensions/adobe/experience/campaign/feedbackevent.schema.json
```diff
@@ -0,0 +1,244 @@
...
+              "title": "Message Identifier",
+              "type": "string",
+              "description": "Identifier of the message."
+            },
+            "xdm:subject": {
+              "title": "Message Subject",
+              "type": "string",
+              "description": "Subject line of the message."
+            },
+            "xdm:size": {
```
khanna-vikas: Same here! xdm:emailMessageSize should be more appropriate, else work on the object definition /email/message on the line no 64.

extensions/adobe/experience/campaign/feedbackevent.schema.json
```diff
@@ -0,0 +1,244 @@
...
+                "https://ns.adobe.com/experience/campaign/interactions": {
+                  "title": "User Interactions",
+                  "type": "object",
+                  "properties": {
+                    "xdm:fbType": {
+                      "title": "FB Type",
+                      "type": "string",
+                      "description": "Type of spam report entered against this message (see RFC 5965 § 7.3)."
+                    },
+                    "xdm:reportBy": {
```
khanna-vikas: reportedBy

extensions/adobe/experience/campaign/feedbackevent.schema.json
```diff
@@ -0,0 +1,244 @@
...
+            },
+            "https://ns.adobe.com/experience/campaign/email": {
+              "title": "E-mail Channel",
+              "type": "object",
+              "properties": {
+                "https://ns.adobe.com/experience/campaign/interactions": {
+                  "title": "User Interactions",
+                  "type": "object",
+                  "properties": {
+                    "xdm:fbType": {
```
khanna-vikas: What is fbType!

extensions/adobe/experience/campaign/feedbackevent.schema.json
```diff
@@ -0,0 +1,244 @@
...
+                "https://ns.adobe.com/experience/campaign/operationalContext": {
+                  "title": "Operational Context",
+                  "type": "object",
+                  "properties": {
+                    "xdm:outboundIP": {
+                      "title": "IP Address",
+                      "type": "string",
+                      "description": "Outbound IP address of the customer used to deliver the message."
+                    },
+                    "xdm:deliveryMethod": {
```
khanna-vikas: can we move  "deliveryMethod" as "method"  part of "delivery object" at line 28?

extensions/adobe/experience/campaign/feedbackevent.schema.json
```diff
@@ -0,0 +1,244 @@
...
+  "description": "Contains all delivery event fields.",
+  "definitions": {
+    "feedbackevent": {
+      "properties": {
+        "https://ns.adobe.com/experience/campaign/feedbackType": {
+          "title": "Feedback Type",
+          "type": "string",
+          "description": "The type of message event."
+        },
+        "https://ns.adobe.com/experience/campaign/retryCount": {
```
khanna-vikas: this can be xdm:retryCount as it has a well understood semantic meaning to the definition

extensions/adobe/experience/campaign/feedbackevent.schema.json
sauagarw: The delivery feedback events that we are defining could be for email/push/sms. Any field specific to these channels will go in their channel block. All other fields which are common to all of these channels are defined in rest of the blocks. So, subject here would also mean subject of the push notification or sms.

extensions/adobe/experience/campaign/feedbackevent.schema.json
sauagarw: Same comment as above. The message block would common for other channels as well.

extensions/adobe/experience/campaign/feedbackevent.schema.json
sauagarw: Same as above.

extensions/adobe/experience/campaign/feedbackevent.schema.json
sauagarw: This field is specific to the email channel. If someone has reported an email as spam, then this field would capture email address of that person. It is a type of user interaction with the email message. Hence, it is under "https://ns.adobe.com/experience/campaign/email->https://ns.adobe.com/experience/campaign/interactions".

extensions/adobe/experience/campaign/feedbackevent.schema.json
sauagarw: This field is specific to the email channel. If the email has been reported as spam, then what type of spam is it. E.g. abuse, fraud. Ref: https://tools.ietf.org/html/rfc5965#section-7.3.

extensions/adobe/experience/campaign/feedbackevent.schema.json
sauagarw: Here "deliveryMethod" captures the protocol by which the email message is delivered. E.g. smtp or http. Since, this field is specific to the email channel, we have kept it under "https://ns.adobe.com/experience/campaign/email ->
https://ns.adobe.com/experience/campaign/operationalContext".

extensions/adobe/experience/campaign/feedbackevent.schema.json
sauagarw: Got it. Done.

extensions/adobe/experience/campaign/feedbackevent.schema.json
sauagarw: Done.

sauagarw:
@khanna-vikas To avoid such hierarchies "_experience.campaign.message._experience.campaign.recipient." we have removed URIs from nested objects. Please let me know in case of any concerns.

PR #832, opened 2020-02-19
B2B Leads

mjasrotia:
New Schemas:
•	xdm/common/essam
•	xdm/common/essa
•	xdm/context/lead
•	xdm/context/market-lead


schemas/context/marketo-lead.schema.json
```diff
@@ -0,0 +1,75 @@
...
+  ],
+  "definitions": {
+    "marketo-lead": {
+      "properties": {
+        "marketo:leadId": {
+          "title": "Marketo Lead ID",
+          "description": "Marketo lead id.",
+		  "type": "string"
+        },
+        "marketo:mktPersonId": {
```
cfraser, line 31: We should consider using the full `marketo` here `mkt` is not descriptive enough. Given that it's in the `marketo` namespace we could maybe eliminate it altogether.

mjasrotia: Resolved in the new PR 
https://github.com/adobe/xdm/pull/854

mjasrotia:
Opened another PR for Leads

https://github.com/adobe/xdm/pull/854

PR #834, opened 2020-02-22
Map data type not explicitly defined

cmathis:
Please link to the issue #…


extensions/adobe/experience/profile-edgeregion.schema.json
```diff
@@ -21,6 +21,8 @@
           "type": "object",
           "meta:xdmType": "map",
           "additionalProperties": {
+            "type": "object",
+            "meta:xdmType": "object",
```
khanna-vikas: @cmathis why do we need to define the meta:xdmType inside additionalProperties?

PR #838, opened 2020-02-26
Tune and expand the proposition-details mix-in for the XDM Decision Event class

jbeckert:
Please link to the issue #…


extensions/adobe/experience/decisioning/proposition-detail.schema.json
```diff
@@ -16,19 +17,24 @@
     "proposition-detail-datatype": {
       "properties": {
         "xdm:activity": {
-          "title": "Decision Activity",
+          "title": "Decision Activity Snapshot",
           "description": "Snapshot of the activity for which the proposition was made. The object contains at least the xdm:id property value which is the identifier of the decision activity.",
           "$ref": "https://ns.adobe.com/experience/decisioning/activity-detail"
         },
         "xdm:channel": {
```
cmathis: May want to add "meta:status": "deprecated" to this field

jbeckert: Thank you, added this meta field

extensions/adobe/experience/decisioning/activity-detail.schema.json
```diff
@@ -7,41 +7,46 @@
...
     "activity-detail-datatype": {
       "properties": {
         "xdm:id": {
           "title": "Activity Identifier",
           "type": "string",
           "format": "uri-reference",
-          "description": "A unique identifier for the decision activity."
+          "description": "A unique, immutable identifier for the decision activity."
+        },
+        "xdm:version": {
```
cmathis: @prabhum2  and @khanna-vikas  - Is this to become the official definition of xdm:version field? Or should this be named xdm:activityVersion or something since it's related to the activity?

extensions/adobe/experience/decisioning/option-detail.schema.json
```diff
@@ -16,19 +16,34 @@
     "option-detail-datatype": {
       "properties": {
         "xdm:id": {
-          "title": "Option Identifier",
+          "title": "Decision Option Identifier",
           "type": "string",
           "format": "uri-reference",
           "description": "A unique identifier for the decision option."
         },
+        "xdm:version": {
```
cmathis: Same here - should this be xdm:decisionVersion?

extensions/adobe/experience/decisioning/option-detail.schema.json
```diff
@@ -16,19 +16,34 @@
...
+          "type": "string",
+          "description": "The version that the decision option object was at when the snapshot was taken."
+        },
         "xdm:name": {
           "type": "string",
-          "title": "Option Name",
+          "title": "Decision Option Name",
           "description": "Option name. The name is displayed in various user interfaces."
         },
+        "xdm:tags": {
```
cmathis: Similar question here - is this the formal definition of xdm:tags in all of XDM?

extensions/adobe/experience/decisioning/placement-detail.schema.json
```diff
@@ -0,0 +1,51 @@
...
+          "title": "Placement Version",
+          "type": "string",
+          "description": "The version that the placement object was at when the snapshot was taken."
+        },
+        "xdm:name": {
+          "title": "Placement Name",
+          "type": "string",
+          "description": "Placement name. The name is displayed in various user interfaces."
+        },
+        "xdm:channel": {
```
cmathis: Different definition for xdm:channel - https://github.com/adobe/xdm/blob/master/schemas/context/experienceevent-channel.schema.json#L19

cmathis:
My main comment here is that we appear to be redefining many of the existing xdm: field names in in different ways. If  fields such xdm:channel, xdm:title, xdm:version are to be reused across different  XDM schema then do we want that field to remain statically defined the same way across all those?

extensions/adobe/experience/decisioning/option-detail.schema.json
jbeckert: Do we have a notion of eTag? This "version" refers to the entity referenced above? It could be called "xdm:entityVersion" or "something:etag".
decisionVersion would not be an ideal choices, because the thing that has the version is not the "decision" per se.

extensions/adobe/experience/decisioning/option-detail.schema.json
jbeckert: The only other usage of "xdm:tags" besides in decisioning/offer management is in "https://ns.adobe.com/xdm/data/datasource" and in campaigns experience events. The latter are exactly the tags introduced by decisioning. 
For datasources the tags are only strings, for decisioning they are defined entities, i.e. they have an identity and two other entities referencing the tags with the same ID really mean to share the tag.

extensions/adobe/experience/decisioning/placement-detail.schema.json
jbeckert: The difference is that the experience event mixin inlines the definition and the placement references the channel. Since channel has an `@id` property it should not really be "inlined", otherwise you end up with a JSON object that has multiple `@id`s. According a discussion with Leonard we should not embed/inline an entity schema in another entity schema.
If you ask me it should be a string whose value is that of the channel entity's `@id` property.

jbeckert:
@cmathis Your comment above is well-taken and I would like to minimize or eliminate usage of the same thing by different names and the usage of different things by the same name.
I'm not aware of the usage of "xdm:title" in the decisioning schemas.
For "xdm:version" we could use "repo:etag" if that is the way forward.
For "xdm:tags" we have existing usages in the decisioning and offer management domain and would not be able to change that. I also thing the change to a semantically less rich tag as use in datasources is not desired.
For "xdm:channel" I could use the property "xdm:channelID" to indicate a reference to the channel entity

extensions/adobe/experience/decisioning/option-detail.schema.json
jbeckert: I moved to a different property name "repo:etag". It is used elsewhere and is used in the same way we are using it here: to denote a particular snapshot in time for an entity/resource

extensions/adobe/experience/decisioning/activity-detail.schema.json
jbeckert: I moved to a different property name "repo:etag". It is used elsewhere and is used in the same way we are using it here: to denote a particular snapshot in time for an entity/resource

PR #839, opened 2020-03-03
OBS-2320 Adding ingestion measurements

tanhuynh:
Fix for #https://jira.corp.adobe.com/browse/OBS-2321


cmathis:
@tanhuynh - This is approved, but before we merge it in we need to fix the build errors. Looks like a formatting issue only.

PR #840, opened 2020-03-06
Add property 'locale' to Environment.

kevinlind:
Adding property 'locale' to the Environment schema to state the language of the environment at the time of the event. The locale code should follow [RFC BCP 47](https://tools.ietf.org/html/bcp47)

Change requested by Mobile SDK as 'locale' is an environment metric sent to Analytics in application lifecycle events.


docs/reference/context/environment.schema.json
```diff
@@ -137,6 +137,16 @@
                     "title": "Internet service provider",
                     "type": "string",
                     "description": "The name of the user's internet service provider."
+                },
+                "xdm:locale": {
```
lrosenthol: You don't need this since XDM is valid JSON-LD, you should be using the standard `@language` value.

docs/reference/context/environment.schema.md
```diff
@@ -38,7 +38,8 @@ Information about the surrounding situation the event observation occurred, spec
   "xdm:operatingSystemVendor": "Apple",
   "xdm:connectionType": "cable",
   "xdm:domain": "verizon.com",
-  "xdm:ISP": "Verizon"
+  "xdm:ISP": "Verizon",
```
lrosenthol: NOTE: this example is not valid XDM as it's plain JSON and not JSON-LD!

docs/reference/context/environment.schema.md
```diff
@@ -240,6 +242,33 @@ The numerical label assigned to a device participating in a computer network tha
 
 
 
+## xdm:locale
```
lrosenthol: See comments as to why this is unnecessary

schemas/context/environment.schema.json
```diff
@@ -143,6 +143,16 @@
           "title": "Internet service provider",
           "type": "string",
           "description": "The name of the user's internet service provider."
+        },
```
lrosenthol: See comments about why you don't need this.

kevinlind:
@lrosenthol Thanks for your feedback. I've updated this PR to use `dc:language` instead of creating a new XDM property.

PR #851, opened 2020-03-27
Remove couple of dup fields which are deprecated

jwen-adobe:
This PR removes couple of duplicated fields which are deprecated anyway. These dup fields can not pass platform protobuf lib creation.


cmathis:
This is never allowed. Those old fields are likely still used by old implementations.

PR #861, opened 2020-04-09
Logical mixin groupings v1

eknee:
Creating a logical mixin collection based on customer feedback.


eknee:
Build is failing because package.json's config setting for schema is on an older version.  Running the latest NPM install which uses version 252 (currently set to 249)

PR #862, opened 2020-04-13
Person mixin and person name updates

eknee:
--added mailing address to Address Mixin
--new field added to person-name datatype called "suffix"
--new fields added to pesron mixin (maritalStatus, nationality, taxId)

### DIFF
--- a/schemas/context/person.example.1.json
+++ b/schemas/context/person.example.1.json
@@ -5,5 +5,10 @@
     "xdm:lastName": "Doe",
     "xdm:fullName": "Jane F. Doe"
   },
-  "xdm:birthDate": "1996-01-03"
+  "xdm:birthDate": "1996-01-03",
+  "xdm:birthDayAndMonth": "01-03",
+  "xdm:birthYear": "1996",
+  "xdm:gender": "female",
+  "xdm:maritalStatus": "single",
+  "xdm:nationality": "CAN"
 }
--- a/schemas/context/person.example.1.json
+++ b/schemas/context/person.example.1.json
@@ -7,7 +7,7 @@
   },
   "xdm:birthDate": "1996-01-03",
   "xdm:birthDayAndMonth": "01-03",
-  "xdm:birthYear": "1996",
+  "xdm:birthYear": 1996,
   "xdm:gender": "female",
   "xdm:maritalStatus": "single",
   "xdm:nationality": "CAN"
--- a/schemas/context/person.example.1.json
+++ b/schemas/context/person.example.1.json
@@ -6,8 +6,6 @@
     "xdm:fullName": "Jane F. Doe"
   },
   "xdm:birthDate": "1996-01-03",
-  "xdm:birthDayAndMonth": "01-03",
-  "xdm:birthYear": 1996,
   "xdm:gender": "female",
   "xdm:maritalStatus": "single",
   "xdm:nationality": "CAN"
--- a/schemas/context/person.example.1.json
+++ b/schemas/context/person.example.1.json
@@ -8,5 +8,5 @@
   "xdm:birthDate": "1996-01-03",
   "xdm:gender": "female",
   "xdm:maritalStatus": "single",
-  "xdm:nationality": "CAN"
+  "xdm:nationality": "CA"
 }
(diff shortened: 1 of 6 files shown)


schemas/context/person.example.1.json
```diff
@@ -5,5 +5,10 @@
     "xdm:lastName": "Doe",
     "xdm:fullName": "Jane F. Doe"
   },
-  "xdm:birthDate": "1996-01-03"
+  "xdm:birthDate": "1996-01-03",
```
lrosenthol, line 8: According to the documentation, you should *not* have `birthYear` or `birthDayAndMonth` if you know the full date.   Please remove.

schemas/context/person.schema.json
```diff
@@ -56,6 +56,27 @@
...
+                },
+                "xdm:maritalStatus": {
+                    "title": "Marital Status",
+                    "type": "string",
+                    "enum": [
+                        "married",
+                        "single",
+                        "not_specified"
+                    ],
+                    "meta:enum": {
```
lrosenthol, line 68: these are the only enums we are going to specify?  Not `divorced`, `widowed`, etc?

schemas/context/person.schema.json
```diff
@@ -56,6 +56,27 @@
...
+                        "married": "Married",
+                        "single": "Single",
+                        "not_specified": "Not Specified"
+                    },
+                    "description": "Marital status of the person.\n",
+                    "default": "not_specified"
+                },
+                "xdm:nationality": {
+                    "title": "Nationality",
+                    "type": "string",
```
lrosenthol, line 78: I would add a pattern match here so that invalid values aren't entered.

schemas/context/person.example.1.json
eknee: Those values are straight from the existing schema.  Why would we not put them in the example?

schemas/context/person.schema.json
eknee: Valid point.  I was trying to capture this as its showing up on a customer schema and wanted to ensure we capture it as well.  I can add other values here if we think this is relevant to add as a field at all.  Thoughts?

schemas/context/person.schema.json
eknee: I'll add this.  Given other fields in XDM use a 2 code pattern should we enforce consistency across the ISO 3166-1 standards to use only 2 or 3 code patterns?

schemas/context/person.example.1.json
lrosenthol: Read the documentation the schema - it is very clear that you *do not* create an actual JSON packet like this.  If you want, create multiple examples, each correct with different sets of values.

schemas/context/person.schema.json
lrosenthol: I am fine with it as a new field in the schema, but you need to make sure that when using an enum, you specify all the correct values - as you can't add them later.

schemas/context/person.schema.json
lrosenthol: Actually, whether to use 2 or 3 may differ based on the use case.  We also use BCP-47 codes in some cases as well

eknee: Resolved in latest commit

schemas/context/person.schema.json
eknee: Resolved in latest commit

schemas/context/person.example.1.json
eknee: Resolved in latest commit

schemas/context/person.schema.json
khanna-vikas: @eknee We need to maintain the consistency for the pattern "Countrycode" and "Nationality" as both semantically lean towards each other. Since we already have 3166-1 standards maintained in xdm via CountryCode, it is advisable to have nationality follow that standard.

eknee: @khanna-vikas this has been updated to use the 3 digit ISO code.

schemas/context/person.example.1.json
```diff
@@ -5,5 +5,8 @@
...
+  "xdm:gender": "female",
+  "xdm:maritalStatus": "single",
+  "xdm:nationality": "CAN"
```
khanna-vikas, line 11: Match it with the specifications, [A-Z]{2}

eknee: Fixed in new push

PR #863, opened 2020-04-20
adding essa schemas

prabhum2:
Please link to the issue #…

### DIFF
--- a/schemas/common/external-source-system-audit.schema.json
+++ b/schemas/common/external-source-system-audit.schema.json
@@ -1,0 +1,69 @@
+{
+  "meta:license": [
+    "Copyright 2020 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/common/external-source-system-audit",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "External Source System Audit Attributes",
+  "type": "object",
+  "meta:extensible": true,
+  "description": "External Source System Audit fields.",
+  "definitions": {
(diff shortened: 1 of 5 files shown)


schemas/common/external-source-system-audit.schema.json
```diff
@@ -0,0 +1,69 @@
...
+  "description": "External Source System Audit fields.",
+  "definitions": {
+    "externalSourceSystemAudit": {
+      "properties": {
+        "xdm:createdBy": {
+          "title": "Created By",
+          "description": "Name of the user who created this record",
+          "type": "string"
+        },
+        "xdm:createdDate": {
```
cmathis, line 22: Why not use the created and modify date that all other schemas use - https://github.com/adobe/xdm/blob/master/schemas/external/repo/common.schema.json#L32

schemas/common/external-source-system-audit.schema.json
```diff
@@ -0,0 +1,69 @@
+{
+  "meta:license": [
+    "Copyright 2020 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/common/external-source-system-audit",
```
cmathis, line 8: How is this different the existing common/auditable schema - https://github.com/adobe/xdm/blob/master/schemas/common/auditable.schema.json

Are we saying everything in the common/auditable is "internal" audit fields? If I include both of these, am I going to be confused when I see both a xdm:createdDate and a repo:createdDate? Should these fields be named something like xdm:externalCreatedDate? These are just some questions I have.

khanna-vikas: As mentioned this specific mixin will be used as a reference to capture the auditable properties of customers data sets. We have CRM system and other Marketo needs to capture and present when was a record got created Vs Modified and by whom. common/auditable are the fields used meant to be used internally by the AEP services.

PR #866, opened 2020-04-24
Fixed profile phones title

eknee:
--incorrect title text was updated to display the correct text


prabhum2:
@eknee - Do you still need this PR ?

eknee:
This is still needed but was held up in discussions with product.  I would like to see this pushed in but I'm not sure where it stands currently.

prabhum2:
moved to https://github.com/adobe/xdm/pull/930

Issue #867, opened 2020-04-24
Add more breaking changes check


jwen-adobe:
What are the schemas that are affected by the issue
-all schemas contain breaking changes

What are examples of products that are impacted by the issue
-all schemas contain breaking changes

PR #868, opened 2020-04-24
Add more breaking changes check and remove some logging

jwen-adobe:
This PR links to the issue #867 which includes some logging cleanup and new breaking changes check below:

-when a field's format got changed
-when a field's minimum got changed to higher value
-when a field's maximum got changed to lower value
-when a new key's lowercase is the same as existing non-deprecated key's lowercase in the same level


bin/incompatibility-check.js
```diff
@@ -60,21 +66,37 @@ function checkBreakingChanges(files) {
...
+                }
+
+                if (isPropertyChanged(differences, brokenProperty)) { //check changed properties
+                    if (brokenProperty.isFieldTypeBroken) {
+                        errLogs.push(workingFile+' breaking changes found!!! Data type of property "' + brokenProperty.name + '" can not be changed.\n');
+                    }
+                    if (brokenProperty.isStringFormatBroken) {
+                        errLogs.push(workingFile+' breaking changes found!!! String format of property "' + brokenProperty.name + '" can not be changed.\n');
+                    }
+                    if (brokenProperty.isMinimumBroken) {
```
cmathis, line 87: Min and Max values are used to determine the appropriate meta:xdmType. As long as the change does not change the binary type (make the field go from long to integer for example), then it is not a breaking change.

jwen-adobe: this check is to ensure the data range can only be expanded.

PR #869, opened 2020-04-24
Updated hierarchy grouping of fields

eknee:
--grouped fields under a master object for each logical grouping mixin.

Example dot notation:
address.workAddress
address.homeAddress
email.personalEmail
email.workEmail
phone.homePhone
phone.mobilePhone
phone.workPhone
phone.faxPhone

Please link to the issue #…

### DIFF
--- a/schemas/context/profile-addresses.schema.json
+++ b/schemas/context/profile-addresses.schema.json
@@ -16,26 +16,33 @@
   "definitions": {
     "profile-addresses": {
       "properties": {
-        "xdm:homeAddress": {
-          "title": "Home Address",
-          "$ref": "https://ns.adobe.com/xdm/common/address",
-          "description": "A home postal address."
-        },
-        "xdm:workAddress": {
-          "title": "Work Address",
-          "$ref": "https://ns.adobe.com/xdm/common/address",
-          "description": "A work postal address."
-        },
-        "xdm:shippingAddress": {
-          "title": "Shipping Address",
-          "$ref": "https://ns.adobe.com/xdm/common/address",
-          "description": "A shipping postal address."
-        },
-        "xdm:mailingAddress": {
-          "title": "Mailing Address",
-          "$ref": "https://ns.adobe.com/xdm/common/address",
-          "description": "A mailing postal address."
-        }              
+        "xdm:address": {
+          "title": "Address",
+          "type": "object",
+          "description": "",
+          "properties": {
+            "xdm:homeAddress": {
+              "title": "Home Address",
+              "$ref": "https://ns.adobe.com/xdm/common/address",
+              "description": "A home postal address."
+            },
+            "xdm:workAddress": {
+              "title": "Work Address",
+              "$ref": "https://ns.adobe.com/xdm/common/address",
+              "description": "A work postal address."
+            },
+            "xdm:shippingAddress": {
+              "title": "Shipping Address",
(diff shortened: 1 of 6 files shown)


schemas/context/profile-addresses.schema.json
```diff
@@ -16,26 +16,33 @@
...
-        },
-        "xdm:mailingAddress": {
-          "title": "Mailing Address",
-          "$ref": "https://ns.adobe.com/xdm/common/address",
-          "description": "A mailing postal address."
-        }              
+        "xdm:address": {
+          "title": "Address",
+          "type": "object",
+          "description": "",
```
khanna-vikas, line 22: Add a description of the object Address, "Collection of various type of addresses for a person profile"

schemas/context/profile-emails.schema.json
```diff
@@ -16,15 +16,22 @@
...
-          "description": "A personal email address."
-        },
-        "xdm:workEmail": {
-          "title": "Work Email",
-          "$ref": "https://ns.adobe.com/xdm/context/emailaddress",
-          "description": "A work email address."
+        "xdm:email": {
+          "title": "Email",
+          "type": "object",
+          "description": "",
```
khanna-vikas, line 22: Same here, define the description.

schemas/context/profile-phones.schema.json
```diff
@@ -16,26 +16,33 @@
...
-        },
-        "xdm:faxPhone": {
-          "title": "Fax Phone",
-          "$ref": "https://ns.adobe.com/xdm/context/phonenumber",
-          "description": "Fax phone number."
-        }                                      
+        "xdm:phone": {
+          "title": "Phone",
+          "type": "object",
+          "description": "",
```
khanna-vikas, line 22: Same here

schemas/context/profile-addresses.schema.json
eknee: Updated description in new push.

schemas/context/profile-emails.schema.json
eknee: Updated description in new push.

schemas/context/profile-phones.schema.json
eknee: Updated description in new push.

eknee:
Fixed profile-phones mixin to have the correct title.

cmathis:
This PR is producing a new concept within AEP where a given field (xdm:workAddress for example) can now exist in two places - at the root of the schema OR nested under a top level address field. This may be okay, but I'm afraid most would consider these two completely different fields that mean different things. We need to make sure all consumers of this data are aligned on how these two fields should be treated.

eknee:
Agreed.  I think that in order to provide the flexibility our customers want with logical groupings this provides a way to present a standard without having them start creating everything themselves within the default _tenant root.  The benefit with this is that additional email types that we don't support can now clearly be placed as a extension to this mixin which will keep customers from purely using the root _tenant object for custom attributes/datatypes.

prabhum2:
@eknee - Do you still need this PR?

eknee:
Still need this PR and would like to see it pushed in.  Waiting on product approval to deploy.

prabhum2:
@eknee  sounds good . Please let me know when you are ready for us to get this pushed.

Issue #871, opened 2020-04-27
XDM support for Player State Tracking fields


midomir:
## What are the schemas that are affected by the issue
* media-timed-asset-view-details.schema.json
* player-state.schema.json [NEW]

## What are examples of products that are impacted by the issue
* Adobe Media Analytics

Internal JIRA issue: https://jira.corp.adobe.com/browse/VA-3232

High-level changes:
* addition of player state tracking fields (fullScreen, mute, closeCaption, pictureInPicture, inFocus)

PR #872, opened 2020-04-27
[VA-3232] XDM support for Player State Tracking

midomir:
Issue #871

### DIFF
--- a/schemas/context/player-state.schema.json
+++ b/schemas/context/player-state.schema.json
@@ -1,0 +1,46 @@
+{
+  "meta:license": [
+    "Copyright 2019 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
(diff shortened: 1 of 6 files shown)


schemas/context/player-state.schema.json
```diff
@@ -0,0 +1,44 @@
...
+    "player-state": {
+      "properties": {
+        "xdm:playerStateSet": {
+          "title": "Player state set",
+          "type": "boolean",
+          "description": "Whether or not the player state is set on that state."
+        },
+        "xdm:playerStateCount": {
+          "title": "Player state count",
+          "type": "number",
```
lrosenthol, line 25: Should that be an integer?  What about a min/max?

schemas/context/player-state.schema.json
```diff
@@ -0,0 +1,44 @@
...
+          "description": "The number of times that player state was set on the stream."
+        },
+        "xdm:playerStateTime": {
+          "title": "Show type",
+          "type": "number",
```
lrosenthol, line 30: Should that be an integer?  What about a min/max?

schemas/context/player-state.schema.json
midomir: Done

schemas/context/player-state.schema.json
midomir: Done

PR #886, opened 2020-05-08
Adding resources for Activity type

khanna-vikas:
Please link to the issue #…


schemas/common/activity-type-reference-details.schema.json
```diff
@@ -0,0 +1,65 @@
...
+        "xdm:active": {
+          "title": "Status of the activity type",
+          "type": "boolean",
+          "description":
+            "A boolean flag to depict if the activity type can be used by the business system,or has been discontinued."
+        }
+      }
+    }
+  },
+  "allOf": [
```
mjasrotia, line 59: We should be adding this here.
 "$ref": "https://ns.adobe.com/xdm/common/extensible#/definitions/@context"

Issue #897, opened 2020-05-13
Change numeric fields to be of type integer


midomir:
## What are the schemas that are affected by the issue
* schemas/context/advertising-break.schema.json
* schemas/context/advertising-timed-asset-reference.schema.json
* schemas/context/advertising-timed-asset-view-details.schema.json
* schemas/context/advertising.schema.json
* schemas/context/media-timed-asset-reference.schema.json
* schemas/context/media-timed-asset-view-details.schema.json
* schemas/context/media-timed-chapter-asset-reference.schema.json
* schemas/context/media-timed-chapter-view-details.schema.json
* schemas/context/media-timed-chapter.schema.json
* schemas/context/media-timed-qoe.schema.json
* schemas/context/media-timed.schema.json

## What are examples of products that are impacted by the issue
* Adobe Media Analytics

Internal JIRA issue: https://jira.corp.adobe.com/browse/VA-3330

High-level changes:
* change the type of some fields from number to integer (only integer values are valid)
* changed the title of some fields to ensure it is aligned with the field name in the reports

PR #898, opened 2020-05-13
Change numeric fields to be of type integer

midomir:
Issue #897

### DIFF
--- a/schemas/context/advertising.schema.json
+++ b/schemas/context/advertising.schema.json
@@ -31,18 +31,21 @@
           "$ref": "https://ns.adobe.com/xdm/context/advertising-timed-asset-view-details"
         },
         "xdm:completes": {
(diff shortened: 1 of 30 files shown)


schemas/context/advertising.schema.json
```diff
@@ -31,18 +31,21 @@
           "$ref": "https://ns.adobe.com/xdm/context/advertising-timed-asset-view-details"
         },
         "xdm:completes": {
-          "title": "Completes",
-          "$ref": "https://ns.adobe.com/xdm/data/measure",
```
khanna-vikas, line 35: @midomir This involves a breaking change. All the previous datasets become unusable with this change as we are trying to change the type of the property from "object" to "integer".

prabhum2:
@midomir - Do you still need this PR ?

PR #899, opened 2020-05-13
Adding profile-push-notification-details schema

iewilliam:
Adding a profile mixin for storing mobile device push notification token for the device. The tokens are stored in an array to allow for profile stitching to combine all the push tokens, each one corresponding to a different mobile device, that the end user has.


schemas/context/profile-push-notification-details.schema.json
```diff
@@ -0,0 +1,41 @@
...
+    "meta:extensible": true,
+    "meta:abstract": true,
+    "meta:intendedToExtend": [
+        "https://ns.adobe.com/xdm/context/profile"
+    ],
+    "description": "This mixin is used to capture information regarding push notifications including mobile application and device details.",
+    "definitions": {
+        "profile-push-notification-details": {
+            "properties": {
+                "xdm:pushDetails": {
```
khanna-vikas, line 21: Represent the name explicitly "pushNotifictionDetails"

schemas/context/pushdetail.schema.json
```diff
@@ -0,0 +1,54 @@
...
+    "$id": "https://ns.adobe.com/xdm/context/pushdetail",
+    "$schema": "http://json-schema.org/draft-06/schema#",
+    "title": "Push notification detail",
+    "type": "object",
+    "meta:extensible": true,
+    "description": "Push notification detail required to send mobile push notification to a specific mobile application on a given device.",
+    "definitions": {
+        "pushdetail": {
+            "properties": {
+                "xdm:pushToken": {
```
khanna-vikas, line 17: Remove the word "push" as the reach to the data element would be pushNotificationDetails.token

schemas/context/pushdetail.schema.json
```diff
@@ -0,0 +1,54 @@
...
+                    "title": "Push Token",
+                    "type": "string",
+                    "description": "Valid push token for the given push platform."
+                },
+                "xdm:pushPlatform": {
```
khanna-vikas, line 22: Same here!

schemas/context/pushdetail.schema.json
```diff
@@ -0,0 +1,54 @@
...
+                    "title": "Push Platform",
+                    "type": "string",
+                    "description": "Push provider, e.g. apns or fcm"
+                },
+                "xdm:isBlacklisted": {
```
khanna-vikas, line 27: Remove "is"

schemas/context/pushdetail.schema.json
```diff
@@ -0,0 +1,54 @@
...
+                    "title": "Blacklist Flag",
+                    "type": "boolean",
+                    "description": "Indicates whether the push token is blacklisted."
+                },
+                "xdm:appId": {
+                    "title": "Application Identifier",
+                    "type": "string",
+                    "description": "Unique Identifier for the mobile application."
+                },
+                "xdm:id": {
```
khanna-vikas, line 37: Is this the Notification Id?

iewilliam: Realistically this would be the ECID that was associated with this push token, so we can still know the ecid for the push token after profile merging was done.

lrosenthol: Then call it ECID - you *cannot* use `xmd:id` it's _reserved_

schemas/context/pushdetail.schema.json
```diff
@@ -0,0 +1,54 @@
...
+            "properties": {
+                "xdm:token": {
+                    "title": "Push Token",
+                    "type": "string",
+                    "description": "Valid push token for the given push platform."
+                },
+                "xdm:platform": {
+                    "title": "Push Platform",
+                    "type": "string",
+                    "description": "Push provider, e.g. apns or fcm"
```
lrosenthol, line 25: Are there only two possible values?

iewilliam: for now the possible values are fcm, apns, and apns-sandbox. This list might grow in the future.

schemas/context/pushdetail.schema.json
```diff
@@ -0,0 +1,54 @@
...
+                    "title": "Blacklist Flag",
+                    "type": "boolean",
+                    "description": "Indicates whether the push token is blacklisted."
+                },
+                "xdm:appID": {
+                    "title": "Application Identifier",
+                    "type": "string",
+                    "description": "Unique Identifier for the mobile application."
+                },
+                "xdm:ECID": {
```
khanna-vikas, line 37: ECID is the namespace value for the given IdentiyMap. Can you rename the property definition to be more generic  and call it as pushNotificationID

schemas/context/pushdetail.example.1.json
```diff
@@ -0,0 +1,7 @@
+{
+    "xdm:appId": "75eafb7e-fa44-4514-86fc-221e32c5aef9",
+    "xdm:pushToken": "99156313-c9df-4e54-9c6c-5740f940c3ca",
```
khanna-vikas, line 3: Modify the example with corresponding attribute/property described in JSON schemas

schemas/context/pushdetail.schema.json
iewilliam: is tokenID ok? conceptually speaking this is the ECID the push token was associated with in the originating profile pre-stitching.

khanna-vikas: Will this notification always be based on ECID? today/ in the near future? I would have modelled it via an object with (Key and Value as a paired object) with 'ECID' being the Key and value of ECID being the value. That way it becomes generic.

iewilliam: Today yes, on mobile we always create a new ecid on install and send in the profile event to platform with that ecid as the primary identity. For maximum flexibility what you suggested make sense, in case we move to some other id in the future, I'm not sure how to encode what you suggested properly in the schema, can you point to an example I can follow?

schemas/context/profile-push-notification-details.example.1.json
```diff
@@ -0,0 +1,11 @@
+{
+    "xdm:pushNotificationDetails": [
+        {
+            "xdm:appID": "75eafb7e-fa44-4514-86fc-221e32c5aef9",
+            "xdm:token": "99156313-c9df-4e54-9c6c-5740f940c3ca",
+            "xdm:platform": "apns",
+            "xdm:blacklisted": false,
+            "xdm:id": "43363765445157027836050171721132739889"
```
lrosenthol, line 8: need to update this to `xdm:ECID`

schemas/context/pushdetail.schema.json
iewilliam: I looked around a bit further and it looks like https://ns.adobe.com/xdm/context/profileStitchIdentity would do the job here, will this work?

schemas/context/pushdetail.schema.json
```diff
@@ -0,0 +1,54 @@
...
+                    "title": "Blacklist Flag",
+                    "type": "boolean",
+                    "description": "Indicates whether the push token is blacklisted."
+                },
+                "xdm:appID": {
+                    "title": "Application Identifier",
+                    "type": "string",
+                    "description": "Unique Identifier for the mobile application."
+                },
+                "xdm:tokenID": {
```
lrosenthol, line 37: shouldn't be ID - since it's not the same type of ID as appID.   Could this be just `token`?

khanna-vikas: Good point. I feel we should make it more enriched by saying "tokenSpecifications" or "tokenDetails"

iewilliam: would "identity" work? I don't think this key is reserved and "identity" is the more accurate term here, this key holds the "active" identity on the app in the mobile phone when we collect the push token from the OS.

lrosenthol: I am OK with `xdm:identity` since that seems to be what it is now...

PR #874, opened 2020-04-29
STATS-10318: Add support for Creative Management Platform specific fields in the AdCloud Experience Event XDM schema

larisaad:
Please link to the issue #…

PR #905, opened 2020-05-27
STATS-10318: Remove allOf segment reference from Ad Cloud Creative sc…

larisaad:
…hema

Please link to the issue #…


larisaad:
@lrosenthol @khanna-vikas I'm afraid there has been a mistake on adding the reference of `adcloudsegment` into the allOf object of the `creative` schema. We didn’t intend to have the top-level `adcloudsegment` in `creative`, but just to have it under the `segments` field. We want to fix this asap, before the schema reaches production. Is this possible? It is deployed since yesterday in int / stage.

larisaad:
This produces breaking changes, but i can assure you that no one besides Ad Cloud is using this.

larisaad:
> What was the previous PR which has gone in INT/Stage??

This one: https://github.com/adobe/xdm/pull/874

PR #911, opened 2020-06-10
Made gdprApplies and timestamp fields mandatory for privacy mixin. Ad…

ankushkd:
…ded privacy mixin to Audience Manager extension


ankushkd:
- Made gdprApplies as required field for consentString
- Made timestamp as required field for profile privacy mixin
- Added experience event privacy mixin in Audience Manager extension

jwen-adobe:
retest

PR #933, opened 2020-07-27
Add xdm:note to the alternateDisplayInfo descriptor

cmathis:
AEP is wanting to introduce a way for users to add additional user specific "notes" to a given field (how it should be used, etc.). They see this as something different from the description and would like it to be separate. Adding this attribute to the descriptor will enable a user to add a note to one of the standard fields.

### DIFF
--- a/schemas/common/descriptors/display/alternateDisplayInfo.schema.json
+++ b/schemas/common/descriptors/display/alternateDisplayInfo.schema.json
@@ -43,6 +43,16 @@
             }
           }
         },
+        "xdm:note": {
+          "title": "Note",
+          "type": "object",
+          "description": "When present, user friendly note to display. Similar to a description, but provides more details about the actual usage of the field.",
+          "patternProperties": {
+            ".+_.+": {
+              "type": "string"
+            }
+          }
+        },
         "meta:enum": {
           "title": "Extended meta:enum values",
           "type": "object",


schemas/common/descriptors/display/alternateDisplayInfo.schema.json
```diff
@@ -43,6 +43,16 @@
             }
           }
         },
+        "xdm:note": {
+          "title": "Note",
+          "type": "object",
+          "description": "When present, user friendly note to display. Similar to a description, but provides more details about the actual usage of the field.",
```
lrosenthol, line 49: Why is this a set of properties and not a simple string - esp. if it's to be "user friendly"??

cmathis: Like all the attributes on this descriptor, they support a localized value.
"xdm:note": {
   "en-us": "My English Note",
   "fr-fr": "My French note", 
}

lrosenthol: If that's the intent, then they all need to be changed such that the key is defined to match a pattern for ISO 3166 while the value is open.

cmathis: I thought the current "patternProperties" regular expression was doing that. If not, please provide the regular expression which does.

PR #937, opened 2020-08-10
endUser identifiers as standard XDM fields and corresponding ExpEvent Mixin

rajindal:
Presently, all the identifiers are placed at https://github.com/adobe/xdm/blob/master/schemas/context/enduserids.schema.json
But these are of type identities. There are no schema objects which contain such identifiers together as standard fields rather than as identities. So this PR introduces endUser identifiers as standard XDM fields and also an Experience Event mixin which includes these identifiers

The use case is for a solution like campaign to map its identifier fields like campaign to such standard fields (created in the PR) as part of solution's out of the box mappings , and leave it to the customer to mark those fields as identities using xdm descriptors. Some customers have requirements where they want to decide which field they want to mark as identities, rather than solution deciding for them.
This cannot be fulfilled using endUserIds schema because its already of type identity, it would not be possible to mark it as non-identity.

Plus, an attempt has been made to further divide these identifiers on the basis of their types, like personal identifiers (Email, PhoneNumber) , DeviceIdentifiers (Experience Cloud Id, Push token) and Solution identifiers (Adobe solution identifiers like campaign id, analytics id, etc).

Corresponding example files have also been made available as part of the PR.


schemas/context/deviceidentifiers.schema.json
```diff
@@ -0,0 +1,43 @@
...
+  "$id": "https://ns.adobe.com/xdm/context/deviceidentifiers",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Device identifiers",
+  "type": "object",
+  "meta:extensible": true,
+  "description": "Condensed, normalized encapsulation of device identifiers.",
+  "definitions": {
+    "deviceidentifiers": {
+      "properties": {
+        "https://ns.adobe.com/experience/ecid": {
```
lrosenthol, line 17: This is a namespace identifier not a item name...

How about `deviceID:experienceCloud`?

schemas/context/deviceidentifiers.schema.json
```diff
@@ -0,0 +1,43 @@
...
+          "title": "Adobe Experience Cloud ID",
+          "type": "string",
+          "description": "A unique identifier from Adobe Experience Cloud."
+        },
+        "https://ns.adobe.com/experience/gcmid": {
```
lrosenthol, line 22: This is a namespace identifier not a item name...

How about `deviceID:googleCloud`?

schemas/context/deviceidentifiers.schema.json
```diff
@@ -0,0 +1,43 @@
...
+          "title": "Google Cloud Messaging id",
+          "type": "string",
+          "description": "A unique identifier for Google Cloud Messaging"
+        },
+        "https://ns.adobe.com/experience/apnsid": {
```
lrosenthol, line 27: This is a namespace identifier not a item name...

How about `deviceID:applePush`?

schemas/context/personalidentifiers.schema.json
```diff
@@ -0,0 +1,38 @@
...
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/context/personalidentifiers",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Personal identifiers",
+  "type": "object",
+  "meta:extensible": true,
+  "description": "Condensed, normalized encapsulation of personal identifiers.",
+  "definitions": {
+    "personalidentifiers": {
```
lrosenthol, line 15: Aren't these defined elsewhere?  Why are we duplicating them here?

schemas/context/solutionidentifiers.schema.json
```diff
@@ -0,0 +1,53 @@
...
+  "$id": "https://ns.adobe.com/xdm/context/solutionidentifiers",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Solution identifiers",
+  "type": "object",
+  "meta:extensible": true,
+  "description": "Condensed, normalized encapsulation of solution identifiers.",
+  "definitions": {
+    "solutionidentifiers": {
+      "properties": {
+        "https://ns.adobe.com/experience/aaid": {
```
lrosenthol, line 17: See earlier comments about proper namespacing

schemas/context/deviceidentifiers.schema.json
cmathis: I think what you have here makes sense. The field name is "ecid" under the ns.adobe.com/experience namespace.

schemas/context/solutionidentifiers.schema.json
```diff
@@ -0,0 +1,53 @@
...
+          "title": "Adobe Campaign end user IDs",
+          "type": "string",
+          "description": "Adobe Campaign extension to end user IDs."
+        },
+        "https://ns.adobe.com/experience/tntid": {
+          "title": "Adobe Target end user IDs",
+          "type": "string",
+          "description": "Adobe Target extension to end user IDs."
+        },
+        "https://ns.adobe.com/experience/adcloud": {
```
cmathis, line 37: Should this be adcloudid like the rest?

schemas/context/personalidentifiers.schema.json
rajindal: These fields are there as complete schema objects like as follows : 
https://github.com/adobe/xdm/blob/master/schemas/context/emailaddress.schema.json
https://github.com/adobe/xdm/blob/master/schemas/context/phonenumber.schema.json

If they are of type string, then they are not together under one schema.

But we need these to be as of type String in one schema so that the personal identifiers makes sense . And in future, if more such personal identifier fields get introduced, they can also get introduced as standard fields in this schema.

schemas/context/solutionidentifiers.schema.json
rajindal: Yes, but this needs to be a solution identifier under experience namespace

schemas/context/solutionidentifiers.schema.json
rajindal: Yes corrected it.

rajindal:
@lrosenthol @cmathis please resume review

schemas/context/deviceidentifiers.schema.json
lrosenthol: I am still confused why we want to use a namespace URL for a property?  It's an incorrect use of NS URLs

schemas/context/personalidentifiers.schema.json
lrosenthol: Then you can't use the same name, since the name is already taken using a different format.

I would make these `xdm:emailString` and `xdm:phonenumberString` then

schemas/context/deviceidentifiers.schema.json
cmathis: @lrosenthol - I don't understand your comment here. Field namespaces can be in the form of a URI correct? 
In this case the namespace is "https://ns.adobe.com/experience/" the field name is "ecid". 

@rajindal - Because this is intended to be a standard "xdm" mixin, each of these fields should be standard "xdm" namespaced fields as well. Something like "xdm:ecId", "xdm:gcmId", etc.

rajindal: the changes have been made

schemas/context/deviceidentifiers.schema.json
rajindal: the changes have been made

schemas/context/deviceidentifiers.schema.json
rajindal: the changes have been made

schemas/context/personalidentifiers.schema.json
rajindal: Renamed the fields. The new name has no reference in the repository

schemas/context/solutionidentifiers.schema.json
rajindal: the changes have been made

rajindal:
@lrosenthol @cmathis review comments incorporated, please resume review.

schemas/context/personalidentifiers.schema.json
```diff
@@ -0,0 +1,39 @@
...
+      "properties": {
+        "xdm:emailaddressid": {
```
lrosenthol, line 17: how about just `emailid`?

rajindal: this name is already used as an object name, thanks

cmathis:
@mjasrotia and @prabhum2  - Will you look at this PR and verify that the field names are correct. Should the field names be camel case ecId, gcId, etc.

schemas/context/personalidentifiers.schema.json
cmathis: @rajindal  - Where is "emailid" used as an object name?

rajindal: @cmathis , it's used as object name in the end-user schema

cmathis:
@rajindal  - I see we have grouped each of the different identifier types into their own datatype, but we have created a single mixins which brings them all in. Do you think it would be beneficial to have three different mixins for each of the different types of identifiers. This way a user can bring in only the types they need to use. 

Also, do you every see these new datatypes being used outside of this mixin? If not, then what about moving the identifiers out of a datatype and simply define them in the mixin? Just an idea. 

Either way we are planning to have this PR included in the release into INT and Stage next week. Just a few minor details to work out as it's a pain to change later.

rajindal:
@Craig Mathis<mailto:cmathis@adobe.com> I actually created these as separate to have the flexibility of including them in different use case mixins if need be.

Like if a use case comes where someone wants to have device details with device identifiers in a mixin for some use case, then this data model will help.

So, that's why for our use case of having all identifiers in one mixin to be available as standard fields, I created the experienceevent-identifier mixin.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

cmathis:
Sounds good. So it sounds like we are free to include in this PR (or another in the future) a new device-identifiers mixin, solution-idenfiers mixin, personal-identifiers mixin, etc. 

This mixin you have included in this PR is more of a grouping all all potential identifiers. Is this right?

rajindal:
Yes, exactly, in my PR, I have grouped all the potential identifiers.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

cmathis:
Okay, we are going to make a few small modifications to this PR. 
1. Convert all the field names to camel case - emailaddressid -> emailAddressId. I know we did no have it this way in the original enduserids mixin, but given that was old and we use camel case in all current field names we will go that way here. 
2. We will create three other mixins for each of the different types of identifiers

cmathis:
@rajindal - Please remind us here why the current enduserids "https://ns.adobe.com/experience/aaid" and other fields could not be used. I know it was because they were of type "identity", but what problem did that cause?

rajindal:
@clubycoder @cmathis 
The use case is for a solution like campaign to map its identifier fields like campaign to such standard fields (created in the PR) as part of solution's out of the box mappings , and leave it to the customer to mark those fields as identities using xdm descriptors. Some customers have requirements where they want to decide which field they want to mark as identities, rather than solution deciding for them.

This cannot be fulfilled using endUserIds schema because its already of type identity, it would not be possible to mark it as non-identity.

cmathis:
The one problem with this idea is that the ecid should always have an identity namespace of "ECID", emailaddressid should always have an identity namespace of "EMAIL", etc. All of these proposed solution and device identifiers in this PR have standard identity namespace defined by the system already. It may cause confusion and issues in down stream processing if an incorrect identity namespace is used. 

I don't understand the purpose of having a standard "identifier" field not be associated with its "identity" namespace. Can you explain why you would ever do that?

rajindal:
@cmathis ,
In campaign, we want to send certain such potential identity fields like ecid, emailid, campaignid, etc as standard fields to XDM and leave it to the customer to mark those fields as identity on XDM UI using identity descriptors.
For ease of customer, we want to group these fields together so as to make a good experience for the customer, where the customer knows where all standard identifiers are, just like endUser has all them together. So that's why this PR.

I agree the downstream application can get confusion with having different namespace for different field, but I wonder how this is being handled today. the same problem can arise today with someone creating a custom field named ecid and then mark it as identity tagged with EMAIL namespace.
If not considering the custom fields, even the enduser schema where we was the namespace code and the id , there also it can be a case of someone mistakenly passing namespace code as EMAIL for ecid object.

cmathis:
I agree that a user can still send the wrong identity namespace with the current identityMap and enduserids fields so we can drop that argument. 

But we are introducing another set of "standard" fields just because Campaign wants to do something differently. The point of a "standard" model is that its used the same across all users of that model. The one and only "standard" approach being used by all existing Adobe solutions today is the identityMap. The enduserids is used in conjunction with that to support query performance, but it's never used without the identityMap being present as well. 

So what we are wanting to do in this PR is present a new "standard" which is going to be tricky. It's still a possibility, but it can't be to drive some UI behavior in Campaign.  We will have @khanna-vikas  and a few others weigh in on this subject. Probably best to schedule a meeting to talk more about it.

rajindal:
@cmathis , yes Campaign can have a UI where we present possible identity fields. But the intention was to leverage XDM descriptors . The intention was to map potential identityFields to XDM fields out of the box , and let the customer use the XDM descriptors to select identities . 
Since endUserIds cannot be used for this purpose, but a similar grouping of fields is required, from here the identifiers grouping idea came. 
The idea came to support Campaign's use case, but it seems like every solution wanting to leverage XDM descriptors this way can use these identifiers.
WDYT?

clubycoder:
There is already a standard for handling solution identifiers and ECID under `endUserIds` and and I'm seeing no reason to change that standard.  Also the cost for changing something so fundamental to our systems would be huge.  The same is true for global gross-customer identifiers like EMAIL and mobile advertising IDs.

On the point of allowing users to decide to not mark ECID or solution identifiers as identifiers, that is something we can't allow because it will break down the semantics and assumptions we rely on in the whole system.  That would also give customers a gun they don't need to shoot themselves in the foot with their implementation.

Campaign needs to follow the existing standards to work with the holistic system.

👎

cmathis:
@rajindal  - We had our normal PR review today and talked about this. For all standard identifiers (aaid, ecid, etc.) we suggest you pass those using the identityMap with the known identity namespace associated. Descriptors are to be used for custom fields where the namespace is not known. 

If there is still a reason for AEP to have a _standard identity field_ that should **not** be consumed by the identity service, then please have the PM of this project talk to the PM of identity and XDM to go over that detail.

PR #939, opened 2020-08-12
Mix-in for experience events and user reactions to prior propositions

jbeckert:
mix-in that is used for attribution of an Experience event to one or more Decision events.


extensions/adobe/experience/decisioning/proposition-response-details.schema.json
```diff
@@ -0,0 +1,65 @@
...
+  "meta:intendedToExtend": ["https://ns.adobe.com/xdm/context/experienceevent"],
+  "description": "A decision response is used to link an experience event that is a response to a prior decision/proposition events. Each decision output has a unique ID that can be echo'ed back with an experience event. The decision is assumed to have contributed to the experience event and is now attributed with by it.",
+  "definitions": {
+    "decision-response-datatype": {
+      "type": "object",
+      "properties": {
+        "xdm:decisionEventId": {
+          "type": "string",
+          "title": "Decision Event Identifier",
+          "description": "The unique identifier from the decision event. For a given decision request and profile the output of the decision is a unique event. If the same decision is made again for the same profile and the output are the same options, it is ist considered a different proposition with a unique propositionID. See https://ns.adobe.com/experience/decisioning/proposition."
```
lrosenthol, line 23: spelling issue - `it is ist`

jbeckert: Corrected. Thanks @lrosenthol for spotting it.

extensions/adobe/experience/decisioning/proposition-response-details.example.1.json
```diff
@@ -0,0 +1,28 @@
+{
+  "https://ns.adobe.com/experience/decisioning/responseDetails": [
+    {
+      "xdm:decisionEventId": "TNT:activity1:experience1:1597208433",
+      "xdm:eventContext": {
+        "xdm:scope": "demo-box",
+        "xdm:activityID": "actitivy1"
```
jonsnyder, line 7: typo: "actitivy1" => "activity1"

jbeckert: Fixed. Thanks Jon. I did move the custom extended eventContext to another example. (proposition-response-details.example.2.json)

extensions/adobe/experience/decisioning/proposition-response-details.schema.json
```diff
@@ -0,0 +1,65 @@
...
+  "type": "object",
+  "meta:abstract": true,
+  "meta:extensible": true,
+  "meta:intendedToExtend": ["https://ns.adobe.com/xdm/context/experienceevent"],
+  "description": "A decision response is used to link an experience event that is a response to a prior decision/proposition events. Each decision output has a unique ID that can be echo'ed back with an experience event. The decision is assumed to have contributed to the experience event and is now attributed with by it.",
+  "definitions": {
+    "decision-response-datatype": {
+      "type": "object",
+      "properties": {
+        "xdm:decisionEventId": {
```
cmathis, line 20: @jbeckert  - I see we have namespaced this field under "xdm". Given the location and purpose of this field, should this field be named "https://ns.adobe.com/experience/decisioning/decisionEventId" so its associated with the "experience/decisioning" namespace?

This same comment would apply the the "xdm:eventContext" below.

jbeckert: it is a local field inside a mix-in structure that has the namespace treatment. Have a look at the examples for a better illustration of this.

cmathis: Yes, I see this and it's something we did in the past because of the XED conversion to avoid the duplicate namespaces in the physical path. The XED conversion no longer duplicates so it's technically not necessary. We can leave it as it this time, but note in the future we need to start having vendor/solution mixins declare new fields in their own namespace. If it's an "xdm" field then that should be declared inside a standard "xdm" mixin and not one from a solution/vendor.

cmathis:
@cluby-adobe - Please review as this involves ExperienceEvent

extensions/adobe/experience/decisioning/proposition-response-details.schema.json
```diff
@@ -0,0 +1,65 @@
...
+            "xdm:scope": {
+              "type": "string",
+              "title": "Decision Event Scope",
+              "description": "The scope identifier that was used as an input for the decision request. This identifier selects the locatation or placement for which the decision was requested, plus any other static context used."
+            }
+          }
+        }
+      }
+    },
+    "decision-response-mixin": {
```
mjasrotia, line 39: This seems like a valid JSON schema, but we always create separate JSON schema files for the XDM constructs to compose a schema like class, mixin and data types. I would suggest to take the "decision-response-mixin" definition out of this file and create another json schema for the MIXIN. The current file can stay as data type.

jbeckert: I split the schema components into two separate files now.

extensions/adobe/experience/decisioning/proposition-attribution-detail.schema.json
```diff
@@ -0,0 +1,40 @@
+{
+  "meta:license": [
+    "Copyright 2019 Adobe Systems Incorporated. All rights reserved.",
```
mjasrotia, line 3: Copyright should be 2020

extensions/adobe/experience/decisioning/proposition-attribution-detail.schema.json
```diff
@@ -0,0 +1,40 @@
...
+  ],
+  "$id": "https://ns.adobe.com/experience/decisioning/proposition-attribution-detail",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Proposition Acceptance Details",
+  "type": "object",
+  "meta:abstract": true,
+  "meta:extensible": true,
+  "description": "A decision response is used to link an experience event that is a response to a prior decision/proposition events. Each decision output has a unique ID that can be echo'ed back with an experience event. The decision is assumed to have contributed to the experience event and is now attributed with by it.",
+  "definitions": {
+    "proposition-attribution-detail-datatype": {
```
mjasrotia, line 16: We don't put names like "class", "mixin", "datatype" in the schema/definition names. Please remove "-datatype". The same applies to the definition name in the mixin as well.

extensions/adobe/experience/decisioning/proposition-attribution-detail.schema.json
```diff
@@ -0,0 +1,40 @@
...
+      "type": "object",
+      "properties": {
+        "xdm:decisionEventId": {
```
mjasrotia, line 19: Id are named as "ID". Please change it to "decisionEventID". Same applies to all Ids

extensions/adobe/experience/decisioning/proposition-attribution.schema.json
```diff
@@ -0,0 +1,43 @@
...
+  "type": "object",
+  "meta:abstract": true,
+  "meta:extensible": true,
+  "meta:intendedToExtend": ["https://ns.adobe.com/xdm/context/experienceevent"],
+  "description": "A decision attribution mixin is used to link an experience event that is a response to one or more prior decision/proposition events. Each decision output has a unique ID that can be echo'ed back with an experience event. The decision is assumed to have contributed to the experience event and is now attributed with by it.",
+  "definitions": {
+    "proposition-attribution-mixin": {
+      "type": "object",
+      "properties": {
+        "https://ns.adobe.com/experience/decisioning/attibution": {
```
mjasrotia, line 20: "attribution" Please fix the typo

extensions/adobe/experience/decisioning/proposition-attribution.schema.json
```diff
@@ -0,0 +1,43 @@
+{
+  "meta:license": [
+    "Copyright 2019 Adobe Systems Incorporated. All rights reserved.",
```
mjasrotia, line 3: 2020

extensions/adobe/experience/decisioning/proposition-attribution-detail.schema.json
jbeckert: Corrected.

extensions/adobe/experience/decisioning/proposition-attribution-detail.schema.json
jbeckert: Corrected.

extensions/adobe/experience/decisioning/proposition-attribution.schema.json
jbeckert: Corrected.

extensions/adobe/experience/decisioning/proposition-attribution.schema.json
jbeckert: Corrected.

extensions/adobe/experience/decisioning/proposition-attribution-detail.schema.json
jbeckert: All schemas in this namespace use this convention, mostly for internal documentation help. None of the JSON definition names will contribute to the composed schema, they will be entirely invisible.
How necessary is it that this is being changed now. It would cause an inconsistency with the other files in this namespace

jbeckert:
The use cases that would need this mix-in are still being detailed.

extensions/adobe/experience/decisioning/proposition-attribution.example.1.json
```diff
@@ -0,0 +1,12 @@
+{
+  "https://ns.adobe.com/experience/decisioning/attribution": [
+    {
+      "xdm:decisionEventID": "TNT:activity1:experience1:1597208433",
+      "xdm:decisionScope": "__view__"
+    },
+    {
+      "xdm:decisionEventID": "3cc33a7e-13ca-4b19-b25d-c816eff9a70a",
+      "xdm:decisionScope": "OD:eyJhY3Rpdml0eUlkIjoieGNvcmU6b2ZmZXItYWN0aXZpdHk6MTFjZmIxZmE5MzM4MWFjYSIsInBsYWNlbWVudElkIjoieGNvcmU6b2ZmZXItcGxhY2VtZW50OjExNzUwMDk2MTJiMDEwMGMifQ=="
```
catacgc, line 9: What's the rationale for this to be base64 encoded? What's the inteneded downstream use?

```
{"activityId":"xcore:offer-activity:11cfb1fa93381aca","placementId":"xcore:offer-placement:1175009612b0100c"}
```

extensions/adobe/experience/decisioning/proposition-attribution-detail.schema.json
scottsimmer: Agree please leave them for this branch I find them helpful and non-intrusive or impacting

prabhum2:
@jbeckert - Please let me know when this is ready for formal xdm review

jbeckert:
@prabhum2 This can be reviewed for the mechanical aspects, the structure, naming, style guidelines etc.
We still have an outstanding discussion for how one of the clients of the mixing will use it but if you consider it to be used by non-Adobe solution clients it is general enough to fit the needs of sending back reactions and responses to prior propositions made in the experience decisioning sub domain. At that abstract level, everything is there.

extensions/adobe/experience/decisioning/proposition-attribution.example.1.json
jbeckert: This is absolutely **not** a requirement mandated by the schema that is proposed here. From a schema perspective it is string. It is not even an identifier. The semantic the mixin only says to use the string called "scope" that was received with the decision. It could have been a string containing a placement identifier.

jbeckert: I changed the example to avoid internal details such as base64 encoded JSON strings for one implementation. The public documentation should not manifest those temporary details

extensions/adobe/experience/decisioning/proposition-attribution.schema.json
```diff
@@ -0,0 +1,43 @@
+{
```
clubycoder, line 1: @jbeckert , Once this gets realigned to provide a core mixin for experienceevent, this mixin should probably be moved to - https://github.com/adobe/xdm/tree/master/components/mixins/experience-event - and renamed to match the other experienceevent mixins for consistency.

clubycoder (changes requested):
Added another specific request to move the experiencevent mixin.

jbeckert:
@clubycoder The decisioning domain has always been an adobe extension. Can we address this in a different PR and timeframe. The current mix-in works in and with the datatype schemas of that extension domain. For instance, it is not common practice that a core component makes references to an extension component. The decisioning domain defines the proposition event and decision options class in the first place and the new mix-in makes reference to those schemas. It would be questionable if this is even allowed, that a core component references those extensions. It is the other way around. Making the mixin completely generic and not refer to any decision proposition or decision options, robs it of its value because anything could be stored in the mixin properties.
There was resistance in the past by the stewards of the XDM system to pull in decisioning as a core sub-domain. This should be done holistically and not by moving one mixin of the domain to the core  system

If the term "attribution" is reserved for other concerns, then this is easy to change. 
What about "response", "reaction", or "interaction", "lifecycle"? 
The issue I see with the first two is that there are a number of Experience Events that will use the mix-in which aren't strictly user _responses_ or _reactions_. Like the event that presents (shows, displays, renders) the decision proposition. 
I would favor lifecycle but could also see interaction as a good term.

clubycoder (approved):
👍  once we get confirmation from the Target team that - https://github.com/adobe/xdm/pull/939/files#diff-8baf7af232216429819417fd0b292424R17 - going the right direction.

PR #942, opened 2020-08-17
Deprecation of current Browser/OS IDs for string based ones.

jbeecher:
Change the current Browser and OS IDs from a signed integer to string. This is needed because the Analytics data is actually an unsigned int for these values which causes some of the values imported to appear as negative.

### DIFF
--- a/extensions/adobe/experience/analytics/experienceevent-all.schema.json
+++ b/extensions/adobe/experience/analytics/experienceevent-all.schema.json
@@ -7219,12 +7219,24 @@
           "properties": {
             "xdm:browserID": {
               "title": "Browser ID",
+              "meta:status": "deprecated",
               "type": "integer",
-              "description": "The Adobe Analytics identifier for the browser used."
+              "description": "The Adobe Analytics identifier for the browser used. Deprecated, use xdm:browserIDStr instead."
             },
             "xdm:operatingSystemID": {
               "title": "Operating System ID",
+              "meta:status": "deprecated",
               "type": "integer",
+              "description": "The Adobe Analytics identifier of the operating system used. Deprecated, use xdm:operatingSystemIDStr instead."
+            },
+            "xdm:browserIDStr": {
+              "title": "Browser ID",
+              "type": "string",
+              "description": "The Adobe Analytics identifier for the browser used."
+            },
+            "xdm:operatingSystemIDStr": {
+              "title": "Operating System ID",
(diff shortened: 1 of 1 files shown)


extensions/adobe/experience/analytics/experienceevent-all.schema.json
```diff
@@ -7219,12 +7219,24 @@
...
             },
             "xdm:operatingSystemID": {
               "title": "Operating System ID",
+              "meta:status": "deprecated",
               "type": "integer",
+              "description": "The Adobe Analytics identifier of the operating system used. Deprecated, use xdm:operatingSystemIDStr instead."
+            },
+            "xdm:browserIDStr": {
+              "title": "Browser ID",
+              "type": "string",
```
lrosenthol, line 7234: is there a validation pattern for this?  For example, is it all nums?

extensions/adobe/experience/analytics/experienceevent-all.schema.json
```diff
@@ -7219,12 +7219,24 @@
...
+              "description": "The Adobe Analytics identifier for the browser used."
+            },
+            "xdm:operatingSystemIDStr": {
+              "title": "Operating System ID",
+              "type": "string",
```
lrosenthol, line 7239: is there a validation pattern for this?  For example, is it all nums?

extensions/adobe/experience/analytics/experienceevent-all.schema.json
jbeecher: We're currently not planning on doing validation within XDM for these fields.  The semantics for the values will be kept within the Analytics Data Connector and matching the separate, matching lookup tables we'll be pushing to AEP.

lrosenthol: *you* may not be doing validation - but *someone* may want to at some point...

So assuming that they match a pattern, please provide.

jbeecher: Provided

extensions/adobe/experience/analytics/experienceevent-all.schema.json
jbeecher: Added

cmathis:
@clubycoder  - Can you review and approve this PR

PR #955, opened 2020-08-28
B2b account

mjasrotia:
Changed the B2B Account schemas as per recent feedback from MARKETO


components/mixins/account/account-details.schema.json
```diff
@@ -0,0 +1,103 @@
...
+  "title": "Account Details",
+  "type": "object",
+  "meta:extensible": true,
+  "meta:abstract": true,
+  "meta:intendedToExtend": ["https://ns.adobe.com/xdm/class/account"],
+  "description": "An XDM Business Account.",
+  "definitions": {
+    "account-details": {
+      "properties": {
+        "xdm:accountComponents": {
```
cmathis, line 19: If this is an array of b2b-account objects, then why not have the "items" simply be a $ref to that datatype? Is there a reason to duplicate those things here?

mjasrotia: I also thought to refer that initially but I believe, there are still few fields which are not included in the array that exist in the "b2b-account" data type. Let me double check again.

mjasrotia: I think, we can refer that data type in the mixin. I just checked. Thanks for pointing this out. I will make the change.

components/datatypes/account-organization.schema.json
```diff
@@ -0,0 +1,131 @@
...
+          "title": "NAICS Description",
+          "description": "A brief description of an organization’s line of business, based on its NAICS code.",
+          "type": "string"
+        },
+        "xdm:SICDescription": {
+          "title": "SIC Description",
+          "description": "A brief description of an organization’s line of business, based on its SIC code.",
+          "type": "string"
+        },
+        "xdm:D-U-N-S-Number": {
```
cmathis, line 103: Cannot use dashes in AEP

components/classes/account.schema.json
```diff
@@ -5,7 +5,7 @@
     "you may not use this file except in compliance with the License. You may obtain a copy",
     "of the License at https://creativecommons.org/licenses/by/4.0/"
   ],
-  "$id": "https://ns.adobe.com/xdm/context/account",
+  "$id": "https://ns.adobe.com/xdm/class/account",
```
cmathis, line 8: You cannot change the $id otherwise you will break any existing schemas that referenced the old $id

mjasrotia: None of the account schemas are in use yet. All are experimental and we are the only ones changes these schemas.

mjasrotia: Will change

components/datatypes/account-organization.schema.json
mjasrotia: Will remove Dashes

cmathis (approved):

PR #957, opened 2020-08-31
Privacy/Personalization/Marketing Preferences

RussStringham:
https://jira.corp.adobe.com/browse/PLAT-68848


components/datatypes/consentpreferences.schema.json
```diff
@@ -0,0 +1,337 @@
...
+          "description": "Timestamp of the complete set of user consent preferences",
+          "type": "string",
+          "format": "date-time"
+        },
+        "xdm:version": {
+          "title": "Preferences Version",
+          "description": "Version of the Privacy Preferences Standard",
+          "type": "string"
+        },
+        "xdm:userLocale": {
```
clubycoder, line 299: @RussStringham , Can you clarify what this field is for and how it relates or doesn't with the existing locale information? - https://github.com/adobe/xdm/blob/master/components/datatypes/placecontext.schema.json

RussStringham: This is not related to language, but legal jurisdiction. If set to EU or a specific country in the UK, then GDPR would apply. If set to California (CA) then CCPA would apply. If set to US then CCPA might apply.

RussStringham: Should the name of this field be changed from xdm:userLocale to xdm:jurisdiction?

components/datatypes/consentpreferences.schema.json
```diff
@@ -0,0 +1,337 @@
...
+      "meta:enum": {
+        "consent": "User Consent",
+        "legitimate_interest": "Legitimate Interest",
+        "contract": "Contract",
+        "compliance": "Compliance with a Legal Obligation",
+        "vital_interest": "Vital Interest of the Individual",
+        "public_interest": "Public Interest"
+      }
+    },
+    "timestamp": {
```
lrosenthol, line 56: you don't need this - we already have a nice definition...

components/datatypes/consentpreferences.schema.json
```diff
@@ -0,0 +1,337 @@
...
+      "type": "object",
+      "properties": {
+        "xdm:choice": {
+          "$ref": "#/definitions/consentValue"
+        },
+        "xdm:basisOfProcessing": {
+          "$ref": "#/definitions/basisOfProcessing"
+        },
+        "xdm:timestamp": {
+          "$ref": "#/definitions/timestamp"
```
lrosenthol, line 72: use `dc:date` instead

components/datatypes/consentpreferences.schema.json
```diff
@@ -0,0 +1,337 @@
...
```
lrosenthol, line 86: see above...

components/datatypes/consentpreferences.schema.json
```diff
@@ -0,0 +1,337 @@
...
+        },
+        "xdm:reason": {
```
lrosenthol, line 88: should it be outOutReason?

components/datatypes/consentpreferences.schema.json
```diff
@@ -0,0 +1,337 @@
...
+              "$ref": "#/definitions/consentOption"
+            }
+          }
+        },
+        "xdm:personalizationPreferences": {
+          "title": "Personalization Preferences",
+          "description": "User's Personalization Preferences",
+          "type": "object",
+          "properties": {
+            "xdm:general": {
```
lrosenthol, line 134: this is really too "general" a name...

components/datatypes/consentpreferences.schema.json
```diff
@@ -0,0 +1,337 @@
...
+              "title": "General Personalization Preference",
+              "description": "Is there consent for personalization, unless overridden by a specific personalization preference",
+              "$ref": "#/definitions/consentOption"
+            },
+            "xdm:content": {
```
lrosenthol, line 139: this is really too "general" a name...

components/datatypes/consentpreferences.schema.json
```diff
@@ -0,0 +1,337 @@
...
+              "title": "Personalize in-app messages",
+              "description": "Allow personalization of messages displayed within my apps",
+              "$ref": "#/definitions/consentOption"
+            },
+            "xdm:offers": {
+              "title": "Personalize Offers",
+              "description": "Allow personalization of offers (my own ads) displayed on my site or in my app",
+              "$ref": "#/definitions/consentOption"
+            },
+            "xdm:eMail": {
```
lrosenthol, line 154: `email` is always all lowercase...and it's also used elsewhere in XDM.  Pick another term.

components/datatypes/consentpreferences.schema.json
```diff
@@ -0,0 +1,337 @@
...
+              "title": "Personalize eMail",
+              "description": "Allow personalization of emails sent to the user",
+              "$ref": "#/definitions/consentOption"
+            },
+            "xdm:snailMail": {
```
lrosenthol, line 159: is this really an appropriate name?  it's a derogatory term, I wouldn't use it.

components/datatypes/consentpreferences.schema.json
```diff
@@ -0,0 +1,337 @@
...
+              "title": "Personalize in-vehicle",
+              "description": "Personalize in-vehicle experiences",
+              "$ref": "#/definitions/consentOption"
+            },
+            "xdm:inHome": {
+              "title": "Personalize in-home",
+              "description": "Personalize in-home experiences",
+              "$ref": "#/definitions/consentOption"
+            },
+            "xdm:iot": {
```
lrosenthol, line 199: `xdm:iotDevices`

components/datatypes/consentpreferences.schema.json
```diff
@@ -0,0 +1,337 @@
...
+              "title": "Personalize ",
+              "description": "Personalize third-party offers/ads displayed on my site or in my app",
+              "$ref": "#/definitions/consentOption"
+            },
+            "xdm:thirdPartyContent": {
+              "title": "Personalize ",
+              "description": "Personalize third-party content displayed on my site or in my app",
+              "$ref": "#/definitions/consentOption"
+            },
+            "xdm:ads": {
```
lrosenthol, line 219: advertising?  advertisements?

components/datatypes/consentpreferences.schema.json
```diff
@@ -0,0 +1,337 @@
...
+              "$ref": "#/definitions/consentOption"
+            }
+          }
+        },
+        "xdm:marketingPreferences": {
+          "title": "Marketing Preferences",
+          "description": "User's Direct Marketing Preferences",
+          "type": "object",
+          "properties": {
+            "xdm:general": {
```
lrosenthol, line 231: see previous comment

components/datatypes/consentpreferences.schema.json
```diff
@@ -0,0 +1,337 @@
...
+              "title": "General Direct Marketing Preference",
+              "description": "Does user consent to Direct Marketing in general, unless overridden by specific choices",
+              "$ref": "#/definitions/marketingOption"
+            },
+            "xdm:eMail": {
```
lrosenthol, line 236: see previous comment

components/datatypes/consentpreferences.schema.json
```diff
@@ -0,0 +1,337 @@
...
+              "title": "Receive SMS",
+              "description": "It is okay to send text messages to the user",
+              "$ref": "#/definitions/marketingOption"
+            },
+            "xdm:phoneCalls": {
+              "title": "Receive phone calls",
+              "description": "The company may contact the user with phone solicitations",
+              "$ref": "#/definitions/marketingOption"
+            },
+            "xdm:snailMail": {
```
lrosenthol, line 261: see previous comment

components/datatypes/consentpreferences.schema.json
```diff
@@ -0,0 +1,337 @@
...
+              "title": "Receive in-vehicle messages",
+              "description": "It is okay to provide offers to the user via technology installed in their vehicle",
+              "$ref": "#/definitions/marketingOption"
+            },
+            "xdm:inHomeMessages": {
+              "title": "Receive in-home messages",
+              "description": "It is okay to provide offers to the user via technology installed in their home",
+              "$ref": "#/definitions/marketingOption"
+            },
+            "xdm:iot": {
```
lrosenthol, line 276: see previous comment

components/datatypes/consentpreferences.schema.json
```diff
@@ -0,0 +1,337 @@
...
+              "$ref": "#/definitions/marketingOption"
+            },
+            "xdm:socialMedia": {
+              "title": "Receive Social Media Messages",
+              "description": "It is okay to direct messages to the user via their social media accounts",
+              "$ref": "#/definitions/marketingOption"
+            }
+          }
+        },
+        "xdm:timestamp": {
```
lrosenthol, line 288: see previous comment

components/datatypes/consentpreferences.schema.json
```diff
@@ -0,0 +1,337 @@
...
+          "title": "Consent Preferences timestamp",
+          "description": "Timestamp of the complete set of user consent preferences",
+          "type": "string",
+          "format": "date-time"
+        },
+        "xdm:version": {
+          "title": "Preferences Version",
+          "description": "Version of the Privacy Preferences Standard",
+          "type": "string"
```
lrosenthol, line 297: this must have a pattern that defines it - please include for validation

components/datatypes/consentpreferences.schema.json
```diff
@@ -0,0 +1,337 @@
...
+        },
+        "xdm:userLocale": {
+          "title": "User Locale",
+          "description": "User location or jurisdiction that applies to the consents for this user",
+          "type": "string"
+        },
+        "xdm:localeSource": {
```
lrosenthol, line 304: this will also need to change names inline with `jurisdiction` chane

components/datatypes/consentpreferences.schema.json
RussStringham: The use case we have right now applies for the reason that a user opted out. However, we didn't want to preclude a future use case where they may be able to specify the reason that the user opted in. We can change the title and description in the future without breaking the schema, but if we change the name of the field to optOutReason, then we cannot change it without breaking the schema.

components/datatypes/consentpreferences.schema.json
RussStringham: Previously I called this "default" but that was deemed even worse. I could call it "generalPersonalization", but that seemed redundant since it is inside the personalization preferences section. Do you have another suggestion?

components/datatypes/consentpreferences.schema.json
RussStringham: I have changed this to lower case

components/datatypes/consentpreferences.schema.json
RussStringham: What would you use instead? traditionalMail, physicalMail?

components/datatypes/consentpreferences.schema.json
clubycoder: I like the idea of changing the name to reference a jurisdiction and being more concrete.  I would also change it to an enum with the jurisdictions we understand.

components/datatypes/consentpreferences.schema.json
RussStringham: I've changed this to "advertising"

components/datatypes/consentpreferences.schema.json
RussStringham: @lrosenthol  where is dc:date defined? How do I use it?

lrosenthol: `dc:date` is one of the standard XMP values - it means a generic date and/or time.

components/datatypes/consentpreferences.schema.json
lrosenthol: physicalMail sounds good to me

components/datatypes/consentpreferences.schema.json
lrosenthol: OK - I'll live with it

components/datatypes/consentpreferences.schema.json
cmathis: @lrosenthol  - Can you provide a link here to the standard definition of all the dc:* and repo:* fields that are used in XDM?

cmathis:
@RussStringham  - Are there still open questions in this PR? Can you provide a list of all the stake holders who should review and approve this PR?

RussStringham:
I'm still willing to incorporate dc:date if @lrosenthol can provide details or examples of how to use it.
CJM team has already approved PR.
Bikas Saha and Sandeep Nayak on AEP architecture team need to approve and have been providing feedback in other channels.
Someone on UPS team needs to approve (working with Mohit Garg, Guanghong Yang, Neeraj Goel, Bhavin Doshi, ...). We have had several meetings and email threads.
Someone on AEP/PQS/Data Lake team - meeting for first time later this week. Concerns not anticipated from this team.
Someone on XDM team

components/datatypes/consentpreferences.schema.json
```diff
@@ -0,0 +1,410 @@
...
+        "no": "The user has explicitly opted out of this feature",
+        "pending": "Pending verification",
+        "unknown":"Unknown",
+        "not_applicable":"Not Applicable"
+      }
+    },
+    "basisOfProcessing": {
+      "title": "Basis Of Processing",
+      "type": "string",
+      "description": "Basis of Processing",
```
jonsnyder, line 38: A better description here would be helpful.

components/datatypes/consentpreferences.schema.json
```diff
@@ -0,0 +1,410 @@
...
+              "title": "Pseudonymous Analysis",
+              "description": "Data can be used for analysis that may include an ID that may or may not be tied to an actual person",
+              "$ref": "#/definitions/consentOption"
+            },
+            "xdm:deviceLinking": {
+              "title": "Device Linking",
+              "description": "Attempts to associate multiple devices used/owned by an individual/household/organization are permitted",
+              "$ref": "#/definitions/consentOption"
+            },
+            "xdm:canUseAdvertiserID": {
```
jonsnyder, line 145: To be consistent with the other field names here, I would call this "useAdvertiserID". The "can do"/"can use" is implied.

jonsnyder (approved):

components/mixins/profile/profile-consent.schema.json
```diff
@@ -0,0 +1,36 @@
...
+  "$id": "https://ns.adobe.com/xdm/mixin/profile-consent",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Privacy/Marketing Preferences (Consent)",
+  "description": "This schema captures privacy, personalization and marketing preferences (consents).",
+  "type": "object",
+  "meta:extensible": true,
+  "meta:abstract": true,
+  "meta:intendedToExtend": ["https://ns.adobe.com/xdm/context/profile"],
+  "definitions": {
+    "profile-consents": {
```
neerajgoel82, line 17: I see a field called "source" in "consentMetadata" which indicates we may want to store the consent information from different sources but I am not sure how we will do that with this structure.

RussStringham: Each of the 36 consents has its own source field. However, for a given profile fragment (or experience event if we include consents in them) the values will generally be identical. Rather than specify the same source value (or timestamp) 36 times, simply specify it once in the metadata section and it applies to all.

neerajgoel82: So, what you are saying is that generally each source will have its own dataset id and hence will fall in different fragments. I guess that makes sense.

components/datatypes/consentpreferences.schema.json
```diff
@@ -0,0 +1,424 @@
...
+              "title": "Data Collection",
+              "description": "Data collection is permitted",
+              "$ref": "#/definitions/consentOption"
+            },
+            "xdm:sellData": {
+              "title": "Sell Data",
+              "description": "Selling of user's Data with/to 2nd or 3rd parties is permitted, such as in compliance with CCPA",
+              "$ref": "#/definitions/consentOption"
+            },
+            "xdm:shareData": {
```
bikas-adobe, line 125: @RussStringham  - Does this need to account for second party vs third party sharing?

components/mixins/profile/profile-consent.example.1.json
```diff
@@ -0,0 +1,52 @@
...
+        "xdm:iotMessages": {
+          "xdm:choice": "yes",
+          "xdm:timestamp": "2019-01-01T15:52:25+00:00",
+          "xdm:basisOfProcessing": "legitimate_interest"
+        },
+        "xdm:pushNotifications": {
+          "xdm:choice": "yes",
+          "xdm:source": "OurApp"
+        },
+        "xdm:eMail": {
```
aechow, line 38: is there a reason for using eMail vs email?

components/datatypes/consentpreferences.schema.json
```diff
@@ -0,0 +1,424 @@
...
+          "maxLength": 20
+        },
+        "xdm:userCountry": {
+          "title": "User Country/Region",
+          "description": "ISO 3166 alpha2 country code or ISO 3166-2 country plus state/province code for use in helping to determine how to interpret consent values",
+          "type": "string",
+          "maxLength": 6,
+          "pattern": "[A-Z]{2}(-[A-Z]{1,3}){0,1}"
+        },
+        "xdm:countrySource": {
```
bikas-adobe, line 381: Will this account for regions within the same country (say CA) or say regions across countries (say EU) in case they have different consent regulations? If yes, then should we rename this to reflect that?

components/mixins/profile/profile-consent.example.1.json
guayang-adobe: @RussStringham let's use all lowecase "email" here to be consistent with other places like preferredChannel. This key is also important for CJM filtering

components/datatypes/consentpreferences.schema.json
RussStringham: end users don't understand the difference between the two types of sharing, so it would be difficult to ask a question that could distinguish between the two. Also, CCPA doesn't distinguish.

components/mixins/profile/profile-consent.example.1.json
RussStringham: this is a typo, should be xdm:email

components/datatypes/consentpreferences.schema.json
RussStringham: ISO 3166 alpha2 country code is a two character code identifying the country.
ISO 3166-2 adds a 1 to 3 character code to identify regions, states, provinces (and in some cases cities) within the country. For the US it is a 2 character state code. 
So customer can use "US" if they will apply the same regulations for all US customers, or "US-CA", "US-NY", ... if they plan to treat California customers differently than others in the US, because of California's stricter regulations. Or they can use "US-CA", ... and only look at the first two characters to treat all users in the US the same.
Do you have a suggestion for an alternate name? userCountryRegion? userCountryOptionalRegion? Country is accurate, because it always identifies the country, but sometimes it can be more precise. userCountryPlus?

components/mixins/profile/profile-privacy.schema.json
```diff
@@ -18,12 +18,12 @@
       "properties": {
         "xdm:optOutConsentLevel": {
           "type": "object",
-          "title": "Opt-out consent level",
-          "description": "Opt-out consent level.",
+          "title": "Opt-out consent level (deprecated)",
+          "description": "Opt-out consent level (deprecated).",
```
cmathis, line 22: If this is to be deprecated, then add the "meta:status" : "deprecated" attribute here. This will make it so the field no longer appears in the UI (or shouldn't).

components/datatypes/consentpreferences.schema.json
```diff
@@ -0,0 +1,424 @@
...
+  },
+  "allOf":[
+    {
+      "$ref":"https://ns.adobe.com/xdm/common/extensible#/definitions/@context"
+    },
+    {
+      "$ref": "#/definitions/consent-preferences"
+    }
+  ],
+  "meta:status":"experimental"
```
cmathis, line 423: Change this value to "stable" if you want this to show on Stage and Prod otherwise it will be INT only.

components/datatypes/consentpreferences.schema.json
bikas-adobe: Do we need EU, EU-Italy, EU-Italy-Tuscany?

components/datatypes/consentpreferences.schema.json
bikas-adobe: I distinctly remember recently responding to a privacy notice where it asked if it was ok for my data to be shared with affiliates within the company (2P) and outside the company (3P). They were different choices. Could you please check with legal/PM on this. While the definition of 2P/3P may be less relevant here - the important question whether users can tell Unilever that they cannot share data with Amex vs cannot share data within Unilever brand companies.

RussStringham: That's not the definition of 2nd versus 3rd party. 2nd party is where the company shares the data directly with another entity (affiliate or other company). 3rd party is where you share with an intermediary and they share with other companies, or more generally where the 2nd party shares the data shared with them with others.

RussStringham: You are correct that banking regulations restrict some 1st-party sharing (including affiliates) without permission, such as details about your checking account being shared with the credit-card division for sending offers for new credit cards based on how much money moves through your checking account.
I'll talk to the PM about adding this use case.

bikas-adobe: Sure. 2P vs 3P distinction is not my main concern. The important question is whether users can tell Unilever that they cannot share data with Amex vs cannot share data within Unilever brand companies. I have seen multiple privacy data sharing forms that make a distinction between these 2 kinds of sharing and I am wondering if that applies to AEP consent collection.

components/datatypes/consentpreferences.schema.json
RussStringham: ISO 3166 does not include "EU" as a country code. To see if GDPR applies you need to check if this field contains the country code for any country in the EU. Note that the code for Tuscany is "IT-52". I will update the pattern to support digits for the region portion of the code.

RussStringham: I am also renaming this field to userCountryRegionCode

components/datatypes/consentpreferences.schema.json
RussStringham: My thoughts: If the different Unilever brands use different IMS Orgs for collecting data, then the xdm:dataSharing consent would apply, just as it does for 2nd/3rd-parties. With demand, we could add a new consent to distinguish between these types of sharing.
If they are part of the same IMS Org, then we would need significant changes to all products to keep the data separate for user's who make this choice. For example, for every query we would need to know the organization making the query, the organization responsible for the experience event or profile attribute and whether sharing was allowed between the two.

components/mixins/profile/profile-privacy.schema.json
```diff
@@ -18,12 +18,13 @@
...
+          "meta:status": "deprecated",
           "properties": {
             "xdm:privacyOptOuts": {
-              "title": "Privacy opt-outs",
-              "description": "Encapsulates different opt-out types for data privacy.",
+              "title": "Privacy opt-outs (deprecated)",
+              "description": "Encapsulates different opt-out types for data privacy (deprecated).",
```
cmathis, line 27: Same deprecated comment here. Any field you want no longer used should have the "meta:status": "deprecated" attribute added.

cmathis:
@RussStringham  - Can you also fix the json-lint errors on this PR?

components/mixins/profile/profile-privacy.schema.json
cmathis: Also, is it correct that this mixin should stay around so a user can use the "identityPrivacyInfo" field? Or should this whole mixin be deprecated?

RussStringham: This field is an element of another field, optOutConsentLevel, that I marked as deprecated. Do I need to mark all of its elements deprecated as well?

The RTCDP product is using the identityPrivacyInfo field. That is why I am not deprecating the whole file.

aechow:
Just wondering, is there a reason we use snake case in the enums for channels and camel case for the xdm keys? e.g. push_notifications vs xdm:pushNotifications

lrosenthol:
@aechow I don't know anything about these "channels" you speak of - but XDM keys are based on XMP (SO 16684), which uses camel case.

RussStringham:
@aechow @lrosenthol I use snake_case for enum values, wherever enums are used, because that is the format I saw being used in other places. If enum values should use camelCase, I can make that change.

components/datatypes/consentpreferences.schema.json
bikas-adobe: Sounds good

RussStringham:
grep-ing the code, I see snake_case, camelCase and PascalCase all used for enums.

components/datatypes/consentpreferences.schema.json
bikas-adobe: Ok. We can add more cases when the need arise to differentiate between different kinds of sharing specifically. The current value can be understood as allow all kinds of sharing.

cmathis (approved):

PR #958, opened 2020-09-01
Adding search annotations for metadata for Windrose

vpetcu:
Project Windrose plans to use these metadata and an important aspect is being able to search based on them.
This commit is marking some metadata for search indexing.
It also deprecates xdm:fabrication in favor of xdm:material and schema:description in favor of xdm:productDescription.

### DIFF
--- a/components/datatypes/product.schema.json
+++ b/components/datatypes/product.schema.json
@@ -22,7 +22,13 @@
         "xdm:SKU": {
           "title": "SKU",
           "type": "string",
-          "description": "The unique SKU (stock keeping unit) of the product assigned by the vendor."
+          "description": "The unique SKU (stock keeping unit) of the product assigned by the vendor.",
+          "meta:descriptors": [
+            {
+              "@type": "xdm:searchdescriptor",
+              "search:indexed": true
+            }
+          ]
         },
         "xdm:name": {
           "title": "Name",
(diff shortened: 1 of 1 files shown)


vpetcu:
@fmeschbe @lrosenthol @cmathis I opened another PR using the new github structure. Could you please have a look?

cmathis:
Is "Project Windrose" a new service on AEP?

vpetcu:
@cmathis Windrose is the rebranded name for DDAM. It's an UI service for asset management using AEP-CS as backend.

cmathis:
@vpetcu  -  I don't think the AEP-CS system used this XDM repo anymore. They converted over to an internal XCM set of schemas.

vpetcu:
@cmathis that's true. They have an internal XCM git repo with schemas.
But this has been discussed on the #content-arch slack channel and the conclusion was that they will deploy schemas from the XDM git repo as well, when needed.

cmathis:
Has this Product schema been copied to the internal XCM repo? If so, please update there instead. If not, then give us a few days to make sure this does affect AEP in a negative way.

vpetcu:
@cmathis this schema will not be copied over to the XCM repo. They wanted to avoid code duplication which makes sense.
AFAIU, they will instead just deploy this schema in AEP-CS directly from here.

>  If not, then give us a few days to make sure this does not affect AEP in a negative way.

+1. Thank you very much.

vpetcu:
@cmathis this is just a kind reminder about this PR. Are there any updates?
cc @fmeschbe @lrosenthol

cmathis (approved):

cmathis:
@vpetcu  - PR has been merged

vpetcu:
@cmathis thank you very much!

Issue #964, opened 2020-09-09
Add item for AEP Web SDK Launch Extension in implementationDetails


jonsnyder:
We'd like to distinguish between hits collected with the AEP Web SDK and those collected with the AEP Web SDK Launch Extension.

## What are the schemas that are affected by the issue
implementationDetails

## What are examples of products that are impacted by the issue
AEP Web SDK Launch Extension

PR #965, opened 2020-09-09
CORE-48374 Use separate name for AEP Web SDK Launch Extension

jonsnyder:
See https://github.com/adobe/xdm/issues/964

PR for code changes in AEP Web SDK https://github.com/adobe/alloy/pull/592
PR for code changes in AEP Web SDK Launch Extension https://github.com/adobe/reactor-extension-alloy/pull/111


components/datatypes/implementationdetails.schema.json
```diff
@@ -26,7 +26,8 @@
             "https://ns.adobe.com/experience/analytics/mbox": "Adobe Target mBox",
             "https://ns.adobe.com/experience/visitorapi": "Adobe Visitor service",
             "https://ns.adobe.com/experience/adcloud/viewability": "Adobe Advertising Cloud viewability",
-            "https://ns.adobe.com/experience/alloy": "Adobe Experience Cloud web SDK (Alloy)"
+            "https://ns.adobe.com/experience/alloy": "Adobe Experience Cloud web SDK (Alloy)",
```
lrosenthol, line 29: We don't use code names in namespace names.  Please adjust

lrosenthol (changes requested):

components/datatypes/implementationdetails.schema.json
jonsnyder: @lrosenthol Currently, production users of the AEP Web SDK are using the value "https://ns.adobe.com/experience/alloy". The reason we went with alloy instead of aepwebsdk is we don't want to negatively affect customers when branding decides to change the name.

If we did change the value, would experience events with the old value be rejected anywhere within platform? I know that one of the goals of XDM is to always make things backward compatible.

jfkhoury:
@dancingcactus @Aaronius

components/datatypes/implementationdetails.schema.json
jfkhoury: Also, Alloy is open source and the name of the Git repo is `alloy`: https://github.com/adobe/alloy

cmathis: Given you already have an "alloy" enum entry, then this is probably okay. but keep in mind an enum value is set in stone. You can never change these enum values which is why it is suggested to not use a potentially changing code name. 
You may also want to check and make sure the plus symbol (+) does not create any problems in your queries or usage of the data.

cmathis: Another think to note is that this is a meta:enum field which is simply a list of suggested values and NOT the only values allowed. You are free to pass any value into this field, but the meta:enum is used by the UI to show potential values in a drop down list.

Aaronius: @jonsnyder, a few tweaks I think we should make. Leave "Alloy" out of the friendly name, capitalize "web", and use "Platform" instead of "Cloud" to be consistent with branding:

"Adobe Experience Platform Web SDK"
"Adobe Experience Platform Web SDK Launch Extension"

@cmathis's admonition about the `+` sign is a good one, but I'm not sure what things you would do to fully validate it or ensure that future tools will properly support it. A hyphen or another slash might be more compliant or appropriate.

Aaronius: Just to give more context, `alloy` is not a code name that goes away or went away when it was officially released. Like @jonsnyder and @jfkhoury eluded to, it's the name that we use in code, APIs, filenames, etc. as a means to have a stable name so that branding can change the more predominately customer-facing names as they'd like without exposing our customers to an unreasonable amount of technical changes. We learned our lesson with mcid/ecid and the like and have taken a similar approach with products like Launch (whose code name is Reactor).

cmathis:
@jonsnyder  - Is this PR ready to be merged? Anyone else that should approve this PR before we do so?

jonsnyder:
@cmathis Yes, we are ready for this to be released

cmathis:
@jonsnyder - Anyone other than yourself that will be using this new meta:enum value and should also approve or be aware of this PR? 

Also, note that a field declared with a meta:enum only (without an actual "enum" attribute) is not constrained to those values only. A user is free to pass any value they want. This meta:enum is only for UI presentation purposes where the field values can be selected via a drop down list.

jfkhoury:
@cmathis We like that feature that customers can pass any values in the future. For example, if a customer is using a 3rd party tag manager, they would send https://ns.adobe.com/experience/alloy/mytagmanager.

I think we are ready to merge this change.

cmathis (approved):

prabhum2 (approved):

PR #970, opened 2020-09-14
Fixed namespace issues with the Customer Journey Management mixins com…

sawhneyh:
…mitted with the usecase 914.

Please link to the issue #…


aashishjassal (approved):

extensions/adobe/experience/customerJourneyManagement/message-delivery-feedback.schema.json
```diff
@@ -34,12 +34,12 @@
           "title": "Message Exclusion",
           "type": "object",
           "properties": {
-            "https://ns.adobe.com/experience/customerJourneyManagement/messageDeliveryfeedback/code": {
+            "https://ns.adobe.com/experience/customerJourneyManagement/messageDeliveryfeedback/messageExclusion/code": {
```
sawhneyh, line 37: @cmathis : We require "code" and "reason" fields  to be sub items of "messageExclusion". Currently when this mixin is merged in AEP, aftr xed conversions "code" and "reason" fields are coming up as "_experience.customerJourneyManagement.messageDeliveryfeedback.messageExclusion._experience.customerJourneyManagement.messageDeliveryfeedback.code".

Check snapshot:  
image

cmathis (approved):

extensions/adobe/experience/customerJourneyManagement/message-delivery-feedback.schema.json
sawhneyh: @cmathis : We require "code" and "reason" field to be a direct sub properties of "messageExclusion" object. Currently after the XED conversion, "code" and "reason" fields are coming as `_experience.customerJourneyManagement.messageDeliveryfeedback.code` under `_experience.customerJourneyManagement.messageDeliveryfeedback.messageExclusion`.

To accomplish this, how we have to define the namespaces for the "code" and "reason" field. Should we have to define these fields without any namespace or using "xdm" namespace. 
Please let me know, I will make the corresponding changes.

cc: @aashishjassal

cmathis: Make sure the parent and child fields have the same namespace (same uri path) and it will come out the way you want it.

sawhneyh: I wanted `code` field to come up as direct sub child of `messageExclusion`. So as I have defined 

`messageExclusion` as `https://ns.adobe.com/experience/customerJourneyManagement/messageDeliveryfeedback/messageExclusion`, 

should I have to define `code` as `https://ns.adobe.com/experience/customerJourneyManagement/messageDeliveryfeedback/code` ?

@cmathis : Please confirm, is above expectation is correct. Though originally, I have defined `code` field exactly like this, but it didn't work.

Is there a way, I can test the XED conversion before raising the PR.

cc: @aashishjassal

cmathis: Yes, what you have above is correct. If this did not work then we may have a bug in the XDM -> XED conversion in the Registry. Let me see if we can get that fixed and then revert back to the way you had before. If I cannot quickly fix this bug then we will figure out another solution.

cmathis:
@sawhneyh, @aashishjassal  - The XDM -> XED converter was fixed. We've reverted your field namespaces in https://github.com/adobe/xdm/pull/973

sawhneyh:
@cmathis : Thanks for raising the PR: https://github.com/adobe/xdm/pull/973. Now field namespaces are as per our expectation.
cc: @aashishjassal

PR #973, opened 2020-09-15
Update cjm field namespaces

cmathis:
Update field namespaces so they appear correctly in AEP


cmathis:
@sawhnehy - This PR reverts the namespace of the fields. We fixed the XDM -> XED converted so it is nested correctly.

PR #978, opened 2020-09-16
Add validation for required field and object type for properties

jwen-adobe:


components/classes/experienceevent.schema.json
```diff
@@ -27,7 +27,6 @@
       }
     }
   },
-  "required": ["@id", "xdm:timestamp"],
```
cmathis, line 30: Not sure we can do this. We may be able to move this required attribute to the time-series behavior which is where these fields come from.

cmathis (approved):

PR #979, opened 2020-09-17
B2B Schema changes

mjasrotia:
Please link to the issue #…


components/mixins/profile/b2b-person-details.schema.json
```diff
@@ -26,6 +26,11 @@
           "description": "Field recording the current marketing/sales status of the Person.",
           "type": "string"
         },
+        "xdm:personSource": {
```
cmathis, line 29: Should this be an enum or set values? What is an example of the value that would be populated here?

PR #981, opened 2020-09-21
New Device Attributes

prabhum2:
Please link to the issue #…

### DIFF
--- a/components/datatypes/application.schema.json
+++ b/components/datatypes/application.schema.json
@@ -29,6 +29,11 @@
           "type": "string",
           "description": "Version of the application."
         },
+        "xdm:sdkVersion": {
+          "title": "SDK Version",
+          "type": "string",
+          "description": "Version of the application sdk."
+        },
         "xdm:applicationCloses": {
           "$ref": "https://ns.adobe.com/xdm/data/measure",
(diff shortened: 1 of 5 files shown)


components/datatypes/application.schema.json
```diff
@@ -29,6 +29,11 @@
           "type": "string",
           "description": "Version of the application."
         },
+        "xdm:sdkVersion": {
```
cmathis, line 32: Which sdk is this? An application could use many sdk's in its implementation. If this is intended to be the airship SDK version, then it seems like that should be in an airship specific field.

components/mixins/shared/external-source-system-audit-details.schema.json
```diff
@@ -15,7 +15,8 @@
     "https://ns.adobe.com/xdm/context/account",
     "https://ns.adobe.com/xdm/context/opportunity-contactrole",
     "https://ns.adobe.com/xdm/context/activitylog",
-    "https://ns.adobe.com/xdm/context/profile"
```
cmathis, line 18: Remove all entries from array to make it a generic mixin usable by any class. The array is now used to constrain to a specific class(es) only.

cmathis (approved):

PR #994, opened 2020-09-28
Experience event mixin for tracking user interactions with push notifications 

iewilliam:
This PR introduces a new mixin for tracking user interaction with a push notification on a mobile app. 
When interacting with a push notification on a mobile device, the end user can either:
- click directly on the notification, launching the application
- initiate a custom action, say by clicking on a button on the notification. This may or may not launch the application

This mixin tracks both types of interactions. 
event type will be set to either pushTracking.applicationOpened or pushTracking.customAction for the 2 cases respectively


components/mixins/experience-event/experienceevent-pushtracking.schema.json
```diff
@@ -0,0 +1,57 @@
...
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Push notification tracking",
+  "type": "object",
+  "meta:extensible": true,
+  "meta:intendedToExtend": ["https://ns.adobe.com/xdm/context/experienceevent"],
+  "description": "push notification tracking details",
+  "definitions": {
+    "experienceevent-pushtracking": {
+      "properties": {
+        "xdm:customAction": {
```
cmathis, line 18: This is going to introduce a root level field named "customAction". Should these new fields be under a "pushNotification" parent field. Also, should this new "pushNotification" field be be added to the existing "experienceevent-marketing" mixin where we have other campaign tracking fields.

iewilliam: Hi Craig, does that mean adding these keys (under a pushNotificationTracking umbrella or something) here: https://github.com/adobe/xdm/blob/master/components/datatypes/marketing.schema.json

cmathis: Yes, if they are marketing related. If it's application related then it would go under https://github.com/adobe/xdm/blob/master/components/datatypes/application.schema.json

Looking to you to know which one you think it falls under.

iewilliam: Hi Craig, taking into account our conversation earlier, do you think it would make more sense to keep this mixin separate from any campaign tracking related mixins? We will use this to track the user interaction only, and keep campaign related info on a separate mixin. We intend to do exactly this for CJM since we would like to keep the mobile SDK and our push service orthogonal to it: on the device we populate this mixin with the user's interaction, and populate the CJM mixins with whatever pre-rendered data that they have provided to us.

cmathis:
@iewilliam  - Would it make sense to put this trackedPushNotifcation under the root level "device" field that exists on ExperienceEvent? Or do you think it belongs as it's own root level concept outside of the context of a device?

components/mixins/experience-event/experienceevent-pushtracking.schema.json
```diff
@@ -0,0 +1,64 @@
...
+                  "description": "user clicked custom action identifier"
+                },
+                "xdm:value": {
+                  "title": "value",
+                  "type": "number",
+                  "description": "value associated with the custom action, if any"
+                }
+              }
+            },
+            "xdm:applicationOpened": {
```
cmathis, line 40: Use the application.launches field instead

components/mixins/experience-event/experienceevent-pushtracking.schema.json
```diff
@@ -0,0 +1,64 @@
...
+              "title": "application opened",
+              "type": "boolean",
+              "description": "Flag indicating whether push notification interaction resulted in opening the application"
+            },
+            "xdm:messageId": {
```
cmathis, line 45: Also add the push platform provider so the messageId can be used later.

components/mixins/experience-event/experienceevent-pushtracking.schema.json
```diff
@@ -0,0 +1,64 @@
...
+                  "description": "value associated with the custom action, if any"
+                }
+              }
+            },
+            "xdm:messageId": {
+              "title": "message ID",
+              "type": "string",
+              "description": "ID associated with this message"
+            },
+            "xdm:platform": {
```
cmathis, line 45: The word "platform" seems very generic here. Would it make sense to name it "pushProvider" or "pushPlatform", or ...?

iewilliam: I can change it to pushProvider.
Maybe we should also change messageID to pushProviderMessageID since it can also be confused with other messageIDs that CJM has in their mixins?

cmathis (approved):
LGTM
@prabhum2 - Please review and approve as well.

prabhum2 (approved):

components/behaviors/time-series.schema.json
```diff
@@ -55,7 +55,9 @@
             "commerce.saveForLaters": "Commerce Save For Laters",
             "delivery.feedback": "Delivery Feedback",
             "message.feedback": "Message Feedback",
-            "message.tracking": "Message Tracking"
+            "message.tracking": "Message Tracking",
+            "pushTracking.applicationOpened": "Push notification launched application",
```
cmathis, line 59: @iewilliam  - Will you update the values so they match the formatting of other meta:enum's

iewilliam: Updated it just now, will this work?
            "pushTracking.applicationOpened": "Push triggered application launches",
            "pushTracking.customAction": "Push custom action clicks"

PR #996, opened 2020-09-29
Currency data type for b2b schemas

prabhum2:
Please link to the issue #…


components/datatypes/b2b-account.schema.json
```diff
@@ -95,13 +95,6 @@
           "description": "Source type for the account.",
           "type": "string"
         },
-        "xdm:accountCurrencyCode": {
```
mjasrotia, line 98: I think we should not remove this property as it could be the default currency of the account and not really related to any amount properties. @peichiang , What is your thoughts about this ?

mjasrotia (approved):

Issue #989, opened 2020-09-24
XDM Event Types are using descriptions instead of friendly names as values


HyveMynd:
## What are the schemas that are affected by the issue
components/behaviors/time-series.schema

## What are examples of products that are impacted by the issue
AEP

PR #1000, opened 2020-09-29
Change Event Type values to Friendly names

HyveMynd:
Please link to the issue #…
#989


cmathis:
@HyveMynd  - Is this field used only by the Ansible UI? Anyone else that needs to approve or be aware of this change?

cc @lshivalin

cmathis (approved):

HyveMynd:
@cmathis primary use is in ansible UI. Unsure of other systems which use it in a way which would have a major effect. However, the eventType enum is not correct according to the enum pattern. I see this as fixing a bug.

prabhum2 (approved):

PR #1007, opened 2020-10-08
Airship Extensions

gatlinhebert:
This is an initial pass at defining Airship-specific fields in an Airship mixin and a wrapper mixin that pulls in other mixins useful for working with Airship engagement data in AEP.


mjasrotia:
@gatlinhebert , Are you guys still working on this PR or is it good for an initial review  ?

gatlinhebert:
@mjasrotia Yes, this is ready for an initial review.

cmathis:
The namespace of these fields is correct, but it's not going to be pretty when its shown in AEP. The way AEP works, it converts the URI field name into a nested structure. As is, your field paths are going to look something like _ns.airship.com.body. I you would rather then field path be _airship.body, then we need to change the domain and field names to be something like "https://ns.adobe.com/airship/body". It still provides a unique namespace for the field, but the ns.adobe.com has some special treatment in AEP to avoid the long ugly path.

extensions/airship/airship-event-extension.schema.json
```diff
@@ -0,0 +1,271 @@
+{
+  "$id": "https://ns.airship.com/airship-event-extension",
+  "title": "Airship Event Extension",
+  "description": "This mixin adds Airship specific event fields.",
+  "meta:intendedToExtend": [
```
cmathis, line 5: Need to add a $schema attribute here. Look at other schemas in this repo and copy it.

extensions/airship/airship-event-extension.schema.json
```diff
@@ -0,0 +1,271 @@
...
+    "https://ns.adobe.com/xdm/context/experienceevent"
+  ],
+  "properties": {
+    "https://ns.airship.com/offset-id": {
```
cmathis, line 9: See my general comment. You probably want this to be named https://ns.adobe.com/airship/offset-id

extensions/airship/airship-event-extension.schema.json
```diff
@@ -0,0 +1,271 @@
...
+            },
+            "https://ns.airship.com/variant-id": {
+              "title": "Airship Variant ID",
+              "description": "The ID of the variant that a push is associated with, if the push was a part of an A/B test (experiment).",
+              "type": "string"
+            },
+            "https://ns.airship.com/push-time": {
+              "title": "Push Time",
+              "description": "The UTC time when the push occurred.",
+              "type": "date-time"
```
cmathis, line 112: This is not valid json-schema. The "type" you have here is actual the "format". The type should be "string". Please change to:
"type": "string",
"format": "date-time"

jwen-adobe:
Added "type": "object" under root

lrosenthol:
Is Airship an internal Adobe product or a 3rd party company/partner?  If the former, then using `ns.adobe.com` is fine - if the latter, then it is not.

extensions/airship/airship-event-extension.schema.json
prabhum2: also i suggest to use offsetID instead of offset-id to be in consistent with other schema property names . Same applies for other property names

cmathis:
@lrosenthol is correct - this is a 3rd party and it should be ns.airship.com, but I know it's going to be very long nested path that user needs to interact with in AEP - see the one example field in the doc here - https://github.com/adobe/xdm/blob/master/NAMESPACES.md. Let me talk with some of the internal group to find out the best approach here. Maybe this will help give AEP the push it needs to stop exposing the raw XED format

gatlinhebert:
I added the recommended changes as well as region properties (per recommendation from email thread). Thanks, everyone!

extensions/airship/airship-event-extension.schema.json
```diff
@@ -0,0 +1,295 @@
...
+          "title": "Alerting Push",
+          "description": "If true, the send event was alerting. Alerting send event has notification text, badge, or sound.",
+          "type": "boolean"
+        },
+        "https://ns.airship.com/triggeringPush": {
+          "title": "Triggering Push",
+          "description": "The specific push notification details associated with an event. An associated push helps you trace an event to the original notification or operation.An associated push object may specify a `time`, if the push was a singular operation sent at a defined time. Otherwise, the object will include a `group_id` if the push was sent at a relative time (`best_time` or `local_time`) an automation pipeline, or another operation resulting in multiple `push_id`s.",
+          "type": "object",
+          "properties": {
+            "https://ns.airship.com/campaign": {
```
prabhum2, line 70: As you are using campaign object multiple times/places I suggest you to create Campaign data type and reference that data type where and when needed

cmathis: Maybe as a local definition that is referenced, but lets please not introduce yet another datatype to represent an Airship Campaign that is never to be used outside of this mixin. Keep everything contained in this one file.

gatlinhebert: That's correct, it's the same Campaign object in multiple places but I wasn't sure if there's a better way to handle it.

prabhum2 (approved):

extensions/airship/airship-event-extension.schema.json
```diff
@@ -0,0 +1,295 @@
...
+                "unsubscribe"
+              ]
+            },
+            "airship:commercialOptedIn": {
+              "title": "Commercial Opted In Date",
+              "description": "The date and time when the user opted into commercial email messages.",
+              "type": "string",
+              "format": "date-time"
+            },
+            "https://ns.airship.com/bounceClass": {
```
cmathis, line 250: @prabhum2  - Fix this

extensions/airship/airship-event-extension.schema.json
```diff
@@ -0,0 +1,295 @@
...
+            "opted_out",
+            "uninstall"
+          ]
+        },
+        "airship:complianceProperties": {
+          "title": "Compliance Event Properties",
+          "description": "Additional properties for a compliance event.",
+          "type": "object",
+          "properties": {
+            "https://ns.airship.com/messageType": {
```
cmathis, line 225: Fix this

cmathis (approved):

prabhum2 (approved):

Issue #1008, opened 2020-10-12
Create Common Retail-specific Mixins


jbranson000:
Enriches and optimizes the standard XDM components for the retail vertical.

PR #1009, opened 2020-10-12
Add files via upload

jbranson000:
https://github.com/adobe/xdm/issues/1008

### DIFF
--- a/components/mixins/experience-event/advertisingDetails.schema.json
+++ b/components/mixins/experience-event/advertisingDetails.schema.json
@@ -1,0 +1,31 @@
+{
+  "meta:license": [
+    "Copyright 2020 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/mixins/experienceevent-social",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Advertising Details",
+  "type": "object",
+  "meta:extensible": true,
+  "meta:abstract": true,
+  "meta:intendedToExtend": ["https://ns.adobe.com/xdm/context/experienceevent"],
+  "description": "Advertising information such as ad impressions and ad clicks. ",
+  "definitions": {
+    "advertising-details": {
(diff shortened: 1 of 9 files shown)


mjasrotia:
@jbranson000 , few general comments/questions on these Schemas.
1. The Schema Ids need to be unique, Please fix the schema Ids in your file, Are are defined as 
"$id": "https://ns.adobe.com/xdm/mixins/experienceevent-social"

2. Schema names doesn't need to be camel case. it could be defined for example as 
"$id": "https://ns.adobe.com/xdm/mixins/experienceevent-advertising-details"

3. I see all the schemas are created as mixins. Is there a need of these many mixins or they could be combined under one mixinn with each property referring a data type ? My concern here is that if the customers would use multiple  (or may be all) mixins to compose their scehemas, then too many mixins might be overwhelming to start with.

4. I see, you are defining just 1 property per mixin and that too is referring to a data type. Is there a use case that these data types would be used outside these mixins ? If no, then could you just put the properties from the data types at the root of each mixin ? 

5. I am assuming, you are yet to create the schemas for each data type right ?

jbranson000:
> @jbranson000 , few general comments/questions on these Schemas.
> 
> 1. The Schema Ids need to be unique, Please fix the schema Ids in your file, All are defined as
>    "$id": "https://ns.adobe.com/xdm/mixins/experienceevent-social"
I updated the script that is generating the JSON to properly define the IDs - this should be resolved now.
> 2. Schema names doesn't need to be camel case. it could be defined for example as
>    "$id": "https://ns.adobe.com/xdm/mixins/experienceevent-advertising-details"
I also update the script to make sure camelCase was not being used for the name.
> 3. I see all the schemas are created as mixins. Is there a need of these many mixins or they could be combined under one mixin with each property referring a data type ? My concern here is that if the customers would use multiple  (or may be all) mixins to compose their schemas, then too many mixins might be overwhelming to start with.
These mixins are being setup to be used in various verticals - and when we were defining them we tried to keep them as granular as possible to give the clients more flexibility as they incorporated them into their own schemas.  They could probably be combined per vertical when after we include the vertical metadata tags.
> 4. I see, you are defining just 1 property per mixin and that too is referring to a data type. Is there a use case that these data types would be used outside these mixins ? If no, then could you just put the properties from the data types at the root of each mixin ?
Yes, several data types that we are creating will be used outside these mixins.
> 5. I am assuming, you are yet to create the schemas for each data type right ?
I have not created any data types yet, although we do have them defined in our workbook.

PR #1010, opened 2020-10-12
PLAT-72130 Add segment namespace to destination mappings

kkoteeswaran:
Please link to the issue #…
https://jira.corp.adobe.com/browse/PLAT-72130 
For destination-segment mappings in destinations dashboards, we need to record the segment namespace as well since segment IDs may not be unique across different namespaces. For example, the same ID might exist for different segments in AAM and AEP. 

So segment namespace needs to be added to the destination-segment schema here - https://github.com/adobe/xdm/blob/master/schemas/destinations/destination-segment.schema.json


mjasrotia (approved):

schemas/destinations/destination-segment.schema.json
```diff
@@ -26,6 +26,11 @@
           "type": "string",
           "description": "Identity of the segment in the related namespace"
         },
+        "xdm:segmentNamespace": {
```
prabhum2, line 29: I thought this is done at run time as per our earlier design . Something changed ?

SELECT *

from

(

Select dateID,

destinationName,

segmentName,

namespaceCode,

totalAddressableIdentityCount taic,

Row_Number() OVER(Partition BY destinationID,

segmentID Order by SUM(totalAddressableIdentityCount) DESC) RN

from

fact_destination_segment_namespace_daily fdsnd

INNER JOIN dim_destinations dd ON (fdsnd.destinationID =

dd.destinationID)

INNER JOIN dim_segments ds ON (fdsnd.segmentID =

dd.segmentID)

Where

dateId = '20200601'

) WHERE RN <= 5

kkoteeswaran: The segment namespace here is different from destination namespace which is captured in the deestination-namespace schema. A segment ID may not be unique across namespaces. For example, the same ID might exist in AAM segments and AEP segments so to distinguish them we need a segment namespace (`aepsegments` for example)

prabhum2 (approved):

cmathis (approved):

Issue #1017, opened 2020-10-19
Create Financial Services Mixins & DataTypes


jbranson000:
Enriches and optimizes the standard XDM components for the financial services vertical.

PR #1018, opened 2020-10-19
1017 card comparison mixin and comparison data type

jbranson000:
https://github.com/adobe/xdm/issues/1017


cmathis:
@jbranson000  - Is this new datatype ever to be used outside of the context of this new "cardComparisonDetails" field? If not, then let's define those field directly under the field its self or as a local definition you reference. Let's not clutter AEP with new datatypes that have no reusability.

cmathis:
What is a "Card" in this case and how are they related to ExperienceEvent datasets?

jbranson000:
@cmathis A "card" could be any type of credit, debit, HSA, FSA, Prepaid, etc.  Most of the mixins we are creating are part of the ExperienceEvent datasets.

cmathis:
@jbranson000 - A potential issue here is that you are wanting to introduce this "xdm:cardComparisonDetails" as a new root level field. In ExperienceEvent data root level fields are generally a high level category which contains fields under it related to that category. I'm assuming this "cardComparisonDetails" would be associated with other banking related fields. What could be the high level category associated with these fields? Once you have that you could do something like:

`      
"properties": {
     "xdm:consumerBanking": { //Or whatever you think this should be
       "type": "object",
       "properties": {
            "xdm:cardComparisonDetails ": {
                "title": "Card Comparison Details",
                "$ref": "https://ns.adobe.com/xdm/datatypes/comparisons",
                "description": "Card comparison details such as count, addition, removal, and view."
             }
       }
}
`

jbranson000:
This datatype and mixin is being merged into 1018 - thus this can be closed.

PR #1019, opened 2020-10-21
Add Identity Namespace schema for PQS use case

shamsadobe:
Introduced Identity Namespace as a global schema. Identity namespace schema allows solutions to defined additional namespace properties other than namespaceCode.

More details can be found here.
https://jira.corp.adobe.com/browse/PLAT-72516


schemas/namespace/identity-namespace.example1.json
```diff
@@ -0,0 +1,7 @@
+{
+  "xdm:namespaceCode": "ECID",
+  "id": "4657",
```
lrosenthol, line 3: all of the values need to be properly namespaced

schemas/namespace/identity-namespace.example2.json
```diff
@@ -0,0 +1,7 @@
+{
+  "xdm:namespaceCode": "Email",
+  "id": "4658",
```
lrosenthol, line 3: all of the values need to be properly namespaced

schemas/namespace/identity-namespace.schema.json
```diff
@@ -0,0 +1,77 @@
...
+  "description": "Identity Namespace Schema.",
+  "definitions": {
+    "identity-namespace": {
+      "properties": {
+        "xdm:id": {
+          "title": "Unique Namespace Id",
+          "type": "string",
+          "description": "Unique Identifier for namespace"
+        },
+        "namespaceName": {
```
lrosenthol, line 21: you're missing the `xdm` here

schemas/namespace/identity-namespace.schema.json
```diff
@@ -0,0 +1,77 @@
...
+          "title": "Namespace Name",
+          "type": "string",
+          "description": "This is human readable name for the namespace"
+        },
+        "xdm:namespaceCode": {
+          "title": "Namespace Code",
+          "type": "string",
+          "description": "This code is a human readable identifier for the namespace"
+        },
+        "description": {
```
lrosenthol, line 31: you're missing the `xdm` here

schemas/namespace/identity-namespace.schema.json
```diff
@@ -0,0 +1,77 @@
...
+          "title": "Namespace Description",
+          "type": "string",
+          "description": "Description for the namespace"
+        },
+        "idType": {
```
lrosenthol, line 36: you're missing the `xdm` here

Also, don't understand what this means...

schemas/namespace/identity-namespace.schema.json
```diff
@@ -0,0 +1,77 @@
...
+          "enum": [
+            "EMAIL",
+            "PHONE_NUMBER",
+            "DEVICE",
+            "COOKIE",
+            "CROSS_DEVICE",
+            "NON_PEOPLE"
+          ]
+        },
+        "xdm:createTime": {
```
lrosenthol, line 49: this should be `dc:createDate`

schemas/namespace/identity-namespace.schema.json
```diff
@@ -0,0 +1,77 @@
...
+          "title": "Create Time",
+          "type": "string",
```
lrosenthol, line 51: also add 'format: date-time`

schemas/namespace/identity-namespace.schema.json
```diff
@@ -0,0 +1,77 @@
...
+          "description": "The date and UTC Time when the namespace was created"
```
lrosenthol, line 52: also state that it is in ISO 8601 format

schemas/namespace/identity-namespace.schema.json
```diff
@@ -0,0 +1,77 @@
...
+        },
+        "updateTime": {
```
lrosenthol, line 54: this should be `dc:modifyDate`

schemas/namespace/identity-namespace.schema.json
```diff
@@ -0,0 +1,77 @@
...
+          "title": "Update Time",
+          "type": "string",
```
lrosenthol, line 56: add `format:date-time`

schemas/namespace/identity-namespace.schema.json
```diff
@@ -0,0 +1,77 @@
...
+          "description": "The date and UTC Time when the namespace was updated"
```
lrosenthol, line 57: state that it is in ISO 8601 format

lrosenthol (changes requested):

schemas/namespace/identity-namespace.example1.json
shamsadobe: ack

schemas/namespace/identity-namespace.schema.json
shamsadobe: It represents a type for Namespace. Example "foo@bar.com" could of type "Email". You could call it a category of namespaces.

schemas/namespace/identity-namespace.schema.json
shamsadobe: The fields "xdm:createTime" and "dc:modifyDate" are having complete different names. Is there a "xdm:modifiedTime" field defined which I could use?

schemas/namespace/identity-namespace.schema.json
lrosenthol: but that's not a namespace - at least in the XDM terminology.  And that is causing the confusion.

Is there an alternate term, for example, identity type?

schemas/namespace/identity-namespace.example1.json
lrosenthol: Welcome to XDM

schemas/namespace/identity-namespace.schema.json
shamsadobe: ya we could say that. @harikb -- any recommendations here

schemas/namespace/identity-namespace.schema.json
lrosenthol: if you wanted it to be in the `xdm` namespace, then use `xdm:createDate` and `xdm:modifyDate`

schemas/namespace/identity-namespace.schema.json
shamsadobe: ack

schemas/namespace/identity-namespace.schema.json
shamsadobe: ack

schemas/namespace/identity-namespace.schema.json
shamsadobe: ack

schemas/namespace/identity-namespace.schema.json
shamsadobe: ack

schemas/namespace/identity-namespace.schema.json
shamsadobe: if we use "xdm:code" or "xdm:name", will it have conflicts because i see there is a datatype defined for "xdm:name" already ("/xdm/common/extensible#/definitions/@context") and probably for other fields as well. Would it makes sense to have our own namespace and define fields using it ?

lrosenthol: as long as the type is the same (in this case, they are both strings) - you are fine.

lrosenthol (approved):
Thanks for the changes!   LGTM!

cmathis:
@shamsadobe  - I'm assuming this represents the very specific Adobe Unified Identity System namespace correct? I also added some comments to your Jira ticket.

shamsadobe:
@cmathis Yes thats correct.

cmathis:
@shamsadobe - Since this schema is very much related to Unified Identity Service, how do you feel about moving it to the /extensions/adobe/experience location within this repo and having each field associated with the https://ns.adobe.com/experience/identity namespace instead?

shamsadobe:
@cmathis This makes more sense. The namespace and the locations were recommended by @khanna-vikas . @khanna-vikas  any thoughts on this. Are we ok with this ?

khanna-vikas:
@shamsadobe & @cmathis  These are global definitions to be containerised alongside /schemas as we bring more of them to satisfy the exports on the data lake by different AEP services. Destinations, SegmentDefinitions, namespace and their relationship mappings should all hold under the core namespace, as they will be internal constructs. We should embrace all of the global definitions in the same way of representation.

cmathis:
@khanna-vikas  - You can still define a global schema in the extensions area. My only reason for putting there is because this is tightly coupled to the UIS definition of a namespace. If you think the definition here is generic and could be used by other non UIS based systems then we can leave it here under the XDM namespace.

khanna-vikas:
@cmathis I believe that global definition of schemas has two pathways. Solutions define their Global schemas as they are never going to get extended by the tenants and absorb the data based on their definitions. eg. Adcloud. Here they are control body of their schemas and should exist under their namespace.

2nd set of global definitions exists where these logical schemas will have datasets which will be produced and consumed by the internal services on the platform. The reason for the global scope was to introduce consistency. The definition is generic and evolve over time but will be used by many services to build their experiences. I feel it should be under XDM namespace.

cmathis:
@khanna-vikas  - Great definition of how to decide. It stays here then.

cmathis:
@shamsadobe  - My only request then is to simply move this into a folder named "identity" rather then "identity-namespace". No good reason other then we may have other identity related schemas in the future.

shamsadobe:
@cmathis I have made the folder location changes as requested. Thanks!

cmathis (approved):

prabhum2 (approved):

schemas/identity/identity-namespace.schema.json
```diff
@@ -0,0 +1,95 @@
...
+          "format": "date-time",
+          "type": "string",
+          "examples": ["An example form is \"2004-10-23T12:00:00-06:00\"."],
+          "description": "The date and time when the namespace was updated in ISO 8601 format"
+        }
+      },
+      "required": [
+        "xdm:namespaceID",
+        "xdm:namespaceCode",
+        "xdm:namespaceName",
```
ramandumcs, line 78: Do we need "xdm:namespaceName" &  "xdm:namespaceIdType" as required fields?. Can we keep this optional

PR #1031, opened 2020-11-03
PLAT-73212 Update createTime datatype

kkoteeswaran:
Please link to the issue #…
https://jira.corp.adobe.com/browse/PLAT-73212

Datatype of createTime fields need to be changed from string to date-time in all destinations XDM schemas. This is required since data types will be strictly enforced in datawarehouse and to avoid additional transformations there.


mjasrotia:
@kkoteeswaran Adding a format might make the previously ingested data bad. Is that OK ?

kkoteeswaran:
That's alright @mjasrotia These schemas are not used in production yet. They have only been used to write datasets to ADL during testing in INT/Stage.

mjasrotia (approved):

prabhum2 (approved):

cmathis (approved):

PR #1032, opened 2020-11-04
Replacement Consents & Preferences Schema

RussStringham:
[PLAT-71029](https://jira.corp.adobe.com/browse/PLAT-71029)


components/datatypes/consent-preferences.schema.json
```diff
@@ -0,0 +1,261 @@
...
+    "choices": {
+      "properties": {
+        "xdm:collect": {
+          "title": "Data Collection",
+          "description": "Data collection is permitted",
+          "$ref": "#/definitions/consent-fields"
+        },
+        "xdm:share": {
+          "title": "Share Data",
+          "description": "Sharing of user's Data with 2nd or 3rd parties is permitted",
```
Aaronius, line 88: Should `Data` be lower-cased here?

components/datatypes/consent-preferences.schema.json
```diff
@@ -0,0 +1,261 @@
...
+          "type": "object",
+          "properties": {
+            "xdm:preferred": {
+              "title": "Preferred Channel",
+              "description": "User's preferred channel for receiving communications",
+              "type": "string",
+              "enum": [
+                "email",
+                "push",
+                "in_app",
```
Aaronius, line 125: As I look through XDM is seems like we have some enums camelCase, some TitleCase, and some snake_case. Is there an XDM standard outlined somewhere?

components/datatypes/consent-preferences.schema.json
```diff
@@ -0,0 +1,261 @@
...
+              "description": "User agrees to receive text messages",
+              "$ref": "#/definitions/marketing-fields"
+            }
+          }
+        }
+      }
+    },
+    "metadata": {
+      "properties": {
+        "xdm:t": {
```
Aaronius, line 179: Are we using `t` instead of `timestamp` for optimization purposes?

Aaronius (commented):
Thanks so much @RussStringham for working on this. I don't suppose I'm part of the formal review process, but I added some comments anyway.

components/datatypes/consent-preferences.example.1.json
```diff
@@ -0,0 +1,37 @@
+{
+  "xdm:consents": {
+    "xdm:collect": {
+      "xdm:v": "y"
```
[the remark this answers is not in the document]
RussStringham: Yes. "value" was too long when it is repeated 10 times (once for each consent in current schema or 35 times for each consent expected in the future).

components/datatypes/consent-preferences.schema.json
```diff
@@ -0,0 +1,261 @@
...
+      "meta:enum": {
+        "y": "Yes",
+        "n": "No",
+        "p": "Pending verification",
+        "u": "Unknown",
+        "LI": "Legitimate Interest",
+        "CT": "Contract",
+        "CP": "Compliance with a Legal Obligation",
+        "VI": "Vital Interest of the Individual",
+        "PI": "Public Interest"
```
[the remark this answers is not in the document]
RussStringham: I originally had these split into two fields with the full strings. Both the combining and the abbreviations are to reduce the schema size as my original schema was deemed to be much too large.
The lower case, single letter options represent "consent" values, while the upper case, two-letter values represent basisOfProcessing values where user consent is not required. My understanding is that in the UI customers will see the full description for each value, so the shortened versions won't be a problem.
Customers told us that the terms in/out were confusing in some contexts, especially in the old consent schema: What does 
  "general_opt_out": "in"
mean? Are you opting in to opting out or are you not opting out?

components/datatypes/consent-preferences.schema.json
RussStringham: I asked about this previously and was told all are in common use with no standard. I suppose I could switch these to camelCase to eliminate the underscores, reducing the JSON by a byte each time one occurs.

components/datatypes/consent-preferences.schema.json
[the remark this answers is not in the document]
Aaronius: I think when you say "UI Customers" you're referring to the end user on our customer websites looking at, for example, the CMP UI. If so, I don't think the abbreviations are a problem for them either. I was thinking more about the Adobe customers that are trying to debug their consent integration and are inspecting network requests as they travel through Alloy (or the Mobile SDK), through the edge, and into their dataset, etc. I don't really know what/who is determining the threshold of "too large" though.

Regarding `in`/`out` being confusing in some contexts, that makes sense with `general_opt_out`. Is this new schema going to be used in confusing contexts like `general_opt_out`? I mainly ask because we discussed this problem at length with Karthik and others regarding those confusing contexts and came to a consensus that we would avoid the need for these confusing contexts in the web SDK and edge so that would could properly use `in`/`out` in a way that's understandable to customers. For example, you'll see this in the `defaultConsent` configuration of the SDK: https://experienceleague.adobe.com/docs/experience-platform/edge/fundamentals/configuring-the-sdk.html?lang=en#default-consent

Notice the two possible values for `defaultConsent` are `in` and `pending`. It would be confusing, IMO, to have them be `yes` and `pending` and I'd love for us to be consistent. There is no confusing concept of "configuring opt-out" in the web SDK or edge.

components/datatypes/consent-preferences.schema.json
Aaronius: I don't personally have much of a preference. I was just hoping it was standardized. Thanks for checking into this.

cmathis:
@RussStringham  - We need to update the field names in this PR to something meaningful. Given this is intended to introduce standard fields to our Profile schema we need to make sure the fields are self documenting as much as possible.

RussStringham:
@cmathis I have updated "xdm:v" to "xdm:val" and "xdm:t" to "xdm:time". Is this sufficient or do others need to be updated as well. I shortened them because the UPS team said that the long field names/value were taking up too much space in their system.

components/datatypes/consent-preferences.schema.json
```diff
@@ -179,7 +179,7 @@
     },
     "metadata": {
       "properties": {
-        "xdm:t": {
+        "xdm:time": {
           "title": "Consent/Preferences timestamp",
           "description": "Timestamp of this complete set of user consents and preferences",
```
lrosenthol, line 184: (you should probably note that it is in ISO 8601 or RFC3339)

components/datatypes/consent-preferences.schema.json
```diff
@@ -179,7 +179,7 @@
     },
     "metadata": {
```
lrosenthol, line 180: in what way is this metadata?  Perhaps there is a better term?

cmathis:
@Aaronius - Are you one of the stake holders or representing a team that will be using these consent fields? If so, we would like to see your verbal approval before moving forward.

Aaronius:
If I understand correctly, Alloy will be passing this new consent object through to Konductor, but Alloy won't have any awareness of the structure or contents of the object itself. As such, I probably don't warrant much as far as approval rights go. This may affect @raresvla though, who would most likely represent Konductor.

Regardless, I do have a vested interest in making the lives our customers better, so I'll leave some thoughts and you can do what you'd like with them. With regard to this PR, the most important thing I see is that we've really thought through how to represent "opted in" and "opted out" (discussed in https://github.com/adobe/xdm/pull/1032#discussion_r520180572).  I think the values of `yes` and `no` are confusing and not as self-documenting as `in` and `out`. As Russ mentioned, in previous works we've had field names like `general_opt_out` which aren't conducive to the values of `in` and `out`. This seems to have influenced the proposal of using `yes` and `no` instead of `in` and `out`. However, it seems that an ill-conceived concept like `general_opt_out`, which is hopefully considered legacy at this point, is driving the language of XDM, when I'd rather see XDM use language that is most appropriate for consent use cases and have implementations consume and support that language.

I also wonder if it may be a mistake to be naming XDM fields based on limitations of a particular system implementation (UPS storage limitations, specifically). I was under the assumption that XDM was striving to be implementation-agnostic, but maybe that's not the case or isn't practical. I don't know enough about UPS to understand how they may solve their storage woes, but this may be another case of the tail wagging the dog.

jfkhoury (approved):

ankushkd:
Functionally PR looks fine. I am not sure about shortening of names of XDM fields. This should be handled in persistence level I suppose but will let @khanna-vikas  @prabhum2 confirm.

components/datatypes/consent-preferences.schema.json
```diff
@@ -0,0 +1,256 @@
...
+  "$id": "https://ns.adobe.com/xdm/datatypes/consents-and-preferences",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Privacy/Personalization/Marketing Preferences (Consent)",
+  "description": "This schema captures privacy, personalization and marketing preferences (consents).",
+  "type": "object",
+  "meta:extensible": true,
+  "meta:abstract": true,
+  "definitions": {
+    "choice-value": {
+      "title": "Choice Value",
```
amy-durocher, line 17: Does it make sense to change to "Consent options" or "Consent types"?

components/datatypes/consent-preferences.schema.json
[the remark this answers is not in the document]
amy-durocher: I completely agree with aaronius evaluation here. We should be consistent and provide very clear options. I much prefer using the language "Opt-in" "Opt-out" (qualifies what in and out really mean).

components/datatypes/consent-preferences.schema.json
```diff
@@ -0,0 +1,256 @@
...
+        "LI": "Legitimate Interest",
+        "CT": "Contract",
+        "CP": "Compliance with a Legal Obligation",
+        "VI": "Vital Interest of the Individual",
+        "PI": "Public Interest"
+      }
+    },
+    "timestamp": {
+      "title": "Preference timestamp",
+      "description": "Time this specific preference was collected. This value should conform to ISO 8601 standard. An example form is \"2004-10-23T12:00:00-06:00\".",
```
amy-durocher, line 35: I know a bit obvious: "**Date and** time this specific preference was collected. This value should conform to ISO 8601 standard. An example form is \"2004-10-23T12:00:00-06:00\"."

components/datatypes/consent-preferences.schema.json
```diff
@@ -0,0 +1,256 @@
...
+      "type": "string",
+      "format": "date-time"
+    },
+    "source": {
+      "title": "Preference Source",
+      "description": "Through which interface did the user provide this preference value",
```
amy-durocher, line 41: Should it be channel instead of interface? I would slight reword: "Source or channel the user provided the preference value."

components/datatypes/consent-preferences.schema.json
```diff
@@ -0,0 +1,256 @@
...
+      "type": "object",
+      "properties": {
+        "xdm:val": {
+          "$ref": "#/definitions/choice-value"
+        },
+        "xdm:time": {
+          "$ref": "#/definitions/timestamp"
+        },
+        "xdm:reason": {
+          "title": "Opt out reason",
```
amy-durocher, line 74: Can you please add hyphen "Opt-out"

prabhum2:
@RussStringham - Please let me know once this PR is ready to be pushed in Int and Stage

prabhum2 (approved):

PR #1033, opened 2020-11-04
Addressing changes requested by PQS use case

shamsadobe:
- I have addressed some comments from Anil and @ramandumcs.
- Made "namespaceID" as integer from string
- Removed name & id type from required fields

https://jira.corp.adobe.com/browse/PLAT-72516


shamsadobe:
"Data type of property "definitions/identity-namespace/properties/xdm:namespaceID/type" can not be changed" - is there a way to know which schema compatibility is broken with this change ?. Log's did not provide any name of the mixin/class/schema

mjasrotia (approved):

prabhum2 (approved):

PR #1037, opened 2020-11-06
Add Mobile Lifecycle Details schema and Mobile Application datatype

kevinlind:
Adds Mobile Application as a global datatype. The Mobile Application datatype allows mobile channels to provide details regarding the mobile application and application lifecycle.

Adds Mobile Lifecycle Details mixin for the AEP Mobile SDK. This mixin uses the Mobile Application datatype and the existing Environment and Device datatypes. It provides definitions for automatically collected data by the AEP Mobile SDK.

https://jira.corp.adobe.com/browse/PLAT-73375


mjasrotia:
@kevinlind , We already have a well defined mixin "experienceevent-environment-details" which captures the "device" and "environment" details. Is there a reason why are you redefining them in the new MIXIN you created ? If you are trying to model the life cycle of a mobile here, would you consider adding the "xdm:mobileApplication" property from the new mixin to the existing mixin "experienceevent-environment-details" ? That way you will get all the properties you desire and we won't have to add a new mixin. Just a thought.

components/datatypes/mobile-application.json
```diff
@@ -0,0 +1,133 @@
+{
+  "meta:license": [
+    "Copyright 2020 Adobe. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/context/mobile-application",
```
mjasrotia, line 8: We use the "https://ns.adobe.com/xdm/datatypes/......" for all new data types we define now.

components/datatypes/mobile-application.json
```diff
@@ -0,0 +1,133 @@
...
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Mobile Application",
+  "type": "object",
+  "description": "Details related to mobile application generated interactions. Mobile Application refers to a mobile software experience that can be installed, run, closed, or uninstalled by an end user. Details additionally include session and application lifecycle metrics reported by the application at the time of the experience.",
+  "meta:extensible": true,
+  "definitions": {
+    "mobileApplication": {
+      "properties": {
+        "xdm:id": {
```
mjasrotia, line 17: Please rename this as "ID"

components/datatypes/mobile-application.json
```diff
@@ -0,0 +1,133 @@
...
+          "title": "Name",
+          "type": "string",
+          "description": "Name of the application."
+        },
+        "xdm:version": {
+          "title": "Version",
+          "type": "string",
+          "description": "Version of the application."
+        },
+        "xdm:isLaunch": {
```
mjasrotia, line 32: Would you consider define the events as "measure" datatype as that is the preferred way of defining events today on the platform. That way, analytics will have a consistent manner to process all events on the platform.

components/datatypes/mobile-application.json
```diff
@@ -0,0 +1,133 @@
...
+          "type": "string",
+          "description": "Specifies the launch event type. Install indicates this is the first application launch after the application was installed. Upgrade indicates this is the first application launch after the application was upgraded. Launch is an application launch which is not after an application install or upgrade.",
+          "enum": ["launch", "install", "upgrade"],
+          "meta:enum": {
+            "launch": "Launch",
+            "install": "Install",
+            "upgrade": "Upgrade"
+          }
+        },
+        "xdm:isClose": {
```
mjasrotia, line 47: Would you consider define the events as "measure" datatype as that is the preferred way of defining events today on the platform. That way, analytics will have a consistent manner to process all events on the platform.

components/datatypes/mobile-application.json
```diff
@@ -0,0 +1,133 @@
...
+          "description": "Number of days since last upgrade event."
+        },
+        "xdm:launchesSinceLastUpgrade": {
+          "title": "Launches since last upgrade",
+          "type": "integer",
+          "description": "Number of launch events since last upgrade event."
+        },
+        "xdm:installDate": {
+          "title": "Install date",
+          "type": "string",
```
mjasrotia, line 103: Please add "format": "date-time"

components/datatypes/mobile-application.json
```diff
@@ -0,0 +1,133 @@
...
+          "description": "Date of first launch after installation."
+        },
+        "xdm:launches": {
```
mjasrotia, line 106: Would you consider defining this as "measure" datatype as that is the preferred way of defining events today on the platform. That way, analytics will have a consistent manner to process all events on the platform.

extensions/adobe/experience/mobile-lifecycle-details.schema.json
```diff
@@ -0,0 +1,46 @@
+{
+  "meta:license": [
+    "Copyright 2020 Adobe. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/experience/mobile-lifecycle-details",
```
mjasrotia, line 8: Please use the following ID pattern to define new MIXINS 
"https://ns.adobe.com/mixins/......."

For this case, it should be
"https://ns.adobe.com/mixins/mobile-lifecycle-details"

prabhum2:
@kevinlind Please let me know when this is PR is ready for XDM team review ?

components/datatypes/mobile-application.json
kevinlind: Thanks, I missed this. The format is "date", without time.

lrosenthol: A `date-time` need not include the time - it is a standard ISO 8601 value.

kevinlind:
@mjasrotia Our team is looking to move away from using the legacy Measure type when defining events and would like to use this experimental mixin to try out a new approach of using booleans.

kevinlind:
I've changed the mobile-lifecycle-details mixin to include the Environment datatype but without the "browserDetails" field. I've met with @cmathis about how to pull specific fields into a mixin and he pointed me to profile-person-details-v2.schema in #1043. Please let me know if I've done this correctly and if this will bring the Environment datatype into the mixin without the "broswerDetails" field.

components/datatypes/mobile-application.json
kevinlind: Is `"format": "date"` an invalid format?  I do see it in other schemas (address, person, product, subscription, campaign, etc). I wonder if all date formats are specified as `date-time`, how can consumers of schemas differentiate between dates using RFC3339 `date-time` and dates using RFC3339 `full-date` (date without time)?

lrosenthol: Yes, it is.  Because we are based on JSON Schema, where there is only `date-time`  (https://json-schema.org/understanding-json-schema/reference/string.html#dates-and-times).   NOTE: while there is support for `date` or `time` only in newer version, we are still based on v6.

lrosenthol: @kevinlind Since RFC 3339 is a subset of ISO 8601, that's fine to use.   But any valid RFC3339 parser will handle any/all valid dates, times or date-times as defined by that spec.   so not sure what the concern is?!?!

kevinlind: This document shows that `date` is a valid XDM format, is that no longer the case? https://github.com/adobe/xdm/blob/master/docs/data_types.md#data-type-definitions.
AEP appears to support it as well when creating a custom mixin from the UI. A field of type Date will set the format to `date` in the generated schema. 

The concern is if a customer wants to use a date format without time, there's no way to know that from the schema if XDM only supports a `date-time` format. XDM producers won't know how to handle the format and will need to default to the full date-time format. Manual producers can possibly look at the field description for a clue to the format, but automated producers, such as some of the code in the mobile SDK, cannot.

kevinlind:
@mjasrotia Regarding your question about using the "environment-details" mixin, I didn't want to include the "PlaceContext" schema into this mixin. Mobile has a separate extension for Places, and we want to allow customers to only include PlaceContext in their schemas if they are using the Places extension. Our intention is to define narrow mixins for our customers based on the mobile extensions they are using.

extensions/adobe/experience/mobile-lifecycle-details.schema.json
```diff
@@ -0,0 +1,288 @@
...
+          "description": "Application information related to an ExperienceEvent, including the name of the application, app version, launches, and closures."
+        },
+        "xdm:device": {
+          "title": "Device",
+          "$ref": "https://ns.adobe.com/xdm/context/device",
+          "description": "Information about the mobile device the application lifecycle event observation occurred in, including device type and model."
+        }
+      }
+    },
+    "environment": {
```
mjasrotia, line 31: @kevinlind , as you mentioned that you do not want to include the browserdetails properties from environment, this definition here will bring that as its referring the "environment" data type. Is that a requirement ?

kevinlind: @mjasrotia Thanks for pointing this out. No, that is not a requirement. Trying to define specific fields from a datatype into a mixin is new to me, so I'm a little unclear on the format. The end result should be what I outlined in `mobile-lifecycle-details.example.1.json` where I can use specific environment fields, but not browserDetails.
```
{
  "xdm:environment": {
    "xdm:type": "application",
    "xdm:carrier": "Verizon",
    "xdm:operatingSystem": "iOS",
    "xdm:operatingSystemVersion": "14.2",
    "xdm:operatingSystemVendor": "Apple",
    "dc:language": "en-US"
  }
}
```

I'll remove this from the PR.

extensions/adobe/experience/mobile-lifecycle-details.schema.json
```diff
@@ -0,0 +1,237 @@
...
+      }
+    }
+  },
+  "allOf": [
+    {
+      "$ref": "https://ns.adobe.com/xdm/common/extensible#/definitions/@context"
+    },
+    {
+      "$ref": "#/definitions/mobile-lifecycle-details"
+    }
```
mjasrotia, line 234: In order to have all the properties defined available on UI to compose schema, you will have to add all the definitions in this allOf section. The way you added "$ref": "#/definitions/mobile-lifecycle-details". This definition will only expose the two properties defined under this.

extensions/adobe/experience/mobile-lifecycle-details.schema.json
```diff
@@ -0,0 +1,279 @@
...
+          }
+        }
+      }
+    },
+    "environment-operatingSystem": {
+      "properties": {
+        "xdm:environment": {
+          "type": "object",
+          "properties": {
+            "xdm:operatingSystem": {
```
mjasrotia, line 50: @kevinlind , $ref is used to refer a schema or a definition from the local or an external schema. In this case you are cloning the properties from the "environment" data type schema and they need to be cloned as such (the way they are defined in the data type). you are using reference as "https://ns.adobe.com/xdm/context/environment#/definitions/operatingSystem"
which is wrong as "type" is not defined as a definition in the environment schema. Its just a scalar property. So, in order to define that here you would have use the same property here as well like below:
    "environment-operatingSystem": {
      "properties": {
        "xdm:environment": {
          "type": "object",
          "properties": {
             "xdm:operatingSystem": {
             "title": "Operating system",
             "type": "string",
             "description": "The name of the operating system used when the observation was made. The attribute should 
               not contain any version information such as '10.5.3', but instead contain 'edition' designations such as 
               'Ultimate' or 'Professional'."
               },
The same comment applies to all the properties here

kevinlind: Thanks for the clarification. My concern with cloning the scalar in the mixin is if the property is changed in the Environment schema, such as a new value added to the connectionType enum, then the cloned property in the mixin won't automatically update. It sounds like that is unavoidable unless the Environment schema is changed to make each property a separate definition (or at least the properties which can change such as the enum types.).

I'll update the properties for this mixin as you suggested. However, is it possible to change the Environment schema to define some or all of the properties as definitions so they may be referenced with $ref, or is that considered a breaking change?

mjasrotia: @kevinlind , your concern is a very valid concern and we are aware of it. We have plans to make these properties as definitions in future but before that, we have to come up with a best strategy of doing it with minimal impact on the schemas.

kevinlind: That's great to hear. I'm wondering if I should change my definition of the Mobile Application schema to declare each property as its own definition? I don't have any use for that now, but it would be nice to have the flexibility for any future need.

mjasrotia:
@kevinlind , If you are OK with the the new PR  (https://github.com/adobe/xdm/pull/1071), please close this PR

kevinlind:
Closing as changes are incorporated into PR #1071

PR #1043, opened 2020-11-10
Adding a test mixin for demographic details with each property as a definition

mjasrotia:
Please link to the issue #…

### DIFF
--- a/components/mixins/profile/profile-person-details-v2.schema.json
+++ b/components/mixins/profile/profile-person-details-v2.schema.json
@@ -1,0 +1,287 @@
+{
+  "meta:license": [
+    "Copyright 2020 Adobe Systems Incorporated. All rights reserved.",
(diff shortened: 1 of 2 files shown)


components/mixins/profile/profile-person-details-v2.schema.json
```diff
@@ -0,0 +1,287 @@
...
+        }
+      }
+    }
+  },
+  "allOf": [
+    {
+      "$ref": "https://ns.adobe.com/xdm/common/extensible#/definitions/@context"
+    },
+    { "$ref": "#/definitions/person" },
+    { "$ref": "#/definitions/person-birthDate" },
```
cmathis, line 272: The definition on line 271 already brings in the xdm:person fields with all of its sub-fields. You don't need all these other ref's in the allOf. You only need #/definitions/person

mjasrotia: Done

cmathis (approved):

PR #1068, opened 2020-11-25
Add tags property to message interaction mixin

ctrubian:
The tags property is an array of strings that contains the tags that have been applied to the link being interacted with.


sawhneyh:
@ctrubian : Approving on a pretext as this is a temporary fix till we start ingesting the url-id and corresponding tag in some lookup table in AEP.
Ideally we should also gets this reviewed this with UPS team as we are now passing the tags with the url in de-normalize way and potentially bloating the ExpEvents which are also getting saved in UPS cosmo.  Earlier they have recommend ingesting attributes in normalize fashion.

ctrubian:
@sawhneyh I'll ask Raghav to take a look, but there is no plan to ever store tracker data in the data lake and have a lookup table whose key is the link ID.  Once a tag management system exists they can provide lookup tables, but I'm sure they'll want to key them off IDs they manage.  So whether the event contains a string that's the value, or a string that's the ID there is no difference in data size.  Given that the ID will likely be a UUID the event is likely some bytes smaller right now than it will be in the future.

ctrubian:
@cmathis please review

sawhneyh:
Hi @ctrubian 

> I'll ask Raghav to take a look, but there is no plan to ever store tracker data in the data lake and have a lookup table whose key is the link ID. 

Any reason why we are not evaluating this route, as with this we can avoid sending the tag-id or tag name with each tracking event.
Though we don't have any concern with sending the tags with an event, but this is an optimization which we shouldn't discard.

> So whether the event contains a string that's the value, or a string that's the ID there is no difference in data size. Given that the ID will likely be a UUID the event is likely some bytes smaller right now than it will be in the future.

I agree, sending tag-id or tag-name is almost same. The way to avoid sending id/name is by only having a lookup table with 1.

1.  key as ` url-id`
2.  Value as ` corresponding tags` associated with that url.

prabhum2:
@ctrubian - Please let us know when this is ready to go into registry

ctrubian:
> Any reason why we are not evaluating this route, as with this we can avoid sending the tag-id or tag name with each tracking event.

I should say Tracker has no plans to do this.  Tags will be applicable to many objects beyond links, and there is no reason to create a coupling in tracker where it has to understand what a tag is beyond it being some metadata.  When link management is tackled that service can own managing tags and any lookup tables for them.  If they decide they want to structure the lookup tables where something like a tracker link ID is the key then that's their choice, but I suspect using an ID managed by link management will be more efficient.

ctrubian:
@prabhum2 We're ready.  Thanks

prabhum2 (approved):

PR #1092, opened 2020-12-22
add logic for uber schema

jwen-adobe:
Please link to the issue #…


bin/xed-validation/tempConverter.js
```diff
@@ -100,6 +100,9 @@ class Converter extends EventEmitter {
           }
         };
 
+        //special fixes for mergeAllof resolver, can also be fixed in schema level
+        if (i == "xdm:probabilityPercentage") o[i].type = "number";
```
cmathis, line 104: Don't know what schema this is from, but if it's missing a "type" attribute then let's fix it in the schema and include that in this PR. No reason to have a hardcoded bandaid here.

jwen-adobe: changed.

cmathis (approved):
Looks good

PR #1104, opened 2021-01-12
Migrate decision object schemas to decisioning namespace

jbeckert:


cmathis:
https://wiki.corp.adobe.com/display/DMSArchitecture/Content+Data+Model+and+Content+Decisioning

extensions/adobe/experience/decisioning/calendar-constraint-details.schema.json
```diff
@@ -0,0 +1,39 @@
...
+  "$id": "https://ns.adobe.com/experience/decisioning/calendar-constraint-details",
+  "meta:extensible": true,
+  "type": "object",
+  "title": "Calendar Constraint Details",
+  "description": "A calendar constraint is a component of a decision option that defines the date range when option is valid. Outside that date range the option cannot be proposed.",
+  "definitions": {
+    "calendar-constraint-details": {
+      "type": "object",
+      "properties": {
+        "xdm:startDate": {
```
cmathis, line 18: Given the namespace of this new component, let's name this field https://ns.adobe.com/experience/decisioning/startDate

jbeckert: This is a datatype schema (ie. a separate schema file) and using those schemas for properties in mixins or other datatype schemas requires to use a property prefix that gets erased. Otherwise I see doubly nested paths like "_experience.decisioning.myProp._experience.decisioning.datatypeProp1"

cmathis: In AEP it will be _experience.decisioning.myProp.datatypeProp1. I understand the XED validator in the PR check is not doing this, but it will in AEP. 

The goal here is to not have XDM namespaced components sitting out in the extension folder. If you need it to be XDM that is fine, but let's put those under the standard components/datatypes location then.

cmathis (commented):
LGTM

extensions/adobe/experience/decisioning/option-details.schema.json
```diff
@@ -0,0 +1,91 @@
...
+      "$ref": "#/definitions/option-snapshot-identifier"
+    },
+    {
+      "$ref": "#/definitions/option-snapshot-tags"
+    },
+    {
+      "$ref": "#/definitions/option-snapshot-measures"
+    }
+  ]
+}
```
mjasrotia, line 91: meta:status missing

extensions/adobe/experience/decisioning/placement-details.schema.json
```diff
@@ -0,0 +1,76 @@
...
+            "description": "[Multipurpose Internet Mail Extensions (MIME)](https://tools.ietf.org/html/rfc2046)",
+            "pattern": "\\w+\\/[-.\\w]+(?:\\+[-.\\w]+)?",
+            "description": "A constraint for the media type of the components that is expected in that placement. There could be more than one media type possible for one component such as different image format."
+          }
+        }
+      }
+    },
+    "placement-snapshot-identifier": {
+      "properties": {
+        "https://ns.adobe.com/experience/decisioning/id": {
```
mjasrotia, line 53: If this kind of id is defined in multiple schemas and within different definitions, can we give them a sematic name like placemantID ? Also, "id" should be upper case "ID". This comment applied to all the IDs defined in this PR. Could you please check ?

extensions/adobe/experience/decisioning/placement.schema.json
```diff
@@ -0,0 +1,33 @@
...
+  },
+  "allOf": [
+    {
+      "$ref": "https://ns.adobe.com/xdm/data/record"
+    },
+    {
+      "$ref": "#/definitions/placement"
+    },
+    {
+      "$ref": "https://ns.adobe.com/experience/decisioning/placement-details#/definitions/placement-details"
```
mjasrotia, line 29: Seems like, placement-details would always be used with placement as the name signifies. Should "placement-details " even be a data type ? Can we add those properties here  ?

extensions/adobe/experience/decisioning/ranking.schema.json
```diff
@@ -0,0 +1,35 @@
...
+        }
+      }
+    }
+  },
+  "allOf": [
+    {
+      "$ref": "#/definitions/ranking"
+    }
+  ],
+  "meta:status": "experimental"
```
mjasrotia, line 34: Does this need to be "experimental" ?

extensions/adobe/experience/decisioning/ranking-details.schema.json
```diff
@@ -0,0 +1,56 @@
...
+  },
+  "allOf": [
+    {
+      "$ref": "#/definitions/ranking-details"
+    },
+    {
+      "required": ["https://ns.adobe.com/experience/decisioning/priority"]
+    }
+  ],
+  "meta:status": "experimental"
```
mjasrotia, line 55: Does this need to be "experimental" ?

extensions/adobe/experience/decisioning/placement.schema.json
jbeckert: In several other snapshots placement-details are captured within those datatypes, often in arrays. We need that reusable field set. But what could be done for all the entity types is the following. They could absorb the definitions, keep them separate and when another composite datatype needs them we can go after the -details definition of the entity schema. The downside is that it makes it impossible for custom schemas to reuse those definitions because they will not become a bonafide datatype that can be reused at the logical level. The file and "definitions" structure is invisible after the registry has composed the schemas

extensions/adobe/experience/decisioning/ranking.schema.json
jbeckert: Yes, it fits the definition https://github.com/adobe/xdm/blob/master/docs/status.md 
"stabilizing" would also work but its use is discouraged

extensions/adobe/experience/decisioning/ranking-details.schema.json
jbeckert: Yes, it fits the definition https://github.com/adobe/xdm/blob/master/docs/status.md 
"stabilizing" would also work but its use is discouraged

mjasrotia: The reason I asked this is that the experimental schemas are not available on UI in INT/STG/PROD (except through API calls).Since this is an internal solution and  also all other schemas are "stable", you can convert this to "stable" too. Otherwise, you will have get your IMSOrgs added to the allowed list to be able to see these schemas on UI while composing schemas. The same argument hold good for the other "experimental" schemas too

extensions/adobe/experience/decisioning/ranking.schema.json
mjasrotia: Please see the previous comment

extensions/adobe/experience/decisioning/ranking-details.schema.json
jbeckert: I updated the meta:status to allow all orgs to see the schemas. We have workflows that can't distinguish between Beta orgs and everyone else. The schemas will be needed for all production orgs.

extensions/adobe/experience/decisioning/ranking.schema.json
jbeckert: I updated the meta:status to allow all orgs to see the schemas. We have workflows that can't distinguish between Beta orgs and everyone else. The schemas will be needed for all production orgs.

mjasrotia (approved):

prabhum2 (approved):

PR #1115, opened 2021-01-21
updated opportunities close date

prabhum2:
makes two changes to the opportunity-details mixin
1. Rename the current “closeDate” field of XDM Business Opportunity Details mixin to “expectedCloseDate”.
2. Introduce a new “actualCloseDate”” field in the same mixin, and set its data type as DateTime


components/mixins/opportunity/opportunity-details.schema.json
```diff
@@ -82,12 +82,18 @@
           "minimum": 0,
           "default": 0
         },
-        "xdm:closeDate": {
+        "xdm:expectedCloseDate": {
           "title": "Close Date",
```
mjasrotia, line 86: Can we update the titles for both these properties ?

mjasrotia (approved):

peichiang (approved):

PR #1117, opened 2021-01-22
Fix decision object schema namespaces

jbeckert:
Deprecate datatype schemas and inline them in the class or mix-ins as definitions 
Move examples for deprecated schemas to the embedded schemas where it makes sense


jbeckert:
Closing this PR in favor of [PR 1119](https://github.com/adobe/xdm/pull/1119)

PR #1120, opened 2021-01-25
Fix decision object schema namespaces

jbeckert:
Convert two existing schemas to non-UI, global dataset schemas and move their properties to the correct namespace


extensions/adobe/experience/offer-management/offer-placement.schema.json
```diff
@@ -57,10 +66,13 @@
...
     {
-      "$ref": "#/definitions/offer-placement"
+      "$ref": "https://ns.adobe.com/xdm/data/record"
+    },
+    {
+      "$ref": "https://ns.adobe.com/experience/decisioning/placement"
     },
     {
-      "required": ["xdm:name", "xdm:channel", "xdm:componentType"]
+      "$ref": "#/definitions/offer-placement"
```
mjasrotia, line 75: @jbeckert , This schema is already deprecated and continues to be so. You are still making changes to the allOF section here. Is there a need for this or am I missing something.

jbeckert: It was a mistake to mark the status as deprecated. What needed to be deprecated were the "xdm:" properties in this schema

jbeckert: Let me amend the schema and bring back the status to "stable"

mjasrotia: Sounds good

mjasrotia (approved):

prabhum2 (approved):

Issue #1122, opened 2021-01-26
Travel & Hospitality Mixins & Data Types


jbranson000:
Creation of all travel & hospitality industry mixins and datatypes.

PR #1125, opened 2021-01-26
Loyalty Mixin

jbranson000:
#1122

Includes Loyalty mixins:
Loyalty Details


mjasrotia:
Can we come with a common top level object for vehicle/room/property comparisons ? Something like "assetComparison" or "rateComparison" or "searchComparison" ? If yes, we can follow the same approach for PR #1125 as well where we are creating three mixins with three different high level objects

jbranson000:
Closing and combining with another PR.

PR #1138, opened 2021-02-03
commerce datatype for industry mixin

nishithshah7:
Please link to the issue #…


mjasrotia:
@nishithshah7 , The commerce data type is used in one of the experience event mixins and adding all the properties at the root here will explode the uber schema which is already using the commerce data type. For all the nested objects you have added to this data type (checkouts, productListTotal, wishList, productReviewed and others,) , I would suggest to add them either as its own local definitions within the commerce data type so that they could be used in parts if required in other mixins or, if possible, create new data types if they make sense and have the potential to be re-used across mixins.

Issue #1147, opened 2021-02-10
New Segment Status called "REVOKED" required in Segment Definition for external Segments


knarayanan88:
## What are the schemas that are affected by the issue
Segment Definition

## What are examples of products that are impacted by the issue
AEP Segment Match/Interchange, AEP Ansible

PR #1148, opened 2021-02-10
Add REVOKED segment status to signal a segment that was shared is revoked

knarayanan88:
Adding REVOKED segment status to explicitly revoke a segment from the sharee so that Ansible/Segment Builder can act differently for such segments.

Please link to the issue 

Issue Tracked => #1147

### DIFF
--- a/components/classes/segmentdefinition.schema.json
+++ b/components/classes/segmentdefinition.schema.json
@@ -43,8 +43,8 @@
         "xdm:segmentStatus": {
           "type": "string",
           "title": "Segment status",
-          "description": "Current status of segment from external system normalized to active, inactive, deleted or draft.",
-          "enum": ["ACTIVE", "INACTIVE", "DELETED", "DRAFT"]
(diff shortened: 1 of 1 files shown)


mjasrotia (approved):

shieldz:
Thanks, this will unblock the 2nd-party audience sharing workflows, where the owner of the segment comes from a separate IMS Org.  It will also be used when promoting/revoking access to segment definitions between sandboxes.  @cmathis let me know if you have additional questions.

HyveMynd:
@shieldz Is there a difference between `INACTIVE` and `REVOKED`? Is there a reason we should add another status and not reuse what is there?

shieldz:
Yes, and it has to do with user roles.  If a segment comes from a different org or sandbox, an admin user from the receiving org/sandbox shouldn't be able to "reactivate" a REVOKED segment, but they could reactivate an INACTIVE one

cmathis (approved):

PR #1149, opened 2021-02-10
CORE-55883: XDM schema addition for expressing time association between identities

knayyaradobe:
Please link to the issue # https://jira.corp.adobe.com/browse/CORE-55883


balachan354:
Requirements for this change are documented at https://wiki.corp.adobe.com/display/DMSArchitecture/XDM+schema+changes+for+Shared+devices

cmathis:
@knayyaradobe - A couple questions:
1. Is this new "associatedIdentities" something that is expected to be populated before sending data to AEP, or is it a field populated by an internal process and populated during export?
2. In this PR you are creating two new datatypes for the purpose of referencing those under this new "associatedIdentities". Do you ever see these datatypes being used outside of this one use case? If not then do NOT create new datatypes - simply nest those fields under their parent field and avoid the whole $ref thing. Only create new datatypes when you expect many other uses of it.

knayyaradobe:
1. "associatedidentities" will be populated  by an internal process and populated during export
2. This is not foreseen to be used by any other field, will nest them in the original file.
3. The new field associatedidentities has a different structure than the current identityMap and will be used by the Shared devices feature to show time history association for devices with individuals

cmathis:
@knayyaradobe 

Given this is a system generated value and I'm assuming this "system" is Unified Profile, then let's add this new field to the existing UPS mixin that defines the system generated values. That would be this file here - https://github.com/adobe/xdm/blob/master/extensions/adobe/experience/profile/profile-all.schema.json

This "profile-all" mixin is automatically added to all Profile union views. 

This new field would have the name of "https://ns.adobe.com/experience/profile/associatedIdentities" and sit next to the current "identityNamespaces" that you will see there.

knayyaradobe:
@cmathis  I have made the suggested changes, please review.

extensions/adobe/experience/profile/profile-all.schema.json
```diff
@@ -28,6 +28,44 @@
           "additionalProperties": {
             "$ref": "https://ns.adobe.com/xdm/context/namespace"
           }
+        },
+        "xdm:associatedIdentities": {
```
cmathis, line 32: Field name should be https://ns.adobe.com/experience/profile/associatedIdentities

cmathis (approved):

prabhum2 (approved):

PR #1164, opened 2021-02-20
New generic descriptorRelationship

cmathis:
Introduces a new descriptor for all relationships on AEP. This will eventually replace the existing descriptorOneToOne, descriptorOneToMany, etc. Initially this will act as a single relationship descriptor that supports both M:1 and 1:1 defined via a "xdm:cardinality" attribute.

### DIFF
--- a/schemas/descriptors/relationshipdescriptor.schema.json
+++ b/schemas/descriptors/relationshipdescriptor.schema.json
@@ -82,7 +82,11 @@
           "default": "M:1"
         }
       },
-      "required": ["xdm:sourceProperty", "xdm:destinationSchema", "xdm:destinationVersion"]
+      "required": [
+        "xdm:sourceProperty", 
+        "xdm:destinationSchema", 
+        "xdm:destinationVersion"
+      ]
     }
   },
   "allOf": [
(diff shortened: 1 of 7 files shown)


schemas/descriptors/relationshipdescriptor.schema.json
```diff
@@ -59,15 +59,30 @@
           "type": "string",
           "description": "Provides a user friendly name or label to this relationship"
         },
-        "xdm:apiLabel": {
```
cmathis, line 62: Was added for the B2B implementation on AEP, but was never used.

schemas/descriptors/relationshipdescriptor.schema.json
```diff
@@ -59,15 +59,30 @@
...
-          "title": "API Label",
+        "xdm:sourceNamespace": {
+          "title": "Source Property Namespace",
           "type": "string",
-          "description": "Provides a user friendly name or label which is used in PQL/SQL or other needs"
+          "description": "Identity Service namespace associated with the source property"
         },
-        "xdm:description": {
```
cmathis, line 67: Was added for the B2B implementation on AEP, but was never used.

PR #1165, opened 2021-02-22
Removing owner fields from B2B Entities

mjasrotia:
Also moving the "b2b-account" datatype attributes at the root of "XDM Business Account Details" MIXIN


peichiang (commented):
The change looks good to me except for the pending issue of whether to completely get rid of the “Account Organization” data type. Waiting for Liz's confirmation.

peichiang (approved):

prabhum2 (approved):

PR #1180, opened 2021-03-04
Consent subscription support for CJM

RussStringham:
https://jira.corp.adobe.com/browse/CJM-7442


RussStringham:
In the previous version of the XDM, the "marketing.preferred" and "marketing.any" objects could appear in the main section, "consents.marketing.preferred/any" and in the idSpecific section, such as "consents.idSpecific.email.john@xyz.com.marketing.preferred/any". We have no use cases where these should be specified in the idSpecific section, and no data that has been ingested into AEP where this is the case. In my update, these two fields are only allowed to appear in the main section.

components/datatypes/consent-preferences.schema.json
```diff
@@ -187,9 +244,10 @@
...
           "title": "Consents and Preferences",
           "description": "Specific Consent and Preference Options",
+          "type": "object",
           "allOf": [
             {
-              "$ref": "#/definitions/choices"
+              "$ref": "#/definitions/user-level-choices"
             },
             {
               "properties": {
```
prabhum2, line 253: is this an empty property block?

components/datatypes/consent-preferences.schema.json
```diff
@@ -113,46 +154,6 @@
           "description": "User's Direct Marketing Preferences",
           "type": "object",
           "properties": {
-            "xdm:preferred": {
```
prabhum2, line 116: @RussStringham - Can you please deprecate these properties instead of deleting them .

components/datatypes/consent-preferences.schema.json
RussStringham: No, you need to expand the view (by default the diff was not showing 8 lines, which made it appear like an empty block).

components/datatypes/consent-preferences.schema.json
RussStringham: These properties aren't being (fully) deleted. The actual XDM is defined using ref macros, and they are being incorporated into the same area from a different macro. With the original definition, the macros exposed these fields in two places. However, they should only be used in one of those places. This update makes sure they are only exposed in that one place (they have never been populated with data in the other place). See my 2nd comment in the "conversations" section of this pull request for more details.

components/mixins/profile/profile-consents.example.1.json
```diff
@@ -54,6 +79,16 @@
         "john@xyz.com": {
           "xdm:marketing": {
             "xdm:email": {
```
RussStringham, line 81: Previously, at this place in the idSpecific section, I could have specified the "xdm:preferred" and "xdm:any" fields. With this update they are no longer allowed here, but as this example file shows, they are still supported above at lines 15 and 16.

RussStringham: Actually this comment should have been on line 80 instead of 81

prabhum2:
@RussStringham - is there anyone else needed to be reviewed and be approving this ?

RussStringham:
The PR requires approval from the UPS and CJM teams

guayang-adobe (approved):

guayang-adobe:
LGTM for CJM use cases

rkpandey (approved):
LGTM

PR #1186, opened 2021-03-11
Remove "experimental" status from decision strategy details

jbeckert:
- Change the text strings of the strategy's step enum 
- Add examples for strategy steps and scope details
- Add qualifier property to ranking details, now that rankingStrategy as a second option is supported
- Add repository ID to the standard content properties to help resolve the asset ID when multiple repositories per org are supported
- Remove duplicated namespace prefixes from property names in example to match the schema


cmathis:
@jbeckert  - What happened to the idea of moving these "decision" schemas into the standard XDM namespace and copied into the "components" folder.

jbeckert:
That attempt was stalled. There needs to be a top down consensus with Product before we can attempt that.

mjasrotia (approved):

PR #1190, opened 2021-03-15
Removed the inferred person mixin

mjasrotia:
https://jira.corp.adobe.com/browse/PLATIR-13072


prabhum2 (approved):

cmathis (approved):
This will break two schemas that will need to be fixed.

PR #1193, opened 2021-03-17
Miscellaneous field updates

jbranson000:
Includes field updates to the Loyalty Mixin, and minor field naming convention updates to social network usage, form applications, and user account details.

### DIFF
--- a/components/datatypes/industry-verticals/form-applications.schema.json
+++ b/components/datatypes/industry-verticals/form-applications.schema.json
@@ -14,54 +14,54 @@
   "definitions": {
     "form-applications": {
       "properties": {
-        "xdm:formApplicationApproved": {
+        "xdm:approved": {
           "title": "Form Application Approved",
           "$ref": "https://ns.adobe.com/xdm/data/measure",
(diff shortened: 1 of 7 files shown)


components/datatypes/industry-verticals/form-applications.schema.json
```diff
@@ -14,54 +14,54 @@
...
         },
-        "xdm:formApplicationDeclined": {
+        "xdm:declined": {
           "title": "Form Application Declined",
           "$ref": "https://ns.adobe.com/xdm/data/measure",
           "description": "Captured when a visitor application has been declined."
         },
-        "xdm:formApplicationFee": {
-          "title": "Form Application Fee",
+        "xdm:applicationFee": {
```
mjasrotia, line 27: Given the changes you did here, should this be just "fee" ?

jbranson000: I purposely left it as "application fee", because that made more sense in context rather than just "fee".

mjasrotia (approved):

prabhum2 (approved):

PR #1194, opened 2021-03-17
Initial commit, adding multiple new commerce fields.

jbranson000:
This includes multiple field additions to the commerce mixin.


components/datatypes/marketing/commerce.schema.json
```diff
@@ -40,28 +40,317 @@
...
+          "title": "Product Brand",
+          "type": "string",
+          "description": "Tracks the brand of an individual product."
+        },
+        "xdm:cartID": {
+          "title": "Cart ID",
+          "type": "string",
+          "description": "If applicable, tracks a persistent cart identifier string that would expire upon a successful order event."
+        },
+        "xdm:category": {
```
mjasrotia, line 92: category is too generic. Should we rename this to "productCategory" like "productBrand" ? Also, Do we have to add these properties to the event feed ? Don't they belong to the Product Class and its MIXINS ? We already have a ProductCategory mixin.

components/datatypes/marketing/commerce.schema.json
```diff
@@ -40,28 +40,317 @@
...
+          "title": "Checkout Step",
+          "type": "string",
+          "description": "Tracks the name of a single step in the checkout, if applicable."
+        },
+        "xdm:checkoutType": {
+          "title": "Checkout Type",
+          "type": "string",
+          "description": "Different types of checkout types that could be used by the visitor, i.e. 'standard' or 'express'."
+        },
+        "xdm:costOfGoodsSold": {
```
mjasrotia, line 117: Should this be a currency ?

components/datatypes/marketing/commerce.schema.json
```diff
@@ -40,28 +40,317 @@
...
+          "title": "Split Payment Methods",
+          "type": "boolean",
+          "description": "Captured if a visitor chooses to use multiple payment methods (i.e. credit card & points) to complete a purchase."
+        },
+        "xdm:storeId": {
+          "title": "Store ID",
+          "type": "string",
+          "description": "Store ID of BOPIS or BOSS store."
+        },
+        "xdm:subcategory": {
```
mjasrotia, line 329: See the above comment for productCategory

components/datatypes/marketing/commerce.schema.json
jbranson000: I've updated these.  For the legacy commerce mixin, we will likely need these as some clients will pass them through in the experience event rather than the classification.  Mitch's forthcoming updates might address some of that.

prabhum2:
will create a new PR

Issue #1196, opened 2021-03-19
Identity Graph Flattened Export Schema 


knarayanan88:
Global Identity Graph Export Schema in Flattened Identity-Centric consumer format.

## What are the schemas that are affected by the issue
None
## What are examples of products that are impacted by the issue
None

PR #1197, opened 2021-03-19
Identity Graph Flattened Export Schema

knarayanan88:
Please link to the issue #1196 
This schema is supposed to be a global schema across tenants for writing out flattened identity-centric consumer schema.
Each row is unique based on composite key: identityId, namespaceCode, algo.

The membersByTimeRange is a nested object that contains the history of member identities in the same person graph across different timeRanges. Time Ranges switch when there is a shared device scenario where a same shared device is used by multiple people at different points of time.

Purpose of this schema is for Siphon to automate creation of custom partitioning/bucketing based on the schemaId of this global schema.


schemas/identity/identity-graph-flattened-export.schema.json
```diff
@@ -0,0 +1,108 @@
...
+                }
+              }
+            }
+          }
+        },
+        "xdm:lastUpdatedTime": {
+          "title": "Last Updated Time",
+          "type":"string",
+          "format":"date-time",
+          "description": "Last updated time when this identity was updated."
```
andreitaleanu, line 90: Is this column something we could leverage when computing the deltas in the future? For instance, instead of joining both the old & new snapshots, maybe we could load only the newer snapshot and look at the `lastUpdatedTIme`?

schemas/identity/identity-graph-flattened-export.schema.json
```diff
@@ -0,0 +1,108 @@
...
+        "xdm:membersByTimeRange": {
+          "type": "array",
+          "items": {
+            "type": "object",
+            "properties": {
+              "xdm:startTs": {
+                "title": "Start Timestamp",
+                "type": "integer",
+                "minimum": -9223372036854776000,
+                "maximum": 9223372036854776000,
```
andreitaleanu, line 58: Are these intented to be java.Long.MIN_VALUE / java.Long.MAX_VALUE?

andreitaleanu (approved):
👍 This looks good as far as I can tell, but it's hard to say at this point whether we'll need to evolve the schema in the near future (at least from CJA Stitching's point of view). Let's unblock testing and move forward with this and come back to address any underlying concerns down the road, if required.

schemas/identity/identity-graph-flattened-export.schema.json
knarayanan88: Yes. It is max/min long values.

schemas/identity/identity-graph-flattened-export.schema.json
knarayanan88: Yes, for first phase using full daily export, this may not be useful. But when we move to incremental exports in second phase, we will use this field to help Siphon handle Out-of-order processing of messages.
Platform-SDK will provide interfaces to get deltas based on time-stamps.

cmathis (approved):

schemas/identity/identity-graph-flattened-export.schema.json
andreitaleanu: Cool, that makes sense :)

prabhum2 (approved):

PR #1199, opened 2021-03-19
Define datatypes for basic consent fields

RussStringham:
Helpful to allow customers to extend the XDM with their own consents and preferences


cmathis (approved):
Looks good. 

Many of the new datatypes are setup to allow a user to extend the consent structure later.

PR #1211, opened 2021-03-25
ExperienceEvent producedBy attribute

cmathis:
Fields used by the data producer to specify if the event was generated by the individual themselves or a system/service that generated it for the individual. 
For example: Individual opened an email would be an event where the producedBy="self", system sent an email to the individual would be an event where the producedBy="system"


components/classes/experienceevent.schema.json
```diff
@@ -24,8 +24,18 @@
...
+          "description": "An ID to correlate or merge multiple Experience events together that are essentially the same event or should be merged. This is intended to be populated by the data producer prior to ingestion.",
+          "type": "string"
+        },
+        "xdm:producedBy": {
+          "title": "Produced By",
+          "description": "Value that describes the producer of the event - suggested values would be 'self', 'system', 'salesRep', etc. Can be used to filter out certain producers if needed.",
           "type": "string",
-          "description": "An ID to correlate or merge multiple Experience events together that are essentially the same event or should be merged. This is intended to be populated by the data producer prior to ingestion."
+          "default": "self",
+          "meta:enum": {
```
cmathis, line 35: Note the "meta:enum" is used to provide a list of values that are selectable in a UI dropdown. It does not constrain the field to these values only - any other values can be provided.

prabhum2 (approved):

PR #1217, opened 2021-03-30
Added email namespace in message execution schema

sauagarw:


docs/reference/adobe/experience/customerJourneyManagement/messageexecution.schema.json
```diff
@@ -52,6 +52,11 @@
                     "title": "Journey Action ID",
                     "type": "string",
                     "description": "Journey Action ID, for which MessageExecution is triggered."
+                },
+                "https://ns.adobe.com/experience/customerJourneyManagement/messageExecution/emailNamespace": {
```
sawhneyh, line 56: Why we require this field to be a part of Feedback/ExpEvent event. Can't consent service makes call to Message runtime to find the emailNamespace for corresponding MessageExecution?

guayang-adobe: Message Runtime already populates this field in "pid" payload in most URLs including unsubscribe links, which applies to the link use case for list-unsubscribe. Let's keep it consistent and provide it in the mailto use case as well.

guayang-adobe:
I feel the field is more suitable in messageProfile, i.e. 
new field "https://ns.adobe.com/experience/customerJourneyManagement/emailChannelContext/namespace" in https://github.com/adobe/xdm/blob/master/docs/reference/adobe/experience/customerJourneyManagement/messageprofile.schema.json.

guayang-adobe (approved):
I feel the field is more suitable in messageProfile, i.e. 
new field "https://ns.adobe.com/experience/customerJourneyManagement/emailChannelContext/namespace" in https://github.com/adobe/xdm/blob/master/docs/reference/adobe/experience/customerJourneyManagement/messageprofile.schema.json. But I'm fine if we decide to put it in here.

docs/reference/adobe/experience/customerJourneyManagement/messageexecution.schema.json
sauagarw: @sawhneyh  Consent service does not directly interface with message runtime. All the message context is passed to them from external sources. MR passes the required metadata to ESS-EFS which then flows to downstream systems.
@guayang-adobe  Even if pid is passed in mailTo, EFS still has to process it & set the fields in xdm as communication between consent & efs is via xdm event. So, we would still need to introduce a field for email namespace.

sawhneyh: The reason I am not comfortable using XDM field for communicating the emailNamespace to the consent service, is that this data will also be ingested in AEP. The end user can also view this data, and it may be confusing why this field exist in XDM.

docs/reference/adobe/experience/customerJourneyManagement/messageprofile.schema.json
```diff
@@ -47,6 +47,11 @@
                         "apns": "Apple Push Notification service",
                         "fcm": "Firebase cloud messaging service"
                     }
+                },
+                "https://ns.adobe.com/experience/customerJourneyManagement/emailChannelContext/namespace": {
```
sawhneyh, line 51: Minor comment, It will be more readable if we add this new field after line 41, so that all related fields are together in this file.

sawhneyh (approved):

chetanmeh (approved):

sauagarw:
@cmathis The PR is ready to be merged.

PR #1231, opened 2021-04-08
Plat 94752

jbranson000:
Contains all updates for PLAT-94752


components/mixins/experience-event/experienceevent-card-actions.schema.json
```diff
@@ -19,7 +19,7 @@
   "definitions": {
     "card-actions": {
       "properties": {
-        "xdm:personalFinances": {
+        "xdm:cardActions": {
```
hanessian, line 22: @jbranson000 - For this, I think we wanted to leave it under personal finances, but add an additional cardActions root node beneath it. So it should be personalFinances.cardActions.actions here. That way cardactions still falls under personalFinances, but we don't have all those rogue fields at the root like we did before. Thanks!

hanessian (approved):

mjasrotia (approved):

prabhum2 (approved):

PR #1234, opened 2021-04-09
Th updates 040921a

jbranson000:
Miscellaneous reservation mixin updates


jbranson000:
Reopening with 1235 with correct local branch.

PR #1251, opened 2021-04-23
added commerce,order details schemas

prabhum2:
added new 
- commerce-details mixin extending experience event class
- order-details class
Updated
- product class


prabhum2:
Will create a new one if needed

PR #1272, opened 2021-05-11
Introduced a new mixin to capture the time profiling

jainaks:
The PR is introducing:
1. A new datatype, named time-profiling, for capturing the profiling information for any component/service.
2. A new schema for a mixin, named "message-execution-timeline" to be used for extending the Feedback events generated by Customer Journey Management.
It captures the timeline information of any execution as part of the feedback event and can be used to deduce the E2E latency of execution. 
The linked issue is:  [CJM-9943](https://jira.corp.adobe.com/browse/CJM-9943)


components/datatypes/time-profiling.schema.json
```diff
@@ -0,0 +1,37 @@
...
+  "type": "object",
+  "meta:extensible": true,
+  "description": "This datatype is used where ever you want to store the time profiling information of any service or component.",
+  "definitions": {
+    "time-profiling": {
+      "properties": {
+        "xdm:startTime": {
+          "title": "startTime",
+          "description": "Execution Start time in Epoch millis for the component.",
+          "type": "number"
```
sawhneyh, line 20: rather storing the epoch millis, should we store this in Date-Time format? Data type will be more explicit in way, and we can use the sql date time methods in queries without any typecasting.
 We can specify type as: 

- "type": "string",
- "format": "date-time"

jainaks:
We worked out an alternate approach for the same, the messageExecutionTimeline element will be an array of time-profiling object now.
Each element of the array will have the component code and its timing information.
Following are the major pros of this approach:
1. The internals of the execution pipeline will not be exposed in Schema of the provisioned dataset.
2. Adding a new component will not require any code change, both in schema side and query side.

docs/reference/adobe/experience/customerJourneyManagement/message-execution-timeline.schema.json
```diff
@@ -0,0 +1,93 @@
...
+        },
+        "xdm:serviceGroup": {
+          "title": "serviceGroup",
+          "description": "An optional attribute, to capture the timeline of a service, at sub-tree level.",
+          "type": "string"
+        },
+        "xdm:startTime": {
+          "title": "startTime",
+          "description": "Execution Start time in Epoch millis for the component.",
+          "type": "number"
```
sawhneyh, line 35: I feel, we should be using the date-time format for specifying the start/end date. If we keep type as number, we won't get any automatic validation, whether some one has entered the valid date or not. Number means any negative integer also.
Anyway while querying the data we can convert the date-time format to epoch if required.

Also, I see this as a general practice at other places in XDM: https://github.com/adobe/xdm/blob/cf79e096f7f16270b133b0e8609d04b6faf7cb2f/docs/reference/adobe/experience/decisioning/activity-detail.schema.json

components/datatypes/time-profiling.schema.json
sawhneyh: I feel, we should still be using the date-time format for specifying the start/end date. If we keep type as number, we won't get any automatic validation, whether some one has entered the valid date or not. Number means any negative integer also.

Anyway while querying the data we can convert the date-time format to epoch or use sql datediff function for finding the difference between 2 dates.

Also, I see this as a general practice at other places in XDM: https://github.com/adobe/xdm/blob/cf79e096f7f16270b133b0e8609d04b6faf7cb2f/docs/reference/adobe/experience/decisioning/activity-detail.schema.json

jainaks: @sawhneyh Changed the format of startTime and endTime to date-time.

sawhneyh (approved):

cmathis (approved):

PR #1274, opened 2021-05-11
Update VendorX to be rfc2606 compliant

33tonypapousek:
Anything other than example domains can cause users to visit unsafe domains. An IANA-managed Reserved Domain is generally preferred for this sort of use case.


extensions/README.md
```diff
@@ -13,23 +13,23 @@ but an extension can also introduce a new class as well.
...
     {
-       "$id": "https://ns.vendorx.com/profile-extension",
-       "title": "Vendor X Profile Extension",
+       "$id": "https://ns.example.com/profile-extension",
+       "title": "Example Vendor Profile Extension",
        "description": "Adds the field used by the vendor integration within AEP",
        "meta:intendedToExtend": ["https://ns.adobe.com/xdm/context/profile"], //The class this mixin can be used in
        "properties": {
-          "https://ns.vendorx.com/vendorField": {
+          "https://ns.example.com/vendorField": {
```
cmathis, line 24: Point of this is to show that a vendor named "airship", or "bobsconsulting" would have have a similar namespace of ns.airship.com or ns.bobsconsulting.com. 

ns.example.com does not align with a specific vendor name. This is why we included "vendorx" before as it that was the name of the vendor.

33tonypapousek: That makes sense, but since Adobe doesn't own the vendorx domain, it probably shouldn't be used in this way. Would it be possible to swap it out for an Adobe-owned variant that's safer but gets the point across?

cmathis: Ahh, I had no idea vendorx.com was an actual domain.

Yes, it can be swapped out with whatever. The goal is to simply show that you have some other source/authority that defined this field.

mjasrotia (approved):

PR #1299, opened 2021-06-03
initial rebuild with telecom-subscription datatype and telecom-subscr…

jbranson000:
initial rebuild with telecom-subscription datatype and telecom-subscription-details field grouping


components/mixins/profile/profile-telecom-subscription-details.schema.json
```diff
@@ -0,0 +1,337 @@
...
+  "meta:intendedToExtend": ["https://ns.adobe.com/xdm/context/profile"],
+  "description": "Collects all generic details related to a customer's telecom subscription plan, including pricing, packages, and individual product subscriptions.",
+  "definitions": {
+    "telecom-subscriptions": {
+      "type": "object",
+      "properties": {
+        "xdm:telecomSubscription": {
+          "type": "object",
+          "properties": {
+            "xdm:telecomSubscription": {
```
mjasrotia, line 26: If "https://ns.adobe.com/xdm/datatypes/telecom-subscription" is only referred here and no where else then we should define those properties inline here rather than defining a data type.

mjasrotia:
In general, if we need to have a master subscriptionID for the individual/profile then we should also define one subscriptionID within each subscription array which will allow us to capture multiple subscriptions for the same subscription type.

components/mixins/profile/profile-telecom-subscription-details.schema.json
```diff
@@ -0,0 +1,472 @@
...
+  "meta:extensible": true,
+  "meta:abstract": true,
+  "meta:tags": {
+    "industry": ["telecom"]
+  },
+  "meta:intendedToExtend": ["https://ns.adobe.com/xdm/context/profile"],
+  "description": "Collects all generic details related to a customer's telecom subscription plan, including pricing, packages, and individual product subscriptions.",
+  "definitions": {
+    "telecom-subscriptions": {
+      "type": "object",
```
mjasrotia, line 21: Definition does not need to have a type

components/mixins/profile/profile-telecom-subscription-details.schema.json
```diff
@@ -0,0 +1,472 @@
...
+              "title": "Subscription ID",
+              "type": "string",
+              "description": "Identifier for subscription."
+            },
+            "xdm:subscriptionDetails": {
+              "title": "Subscription Details",
+              "$ref": "https://ns.adobe.com/xdm/context/subscription",
+              "description": "Extension of the subscription data type to include subscription length, fees, status, etc."
+            },
+            "xdm:activationDate": {
```
mjasrotia, line 36: We already have start and end dates in the "https://ns.adobe.com/xdm/context/subscription" datatype. Do we still need activation and deactivation dates here ? Also, if it makes sense, can we move the common properties from these arrays to the subscription data type ?

hanessian (approved):

mjasrotia (approved):

PR #1311, opened 2021-06-10
New JsonSchemas for Label descriptors

veramach-adobe:
Json schemas to support the new label descriptors

@cmathis @prabhum2


schemas/descriptors/label/descriptorLabel.schema.json
```diff
@@ -0,0 +1,43 @@
...
+        }
+      }
+    }
+  },
+  "allOf": [
+    {
+      "$ref": "https://ns.adobe.com/xdm/common/descriptors/schemadescriptor#/definitions/descriptor"
+    },
+    {
+      "$ref": "https://ns.adobe.com/xdm/common/descriptors/labeldescriptor#/definitions/labeldescriptor"
```
cmathis, line 35: No reason to create this external labeldescriptor file. Just add the "xdm:labels" field definition in the local definition here.

schemas/descriptors/label/labeldescriptor.schema.json
```diff
@@ -0,0 +1,72 @@
+{
+  "meta:license": [
+    "Copyright 2021 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/common/descriptors/labeldescriptor",
```
cmathis, line 8: No need for this file. Simply add the "xdm:labels" field definition to the new https://ns.adobe.com/xdm/common/descriptors/descriptorLabel

schemas/descriptors/label/descriptorLabel.schema.json
veramach-adobe: Yes. My bad. There is no reason for that file.

veramach-adobe:
@cmathis Updated with the latest commit

schemas/descriptors/label/descriptorLabel.schema.json
```diff
@@ -0,0 +1,77 @@
...
+              },
+              "labelValue": {
+                "type": "string"
+              }
+            }
+          },
+          "meta:xdmType": "array"
+        }
+      },
+      "oneOf" : [
```
cmathis, line 53: What is the purpose of the "oneOf" here? Are we saying this descriptor can have have either a xdm:labels OR a xdm:sourceProperty? 

Seems like you want both of those to always be required so simply do:
"required": ["xdm:sourceProperty", "xdm:labels"]

veramach-adobe: Yes, Taken care of. I re-used the relationship descriptor schema definitions that defines this way. I should given a closer look.

mjasrotia (approved):

PR #1323, opened 2021-06-21
Add all files from mixins schemas and extension folder for localization

bisin-adobe:
Please link to the issue #…


jwen-adobe:
To get the build pass, you can just add a resolver function to handle the conflict. You can refer to the example of resolver here --> https://github.com/adobe/xdm/blob/727e86d290e3087b6492c5ac50100157dd732211/bin/xed-validation/xedConverter.js#L305

jwen-adobe (approved):

PR #1331, opened 2021-06-30
Change Edge Autofilled Environment Details status to stable

kevinlind:
- Change the Edge Autofilled Environment Details field group status to "stable".
- Create dedicated definitions for each field in the Geo and GeoCoordinates data types
- Reference only the needed field definitions for Geo and GeoCoordinates in the Edge Autofilled Environment Details field group.

Locally on my system, `npm run incompatibility-check` fails because I have removed the references `$ref": "#/definitions/geo` and `$ref": "#/definitions/geocoordinates`. I've searched the repo and did find any direct usages of these references.

### DIFF
--- a/components/datatypes/demographic/geo.schema.json
+++ b/components/datatypes/demographic/geo.schema.json
@@ -15,7 +15,7 @@
   "definitions": {
     "countrycode": {
       "properties": {
-         "xdm:countryCode": {
+        "xdm:countryCode": {
           "title": "Country code",
           "type": "string",
           "pattern": "^[A-Z]{2}$",
--- a/components/datatypes/demographic/geo.schema.json
+++ b/components/datatypes/demographic/geo.schema.json
@@ -13,36 +13,56 @@
   "title": "Geo",
   "description": "The geographic related data where an event was observed.",
   "definitions": {
-    "geo": {
+    "countrycode": {
       "properties": {
-        "xdm:countryCode": {
+         "xdm:countryCode": {
           "title": "Country code",
           "type": "string",
           "pattern": "^[A-Z]{2}$",
           "description": "The two-character [ISO 3166-1 alpha-2](https://datahub.io/core/country-list) code for the country."
-        },
+        }
+      }
+    },
+    "stateprovince": {
+      "properties": {
         "xdm:stateProvince": {
           "title": "State or province",
           "type": "string",
           "description": "The state, or province portion of the observation. The format follows the [ISO 3166-2 (country and subdivision)][http://www.unece.org/cefact/locode/subdivisions.html] standard.",
           "examples": ["US-CA", "DE-BB", "JP-13"],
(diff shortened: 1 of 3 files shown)


components/datatypes/demographic/geo.schema.json
```diff
@@ -56,7 +76,22 @@
       "$ref": "http://schema.org/GeoCoordinates"
     },
     {
-      "$ref": "#/definitions/geo"
+      "$ref": "#/definitions/countrycode"
```
kevinlind, line 79: I've used all lowecase for the definition names, but I was wondering if there is a standard naming format, such as using camelcase or hyphenated names?

extensions/adobe/experience/edge-autofilled-environment-details.schema.json
```diff
@@ -43,15 +43,53 @@
...
+      }
+    },
+    "edge-environment-latitude": {
+      "properties": {
+        "xdm:placeContext": {
+          "type": "object",
+          "properties": {
+            "xdm:geo": {
+              "title": "Geo",
+              "$ref": "http://schema.org/GeoCoordinates#/definitions/latitude"
```
kevinlind, line 79: I'm not sure if this is the correct way to reference a field for a data type which is extended by another data type. In this case, Geo extends GeoCoordinates and the path for this field in this field group should be `placeContext/geo/_schema/latitude`. Does this definition look correct, or do I need another nested property for the `schema` namespace?

extensions/adobe/experience/edge-autofilled-environment-details.schema.json
```diff
@@ -43,15 +43,53 @@
...
               "title": "Geo",
-              "$ref": "https://ns.adobe.com/xdm/common/geo",
-              "description": "The geographic location where the experience was delivered based on the IP address from the client request."
+              "$ref": "https://ns.adobe.com/xdm/common/geo#/definitions/countrycode"
+            }
+          }
+        }
+      }
+    },
+    "edge-environment-postalcode": {
```
flore77, line 59: I think here should be `stateProvince`, we will not be auto filling the postal code.

flore77 (commented):
@kevinlind can we also update the `edge-autofilled-environment-details.schema.md` to contain only the prefilled geo fields? Basically, removing any mention to postalCode and city.

extensions/adobe/experience/edge-autofilled-environment-details.schema.json
kevinlind: Thanks for catching that. I've updated to use state province instead of postal code.

components/datatypes/demographic/geo.schema.json
prabhum2: at this point in time we do not enforce any naming standards on definition names

prabhum2 (approved):

mjasrotia (approved):

PR #1333, opened 2021-07-14
Adding UOM code to the product list items data type

mjasrotia:
https://jira.corp.adobe.com/browse/PLATIR-941


components/datatypes/productlistitem.schema.json
```diff
@@ -56,6 +56,11 @@
           "title": "Price total",
           "type": "number",
           "description": "The total price for the product line item."
+        },
+        "xdm:unitOfMeasureCode": {
```
prabhum2, line 60: any reason why we cannot use existing field group https://github.com/adobe/xdm/blob/master/components/fieldgroups/product/product-measurement.schema.json here instead of creating a new property for the same concept

prabhum2 (approved):

PR #1336, opened 2021-07-16
[PLATO-1671] - Standard Adobe Campaign Schema for Broad and Tracking Logs

jishah-adobe:
Please link to the issue #…


prabhum2:
@jishah-adobe please let us know once all the changes we discussed today are done and ready to be merged

components/fieldgroups/experience-event/experienceevent-campaign-delivery-log.schema.json
```diff
@@ -0,0 +1,174 @@
...
+        "xdm:status": {
+          "title": "Message Status",
+          "type": "string",
+          "description": "Indicates status of the message that is being delivered."
+        },
+        "xdm:reason": {
+          "title": "Reason",
+          "type": "string",
+          "description": "The reason why the message could/would not be delivered.",
+          "meta:enum": {
```
cmathis, line 128: Note this meta:enum is not an actual enum declaration but only a way to define a friendly name to actual values that would exist. If you want this field actually constrained to a set of specific values only then you need a "enum" attribute. As it is here the field will accept any random string.

components/fieldgroups/experience-event/experienceevent-campaign-delivery-log.schema.json
```diff
@@ -0,0 +1,174 @@
...
+        "xdm:serviceId": {
+          "title": "Service Identifier",
+          "type": "string",
+          "description": "Subscribed Service that targeted this delivery."
+        },
+        "xdm:channel": {
+          "title": "Channel",
+          "type": "string",
+          "description": "Channel of the delivery.",
+          "meta:enum": {
```
cmathis, line 68: Add an actual "enum" attribute if you want this field constrained to specific values only.

cmathis:
I think the only thing really missing from this PR is that all the fields must be defined under a top level field named "xdm:directMarketing" or "xdm:marketing". Please take a look at the existing structure of ExperienceEvent and decide where these fields makes the most sense. - https://opensource.adobe.com/xdmVisualization/prod/master/uberschemas.experienceevent-generated.html

components/fieldgroups/experience-event/experienceevent-campaign-delivery-log.schema.json
jishah-adobe: I agree.  However, the reason to not put the constraints is if a new enum value gets added in the source this field group needs to be modified to avoid any failures.  While this is do-able the chances of failures are still high if customer/engineering team forgets to modify the field group.  Enumerations are maintained by the customer and though it is a rare change it can happen.

components/fieldgroups/experience-event/experienceevent-campaign-delivery-log.schema.json
jishah-adobe: Same reason as above

cmathis:
With this PR the XEvent schemas we will now have:
marketing/campaignName and directMarketing/campaignName

Are these "campaignName" two different things? When should someone use one over the other?

components/fieldgroups/experience-event/experienceevent-campaign-delivery-log.schema.json
```diff
@@ -0,0 +1,138 @@
...
+              "title": "External Identifier",
+              "type": "string",
+              "description": "External Identifier that defines the primary identity value for profile and audience."
+            },
+            "xdm:campaignId": {
+              "title": "Campaign Identifier",
+              "type": "string",
+              "description": "Identifier of the marketing campaign to which activity originating this message belongs to."
+            },
+            "xdm:campaignName": {
```
cmathis, line 33: This will introduce a directMarketing/campaignName where we already have a marketing/campaignName (from the marketing field group). When should I use one over the other?

jishah-adobe:
@ rding

piguser (approved):
LGTM

components/fieldgroups/experience-event/experienceevent-campaign-delivery-log.schema.json
jishah-adobe: Hi Craig

components/fieldgroups/experience-event/experienceevent-campaign-delivery-log.example.1.json
```diff
@@ -0,0 +1,16 @@
+{
+  "@id": "0000-0000-0000-0000",
+  "xdm:timestamp": "2021-06-26T15:52:25+00:00",
+  "xdm:directMarketing": {
+    "xdm:campaignId": "100",
+    "xdm:campaignName": "JS - Offer Management",
+    "xdm:broadlogId": "102",
+    "xdm:deliveryId": "103",
+    "xdm:recipientId": "3",
+    "xdm:channel": "Email",
```
mkiki, line 10: Will have to match fixed enumeration value

mkiki (changes requested):

components/fieldgroups/experience-event/experienceevent-campaign-delivery-log.schema.json
```diff
@@ -0,0 +1,138 @@
...
+  "$id": "https://ns.adobe.com/xdm/context/experienceevent-campaign-delivery-log",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Campaign Delivery Logs",
+  "type": "object",
+  "meta:extensible": true,
+  "meta:abstract": true,
+  "meta:intendedToExtend": ["https://ns.adobe.com/xdm/context/experienceevent"],
+  "description": "Campaign Delivery Events such as sent, bounces, failures.",
+  "definitions": {
+    "experienceevent-campaignlog": {
```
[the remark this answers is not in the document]
cmathis: This is not a field name, but a definition name (similar to a function name in code) so the name doesn't really matter. It's how you reference this definition in JSON scheama.

components/fieldgroups/experience-event/experienceevent-campaign-delivery-log.schema.json
cmathis: The marketing field group is defined here:
https://github.com/adobe/xdm/blob/master/components/fieldgroups/experience-event/experienceevent-marketing.schema.json
You can see what it looks like in it's fully resolved form here:
https://opensource.adobe.com/xdmVisualization/prod/master/fieldgroups.experience-event.experienceevent-marketing.html

Note, a field group is used to define fields that can be added to a schema. This PR is introducing a "new" field group (Campaign Delivery Logs) that can be used to add fields to a schema. I want to make sure there is no confusion for a customer when the schema has the potential to contain both a marketing/campaignName (coming from the marketing field group) and this new directMarketing/campaignName (coming from this new field group).

cmathis:
@jishah-adobe  - Is this split into two separate field groups (Campaign Delivery Logs and Campaign Tracking Logs) because we  expect to have two different schema's

components/fieldgroups/experience-event/experienceevent-campaign-delivery-log.schema.json
```diff
@@ -0,0 +1,182 @@
...
+                "Twitter Channel": "Twitter",
+                "iOS Channel": "iOS",
+                "Android Channel": "Android",
+                "LINE Channel": "LINE",
+                "Other": "Other",
+                "Undefined": "Not Defined"
+              },
+              "default": "Undefined"
+            },
+            "xdm:deliveryName": {
```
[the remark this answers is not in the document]
jishah-adobe: Yes it is intended

components/fieldgroups/experience-event/experienceevent-campaign-delivery-log.schema.json
```diff
@@ -0,0 +1,182 @@
...
+              "title": "Delivery Name",
+              "type": "string",
+              "description": "Internal Name that uniquely identifies the delivery."
+            },
+            "xdm:deliveryLabel": {
+              "title": "Delivery Label",
+              "type": "string",
+              "description": "A human-friendly identifier of the delivery which is originating this message."
+            },
+            "xdm:deliveryMode": {
```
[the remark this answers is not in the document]
jishah-adobe: Enumerations are removed across all fields after discussion with Richard

components/fieldgroups/experience-event/experienceevent-campaign-tracking-log.schema.json
```diff
@@ -0,0 +1,135 @@
...
+              "title": "Delivery Mode",
+              "type": "string",
+              "description": "Mode of delivery for the messages sent."
+            },
+            "xdm:trackinglogId": {
+              "title": "Tracking Log Identifier",
+              "type": "string",
+              "description": "Primary Key of the tracking log."
+            },
+            "xdm:url": {
```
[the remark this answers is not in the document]
jishah-adobe: Will resolve this in next iteration if at all required. For now this should suffice for version 1

components/fieldgroups/experience-event/experienceevent-campaign-delivery-log.example.1.json
jishah-adobe: Removed enumeration from the definition and instead mapped this to the solution specific value

components/fieldgroups/experience-event/experienceevent-campaign-delivery-log.schema.json
```diff
@@ -0,0 +1,138 @@
...
+            },
+            "xdm:campaignName": {
+              "title": "Campaign Name",
+              "type": "string",
+              "description": "Internal Name that uniquely identifies the Campaign."
+            },
+            "xdm:campaignLabel": {
+              "title": "Campaign Label",
+              "type": "string",
+              "description": "A human-friendly identifier of the marketing campaign which is originating this message."
```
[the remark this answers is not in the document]
jishah-adobe: Agreed. Modified the description to reflect its purpose

components/fieldgroups/experience-event/experienceevent-campaign-delivery-log.schema.json
```diff
@@ -0,0 +1,138 @@
...
+            },
+            "xdm:broadlogId": {
+              "title": "Delivery Log Identifier",
+              "type": "string",
+              "description": "Identifier of the marketing campaign delivery log this event belongs to."
+            },
+            "xdm:deliveryId": {
+              "title": "Delivery Identifier",
+              "type": "string",
+              "description": "The campaign activity originating this message."
```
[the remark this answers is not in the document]
jishah-adobe: Agreed. Modified the description to reflect its purpose and meaning

components/fieldgroups/experience-event/experienceevent-campaign-delivery-log.schema.json
jishah-adobe: I agree that there will be a confusion around this name of the fields.  It is possible that both may have the same meaning and intent but the one we are defining is for a specific solution.  If there is an overlap should we define this a solution specific namespace then.

components/fieldgroups/experience-event/experienceevent-campaign-delivery-log.schema.json
```diff
@@ -0,0 +1,138 @@
...
+              "title": "Delivery Log Identifier",
+              "type": "string",
+              "description": "Identifier of the marketing campaign delivery log this event belongs to."
+            },
+            "xdm:deliveryId": {
+              "title": "Delivery Identifier",
+              "type": "string",
+              "description": "The delivery originating this message."
+            },
+            "xdm:visitorId": {
```
cmathis, line 53: @jishah-adobe  - What is the different between a visitorId and a recipientId? When would they be different?

components/fieldgroups/experience-event/experienceevent-campaign-delivery-log.schema.json
cmathis: @jishah-adobe - Is it really necessary to have the campaignName here? I'm assuming the actual campaign details would be in another table where you could lookup the current name associated with that campaign.

components/fieldgroups/experience-event/experienceevent-campaign-delivery-log.schema.json
jishah-adobe: Visitor is another form of audience that links visitors to the services that they have subscribed on social platforms (mostly twitter and facebook).  Campaign maintains different event logs for each audience type (recipient, visitor, services, real and batch transactional events, exclusions and so on).  Each of them links to a different look up tables and basically we put the primary identifiers here in the schema

components/fieldgroups/experience-event/experienceevent-campaign-delivery-log.schema.json
jishah-adobe: There is a specific use case where customers may want to add filters on campaign names when doing segmentation queries.  The ID that we store are internal ID and customers don't know about them.  Customers know about 2 fields (Name and Label).  Name is Unique across campaigns (alternate unique key) while labels are not unique but are set by customers and they follow some naming conventions.  So both are necessary to ensure audience segmentation can be done based on the either name or label.  Also from discussion with ACC Engineering Team Campaign Name never gets changed (not edited). Label Modifications is also rare.

prabhum2:
create a new PR with changes  https://github.com/adobe/xdm/pull/1362

Issue #1344, opened 2021-08-04
CORE-64789 Addition of consentPolicies structure to Profile schema


patnaiku:
The schemas that are impacted by this issue is the profile-union schema. 

This will be used as a part of consent initiative implementation in UPS. Once "consent" of the data ingested into the Profile store has been analyzed and evaluated, UPS adds this structure to the merged profile after segmentation and lookup. The downstream services will be able to use the filtered fields from here to take appropriate action. 

**More on the Consent initiative:** https://wiki.corp.adobe.com/display/DMSArchitecture/Consent+Policy+Filter+Logic 

**Consent in UPS:** https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=DMSArchitecture&title=Consent+Enforcement+Implementation+in+UPS 

**More about the "consentPolicies" schema:** https://wiki.corp.adobe.com/display/DMSArchitecture/Consent+Enforcement+Implementation+Details#ConsentEnforcementImplementationDetails-Additionof%22consentPolicies%22toXDM

Example of consent Policies:

`  "consentPolicies": {
    "consentPolicy1": {},
    "consentPolicy3": {
      "email": [
        {"id": "abc@email.com"},
        {"id":"john.doe@domain.com"},
        {"id":"personal.email@email.com"}
      ]
    }
}`

Protocols followed:

1. ConsentPolicies is map type field where the key is "consent policy id/name" and value is identityMap kind of data type.
2. As consent can be identity specific, keep the value empty if a consent policy evaluates to true for all identities.
3. Don't include consent policies in the profile record that evaluate to false (for all identities).

PR #1345, opened 2021-08-04
CORE-64789 Addition of consentPolicies to Profile schema 

patnaiku:
Issue#: https://github.com/adobe/xdm/issues/1344

JIRA: https://jira.corp.adobe.com/browse/CORE-64789

https://wiki.corp.adobe.com/display/DMSArchitecture/Consent+Enforcement+Implementation+Details#ConsentEnforcementImplementationDetails-Additionof%22consentPolicies%22toXDM


cmathis (approved):

extensions/adobe/experience/profile/profile-all.schema.json
```diff
@@ -66,6 +66,25 @@
...
+        },
+        "https://ns.adobe.com/experience/profile/consentPolicies": {
+          "title": "Consent policies map",
+          "type": "object",
+          "meta:xdmType": "map",
+          "additionalProperties": {
+            "title": "consent policy",
+            "type": "array",
+            "items": {
+              "type": "object",
```
RussStringham, line 78: I don't see why this is an object with a single field, "id"? I think it would be more efficient to simply make this an array of strings.

RussStringham:
Should the example file, profile-all.example.1.json, also be updated to include an example of this?

extensions/adobe/experience/profile/profile-all.schema.json
patnaiku: If we don't anticipate any more fields to be added in the future, then I can make it a simple list of strings.

ankushkd:
LGTM

ankushkd (approved):

prabhum2 (approved):

PR #1353, opened 2021-08-20
Add fields to Application plus new Mobile Lifecycle Details field group

kevinlind:
Issue: [PLAT-103821](https://jira.corp.adobe.com/browse/PLAT-103821)

Changes existing Application data type by adding new metric fields for install, upgrade, launch, and close events. Current fields of type Measure will be deprecated in a later PR, after the initial release. The reasoning is to allow stakeholders time to update their systems to use the new fields without impacting existing customers. 

Changes Environment and Device data types by creating dedicated definitions for each field, allowing field groups to directly reference only the fields required in the field group.

Adds Mobile Lifecycle Details field group which directly references the required fields from Application, Environment, and Device.

Deprecates the experimental Mobile Lifecycle Details Test field group, as it is no longer needed.


mjasrotia:
@kevinlind , Thanks for the details in the PR description. This is very helpful. A quick question,  You are creating the new field-group under the extensions/ folder and not under standard components/fieldgroups/. Don't you see its usage in a standard manner ?

components/datatypes/device.schema.json
```diff
@@ -60,7 +72,11 @@
           "title": "Model number",
           "type": "string",
           "description": "The unique model number designation assigned by the manufacturer for this device. Model numbers are not versions, but unique identifiers that identify a particular model configuration. While the model for a particular phone might be 'iPhone 6S' the model number would be 'A1633', or 'A1634' based on configuration at the time of sale."
-        },
+        }
+      }
+    },
+    "screendimensions": {
```
mjasrotia, line 78: If we are splitting properties into their own definitions then I would suggest, we should still keep related properties together. Example, screenorientation and other screen properties could still exist in the same definition.

emdobrin: @mjasrotia the downside of that is that the new field group will need to pull in all the fields even though it only requires `screendimensions`. Should we keep them as per @kevinlind 's definition in this case?

mjasrotia: @emdobrin , what you are saying is the reason we break properties into definitions and I would agree. I still think that just the screen properties won't explode in future, may be few more but should be good. We still could manage those right ? You can leave it the way you have it but even though we create definitions for properties, we should still try to keep the like/related properties under the same definition.

mjasrotia: @emdobrin , Are we good with the suggestion above ? This should not matter much as the field based workflow is coming in AEP-XDM where you would be able to choose individual properties from a field-group while creating schemas through UI. Please let us know if you are making any changes so that we can finally review and close this PR

kevinlind:
@mjasrotia , this schema is intended for use by the AEP Mobile Lifecycle SDK. I view it similar to the AEP Web SDK Experience Event field group than a standard field group. I've updated the schema ID to be `https://ns.adobe.com/experience/mobile-lifecycle-details`. Do you think I should also change the name to something like `aep-mobile-lifecycle-details` or `aep-mobile-lifecycle-sdk-experenceevent`?

components/datatypes/device.schema.json
kevinlind: @mjasrotia , we would prefer to keep the screen dimensions and orientation properties separate if that's okay. I'm not working on any changes there.

prabhum2:
@kevinlind Can you please remove above conflicts and let me know . Thanks

kevinlind:
@prabhum2 I've searched the repo and didn't find any schemas using the following references:
- https://ns.adobe.com/xdm/context/application#/definitions/application
- https://ns.adobe.com/xdm/context/environment#/definitions/environment
- https://ns.adobe.com/xdm/context/device#/definitions/device
Even though the incompatibility-check fails, I feel it's safe to remove or change these definitions. Let me know if this is okay.

prabhum2 (approved):

PR #1355, opened 2021-08-24
introducing tracking mixin for inapp messages, and new experience eve…

iewilliam:
…nt types that will be emitted for inapp message interaction tracking. The 3 new experience event types are:
- inappmessageTracking.display: the inapp message was displayed to the end user.
- inappmessageTracking.dismiss: end user dismissed the inapp message 
- inappmessageTracking.interact: end user interacted with the inapp message, for example by clicking a button in it. 

The new mixin includes a way to identify the specific interaction that the user chose, in the case of inappmessageTracking.interact. A simple string identifier here suffices for the moment.


components/fieldgroups/experience-event/experienceevent-inappmessage-tracking.schema.json
```diff
@@ -0,0 +1,43 @@
...
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Inapp message tracking",
+  "type": "object",
+  "meta:extensible": true,
+  "meta:intendedToExtend": ["https://ns.adobe.com/xdm/context/experienceevent"],
+  "description": "Inapp message interaction tracking details",
+  "definitions": {
+    "experienceevent-inappmessage-tracking": {
+      "properties": {
+        "xdm:inappMessageTracking": {
```
mjasrotia, line 18: @iewilliam , you are adding a root level property "inappMessageTracking" in the experience-event schema. If there could be more ways in future to track a message or message response, we will have to add another root level property with a new prefix for message tracking. Can we just name it as "messageTracking" ?

components/fieldgroups/experience-event/experienceevent-inappmessage-tracking.schema.json
```diff
@@ -0,0 +1,43 @@
...
+          "title": "Inapp message interaction tracking details",
```
mjasrotia, line 19: We generally keep titles as "Title case" with first letter of each word capital
Inapp Message Interaction Tracking Details

components/fieldgroups/experience-event/experienceevent-inappmessage-tracking.schema.json
```diff
@@ -0,0 +1,43 @@
...
+          "type": "object",
+          "description": "Describes interaction with an inapp message.",
+          "properties": {
+            "xdm:action": {
+              "title": "custom action",
```
mjasrotia, line 24: Please make it title case

components/fieldgroups/experience-event/experienceevent-inappmessage-tracking.schema.json
```diff
@@ -0,0 +1,43 @@
...
+              "type": "string",
+              "description": "identifier for custom action that the user triggered while interacting with the inapp message"
```
mjasrotia, line 26: first letter of first word needs to be capital. The description should end with a "." period

components/fieldgroups/experience-event/experienceevent-inappmessage-tracking.schema.json
iewilliam: This already happened unfortunately, the 2 relevant type of mobile messaging we are tracking is push notification and inapp message. Push notification happened first and it already has its own top level field. Since we can't rename mixins, we can't simply reuse that mixin.

PR #1356, opened 2021-08-24
[CJM-14229] Refactoring offers as a datatype to enable among CJM mixins

usankara:
- Introduced data type for offers, so that it can be re-used across mixins.
- Used it as a part of message-interaction.schema.json & message-delivery-feedback.schema.json


usankara:
@sauagarw, please review. Note that offers data type is separated out here, so that it can be re-used.

sauagarw:
@usankara Are we not adding the offers object to feedback mixin as a part of this PR?

extensions/adobe/experience/customerJourneyManagement/message-interaction.example.1.json
```diff
@@ -8,9 +8,9 @@
   ],
   "https://ns.adobe.com/experience/customerJourneyManagement/messageInteraction/label": "adobe.com",
   "https://ns.adobe.com/experience/customerJourneyManagement/messageInteraction/offers": {
-    "https://ns.adobe.com/experience/customerJourneyManagement/messageInteraction/offers/offerID": "1234",
-    "https://ns.adobe.com/experience/customerJourneyManagement/messageInteraction/offers/placementID": "12345",
-    "https://ns.adobe.com/experience/customerJourneyManagement/messageInteraction/offers/propositionID": "9898",
-    "https://ns.adobe.com/experience/customerJourneyManagement/messageInteraction/offers/offerCount": 5
+    "https://ns.adobe.com/experience/customerJourneyManagement/offers/offerID": "1234",
```
sawhneyh, line 11: @usankara:  As namespace of new Offer data type and place where we are adding this new type in messageInteraction schema doesn't have a common prefix, the new fields added in offer data type, won't be added as direct child of _**_experience.customerJourneyManagement.messageInteraction.messageInteraction.offers.offerID**_ but would be added with complete namespace: **__experience.customerJourneyManagement.messageInteraction.offers.experience.customerJourneyManagement.offers.offerID_**

I don't think we should be adding a redundant prefixes while adding the offers fields in messageInteraction schema.

cc: @mjasrotia , @sauagarw

guayang-adobe (approved):

prabhum2 (approved):

PR #1363, opened 2021-09-13
CJM-15219: Add subscription schema in xdm.

sausharm:
Please link to the issue #…


sausharm:
@prabhum2 @mjasrotia @guayang-adobe Please review the PR

guayang-adobe (approved):

prabhum2 (approved):

mjasrotia (approved):

extensions/adobe/experience/customerJourneyManagement/subscription.schema.json
```diff
@@ -0,0 +1,61 @@
...
+  "$id": "https://ns.adobe.com/experience/customerJourneyManagement/subscription",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Adobe ExperienceEvent - Subscription details",
+  "description": "Details about subscription to subscription list",
+  "type": "object",
+  "meta:extensible": true,
+  "definitions": {
+    "subscription": {
+      "properties": {
+        "xdm:subscriptionListId ": {
```
sawhneyh, line 17: @prabhum2 / @mjasrotia : Can you guide when we should be using the xdm namespace or solution specific namespaces while defining the fields in mixins.

cmathis: You are trying to declare a new Datatype here. If this datatype is specifically for "adobe.com/experience/customerJourneyManagement" use cases (which the $id implies it is) then all these fields should be assigned to that namespace. 

If this was intended to be a generic standard datatype under the "adobe.com/xdm" namespace, then the fields would fall under the "xdm" namespace. 

We want to avoid creating datatypes if they are never intended to be re-used for multiple fields of that type. I see where we are using this datatype in your mixin above for one field, but if its that field only we should simply nest these fields under it.

extensions/adobe/experience/customerJourneyManagement/message-interaction.schema.json
```diff
@@ -63,9 +70,14 @@
...
         },
         "https://ns.adobe.com/experience/customerJourneyManagement/messageInteraction/offers": {
-          "title": "Offer details",
+          "title": "Offer Details",
           "$ref": "https://ns.adobe.com/experience/customerJourneyManagement/offers",
           "description": "Describes offer(s) involved in a message interaction."
+        },
+        "https://ns.adobe.com/experience/customerJourneyManagement/messageInteraction/subscription": {
+          "title": "Subscription Details",
+          "$ref": "https://ns.adobe.com/experience/customerJourneyManagement/subscription",
```
cmathis, line 79: If there are no intentions of re-using that new subscription datatype, then simply nest the fields under this parent "subscription" field. Only create a datatype if you have multiple fields that need to be of that type.

sawhneyh (approved):

sausharm:
@cmathis @prabhum2 @mjasrotia @guayang-adobe I have incorporated the review comments. Please re-review the PR. Also, if changes are ok then please merge it.

prabhum2 (approved):

prabhum2 (approved):

sausharm:
@cmathis Could you please re-review the PR. Also, please merge it if changes are ok.

sausharm:
@prabhum2 @cmathis @mjasrotia 
I saw the changes deployed on stage. But, subscription fields have path something like **_experience.customerJourneyManagement.messageInteraction.subscription._experience.customerJourneyManagement.messageInteraction.subscription.subscriptionFieldName**.
Attching the schema image

But, we wanted that subscription fields should have path something like **_experience.customerJourneyManagement.messageInteraction.subscription.subscriptionFieldName**.

As per our understanding, common path prefix between a field and its parent field, is removed from that particular field path. But, it seems that it is not correct.

So, what changes need to be done for achieving that subscription fields should have path something like **_experience.customerJourneyManagement.messageInteraction.subscription.subscriptionFieldName**.

cmathis:
It shouldn't be repeated, but let me check and will update the schema as needed to make it not repeat.

Issue #1365, opened 2021-09-14
Extend webInteraction datatype to include region


liljenback:
## What are the schemas that are affected by the issue
`components/datatypes.webinfo.schema.json`

## What are examples of products that are impacted by the issue
- Adobe Experience Platform
- Adobe Analytics
- Adobe ActivityMap

PR #1366, opened 2021-09-14
Adding link region to webInteraction to prepare for integration with …

liljenback:
…ActivityMap

Issue: https://github.com/adobe/xdm/issues/1365


components/datatypes/webinfo.schema.json
```diff
@@ -112,6 +112,11 @@
               "meta:titleId": "webinfo##xdm:name##title##72971",
               "meta:descriptionId": "webinfo##xdm:name##description##40881"
             },
+            "xdm:region": {
```
prabhum2, line 115: What is the purpose of this property ? Does this indicate location in the website where user clicked ? xdm:region is used in address to denote geo region , Please rename this to avoid conflict with xdm:region in address

liljenback: As the description states it is the name of the region where the web interaction occurred. It is the region name in closest DOM proximity to the hyperlink that was clicked. I'm not sure it needs to be more descriptive given that it is encapsulated by webinteraction.

prabhum2: sounds good . Can you please rename this property to avoid conflicts and confusion with xdm:region existing in address

prabhum2 (approved):

PR #1376, opened 2021-09-28
Add boolean to track push notification interaction triggered launches…

iewilliam:
… since we now need to move away from using application.isLaunch and application.launches.VALUE for the purpose of push reporting. The mobile SDK's application lifecycle tracking component will now issue its own experience events (with the application mixin attached) for all application launches so we need to do this to avoid double counting.


components/fieldgroups/experience-event/experienceevent-pushtracking.schema.json
```diff
@@ -56,6 +56,11 @@
               "description": "Push provider, e.g. apns or fcm",
               "meta:titleId": "experienceevent-pushtracking##xdm:pushProvider##title##90271",
               "meta:descriptionId": "experienceevent-pushtracking##xdm:pushProvider##description##23691"
+            },
+            "xdm:isLaunch": {
```
mjasrotia, line 60: Since "application" data type already has an "isLaunch" property and if we add another one here, it might add some confusion for the users regarding which one to use when they search for "isLaunch" properties. I would suggest if this is specific to push notifications, can we rename this to something like "isPushTriggeredLaunch" ?

prabhum2 (approved):

mjasrotia (approved):

PR #1395, opened 2021-10-28
include idType (IDFA/GAID) for the adID consent

RussStringham:
PDCL-213


RussStringham:
The build failure is '../components/datatypes/consent/consent-preferences.schema.json breaking changes found!!! Data type of property "definitions/consent-preferences/properties/xdm:consents/properties/xdm:adID/$ref" can not be changed.'

The new $ref is identical to the old $ref, except that it adds an additional field, so this is not a breaking change.

prabhum2 (approved):

mjasrotia (approved):

PR #1397, opened 2021-11-02
Adding landing page context in message interaction mixin

aarjain:
Please link to the issue #…


aarjain:
@prabhum2 , @cmathis : Please review this PR.

prabhum2:
> @prabhum2 : The build is failing as I am trying to remove an existing attribute from a mixin. However, since I had kept the status of that mixin as "experimental" and had not enabled it for any IMS ORGs, I believe this mixin would not be available for anyone to use, making the removal of this field harmless. Is that correct understanding?

Yes , As long as no one is using or uploaded data set for this schema you are good.

aarjain:
@prabhum2 : Since the status was experimental and no orgs were enabled for this, we are sure that this would not have been used anywhere.

sausharm (approved):

aarjain:
@prabhum2 , @cmathis : Please review. The breaking changes in file landingpage.schema.json are safe as no one is using these mixins (as explained in my comments above).

cmathis:
It appears you are trying to convert your solution specific fields to score XDM namespaced fields. This is fine, but this field group should be moved into the official components/fieldgroups/experienceevent folder. In addition, you need to remember we are defining fields which are to be part of the standard XEvent entity/object - they must be added in a place which makes sense in the context of all the other fields which can exist. You can see in the link below what this XEvent schema looks like and decide where these landing page fields make the most sense - possibly under the top level "web" field. https://opensource.adobe.com/xdmVisualization/prod/master/uberschemas.experienceevent-generated.html

aarjain:
@cmathis , @prabhum2 : We have decided to nest everything in our solution namespace. Made the changes in this PR, please review.

cmathis (approved):

mjasrotia (approved):

PR #1421, opened 2022-01-13
Added support of excludeProperty

meichen01:
PLAT-113406 - Add support of "Xdm:excludeProperty" in descriptor (DescriptorLabel)

### DIFF
--- a/schemas/descriptors/label/descriptorLabel.schema.json
+++ b/schemas/descriptors/label/descriptorLabel.schema.json
@@ -36,7 +36,7 @@
               }
             }
           ],
-          "description": "When present, the property of the source schema to which this descriptor excludes. This value is a JSON Pointer, applied to an instance of an object described by `xdm:sourceSchema`.",
+          "description": "When present, the property of the source schema to which this descriptor excludes. This value is a JSON Pointer, applied to an instance of an object described by `xdm:sourceSchema`."
         },
         "xdm:labels": {
           "title": "Labels",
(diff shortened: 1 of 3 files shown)


schemas/descriptors/label/descriptorLabel.schema.json
```diff
@@ -5,7 +5,7 @@
     "you may not use this file except in compliance with the License. You may obtain a copy",
     "of the License at https://creativecommons.org/licenses/by/4.0/"
   ],
-  "$id": "https://ns.adobe.com/xdm/common/descriptors/descriptorLabel",
```
veramach-adobe, line 8: @prabhum2 Please confirm if this change is good

veramach-adobe (approved):
LGTM

schemas/descriptors/label/descriptorLabel.schema.json
```diff
@@ -23,6 +23,21 @@
...
+            },
+            {
+              "type": "array",
+              "items": {
+                "type": "string"
+              }
+            }
+          ],
+          "description": "When present, the property of the source schema to which this descriptor excludes. This value is a JSON Pointer, applied to an instance of an object described by `xdm:sourceSchema`."
+        },
```
mjasrotia, line 40: @veramach-adobe , a quick question. Why this can't be just an array similar to xdm:sourceProperty which can take 1 or more elements. Also, if we are talking about the exclusion of nested properties, there is no way to enforce that in the JSON schema definition here that "excludeProperty" should have the parent mentioned in the "sourceProperty". It will be the responsibility of the consuming service to validate that. Right ?

mjasrotia (approved):

PR #1428, opened 2022-01-26
pii disclaimer

prabhum2:
https://jira.corp.adobe.com/browse/PLAT-112037


hanessian (changes requested):
- There is no space between the end of each sentence and the disclaimer. Can you please add a space?
- Are we able to make the link a hyperlink, or the full link will show? If it will be the full link, let's put the wording ahead of it: "click here to learn more: LINK" instead of embedding the link within the text. Would be best if we could make the "here" the link, but if that's not possible let's just change the wording to the above.

prabhum2:
> * There is no space between the end of each sentence and the disclaimer. Can you please add a space?
> * Are we able to make the link a hyperlink, or the full link will show? If it will be the full link, let's put the wording ahead of it: "click here to learn more: LINK" instead of embedding the link within the text. Would be best if we could make the "here" the link, but if that's not possible let's just change the wording to the above.

PR #1441, opened 2022-02-03
Add campaignKey and remove campaignID

amithana:
Field Group changes to Change Campaign Cadence.

Please link to the issue #…


components/fieldgroups/experience-event/events/change-campaign-cadence.schema.json
```diff
@@ -29,10 +29,10 @@
               "title": "Change Campaign Cadence",
               "type": "object",
               "properties": {
-                "xdm:campaignID": {
```
mjasrotia, line 32: We do not allow breaking changes like removing a property from a schema. If there is a need to replace this, please deprecate the existing field and add another one.

amithana: Got it. Will do that. Thanks.

mjasrotia (approved):

prabhum2 (approved):

PR #1462, opened 2022-02-24
Work Object

cmathis:
- Introduces a new "Work Object" class and a sample Task Field Group that would be used if it the schema was intended for that type of object.
- Updates the event change schema to have a generic parentID field rather then a static object type parent fields. **Breaking change is expected and planned for.**


extensions/adobe/experience/workfront/task.schema.json
```diff
@@ -0,0 +1,38 @@
...
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/experience/workfront/task",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Work Task Atrributes",
+  "type": "object",
+  "meta:extensible": true,
+  "meta:abstract": true,
+  "meta:intendedToExtend": ["https://ns.adobe.com/experience/workfront/object"],
```
cmathis, line 14: Implies this is a Field Group for the Work Object class.

extensions/adobe/experience/workfront/workobject.schema.json
```diff
@@ -0,0 +1,57 @@
...
+          "title": "Parent Object ID",
+          "description": "The @id of a parent object if it exists",
+          "type": "string"
+        },
+        "workfront:objectType": {
+          "title": "Object Type",
+          "description": "Basic string value representing the high level type of object",
+          "type": "string"
+        },
+        "workfront:lastUpdatedByID": {
```
mjasrotia, line 34: Should we consider using the "External Source System Audit Details" field group here?  It has the following two properties:

1. lastUpdatedBy
2. lastUpdatedDate

mjasrotia (approved):

prabhum2 (approved):

PR #1490, opened 2022-03-16
Plat 115210 XDM support for removing suggested values not needed from standard fields

bisin-adobe:
Please link to the issue #…


prabhum2 (approved):

schemas/descriptors/display/alternateDisplayInfo.schema.json
```diff
@@ -106,6 +106,9 @@
         },
         {
           "required": ["meta:enum"]
+        },
+        {
+          "required": ["xdm:excludeMetaEnum"]
```
mjasrotia, line 111: @bisin-adobe , Why do we need to make this required when the same descriptor is used for other purposes as well ?

bisin-adobe: @mjasrotia 
That is an anyOf attribute. It means one of the required field should be available in the payload.
If you want to exclude meta:enum, "xdm:excludeMetaEnum" is mandatory and this json will do the validation

mjasrotia (approved):

Issue #1524, opened 2022-04-28
Make XDM changes for all Alpha temporary fields


lidiaist:
## What are the schemas that are affected by the issue

* media.schema.json

## What are examples of products that are impacted by the issue

*  Adobe Media Analytics
Internal JIRA issue: https://jira.corp.adobe.com/browse/VA-4673

High-level changes:
* support for playhead parameter https://experienceleague.adobe.com/docs/media-analytics/using/media-collection-api/mc-api-ref/mc-api-sessions-req.html?lang=en

PR #1538, opened 2022-05-19
[VA-4675] Modify playhead field to be optional

lidiaist:
Issue https://github.com/adobe/xdm/issues/1524


lidiaist:
@mjasrotia @prabhum2 Please review.  I modified this playhead field to be optional because of our implementation. As an estimate, this change will get to stage on 23 May and prod on 30 May? Thank you!

mjasrotia (approved):

prabhum2 (approved):

Issue #1544, opened 2022-05-25
Add Field Groups and Classes in support of Live Event (Sports and Entertainment) Industry


derek-selby:
## What are the schemas that are affected by the issue
In support of the live event, aka sports and entertainment, industry we have created three new field groups and classes to describe the common data elements of this industry.

The plan is eventually to use these AEP structures to support templatized AJO journeys for marketers.

1. Event Venue Class
2. Live Event Ticket Purchase Field Group
3. Live Event Schedule Class

## What are examples of products that are impacted by the issue
AJO - though these are generic to the sports and entertainment industry, they were built to support AJO journey execution.

PR #1545, opened 2022-05-25
Add Field Groups and Classes in support of Live Event (Sports and Entertainment) Industry

derek-selby:
Please link to the issue #…
[https://github.com/adobe/xdm/issues/1544](url)


derek-selby:
Need to sign the CLA and fix some linting issues.

PR #1546, opened 2022-05-25
Add Field Groups and Classes in support of Live Event (Sports and Entertainment) Industry

derek-selby:
Please link to the issue #…
[https://github.com/adobe/xdm/issues/1544](url)


derek-selby:
Moving changes to a new branch. Will re-submit.

PR #1547, opened 2022-05-27
healthcare plan schemas 

prabhum2:
- Created new healthcare plan class
- Created new healthcare plan details filed group extending healthcare plan class


components/classes/healthcare-plan.schema.json
```diff
@@ -0,0 +1,61 @@
+{
+  "meta:license": [
+    "Copyright 2022 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/classes/healthcare-plan",
```
mjasrotia, line 8: Let's call this just "plan" as it's a general concept and be used for other industries too.

components/classes/healthcare-plan.schema.json
```diff
@@ -0,0 +1,61 @@
...
+          "Title": "Plan ID",
+          "type": "string",
+          "description": "Unique identifier of the insurance plan."
+        },
+        "xdm:planName": {
+          "Title": "Plan Name",
+          "type": "string",
+          "description": "Name of the insurance plan."
+        },
+        "xdm:planType": {
```
mjasrotia, line 32: If its only for primary, secondary kind of healthcare plans, let's move this to the healthcare-plan-details FG. We can also leave it here with no industry specific enums and add "isPrimary", isSecondary" and "isTeritiary" properties in the Field group for healthcare plan details.

components/fieldgroups/healthcare-plan/healthcare-plan-details.schema.json
```diff
@@ -0,0 +1,92 @@
...
+  "description": "Details of a Health Insurance plan provided by an organization.",
+  "definitions": {
+    "plan-details": {
+      "properties": {
+        "xdm:payerID": {
+          "Title": "Payer ID",
+          "type": "string",
+          "description": "Unique identifier of the payer."
+        },
+        "xdm:contractID": {
```
mjasrotia, line 27: Contract looks more like an agreement between Plan and a member so, it should be either removed from here or moved to the plans array in member if required.

components/fieldgroups/healthcare-plan/healthcare-plan-details.schema.json
```diff
@@ -0,0 +1,92 @@
...
+          "format": "date-time",
+          "description": "Coverage end date."
+        },
+        "xdm:lastVerification": {
+          "Title": "Last Verification Date",
+          "type": "string",
+          "format": "date-time",
+          "description": "The date on which this plan was last verified."
+        },
+        "xdm:cost": {
```
mjasrotia, line 77: cost is a relative term which changes across insurance companies and the insured. Not sure if we need to add cost in plan. The same plan could be sold for a different cost depending upon the insured pool. Please check with SMEs if we need to move it somewhere else.

components/fieldgroups/healthcare-plan/healthcare-plan-details.schema.json
```diff
@@ -0,0 +1,92 @@
...
+          "enum": ["Medical", "Dental", "Vision", "Accident"],
+          "meta:enum": {
+            "Medical": "Medical",
+            "Dental": "Dental",
+            "Vision": "Vision",
+            "Accident": "Accident"
+          },
+          "description": "Member plan coverage type."
+        },
+        "xdm:coverageStartDate": {
```
mjasrotia, line 59: Coverage Start and End Dates are Member-Plan specific These dates are already in the Member==>plans[] array so, we can remove from here.

components/fieldgroups/healthcare-plan/healthcare-plan-details.schema.json
prabhum2: This is added based on SME's feedback

mjasrotia:
Moved this to
https://github.com/adobe/xdm/pull/1563

PR #1548, opened 2022-05-27
healthcare  member schema Changes

prabhum2:
- Updated healthcare member field group as per the feedback.


mjasrotia:
Moved this to
https://github.com/adobe/xdm/pull/1563

PR #1557, opened 2022-06-06
Adding subscription, consent, custom email and additional data information in landing page object

aarjain:
Please link to the issue #…


aarjain:
@prabhum2 , @mjasrotia : Please review this PR

extensions/adobe/experience/customerJourneyManagement/message-interaction.schema.json
```diff
@@ -115,6 +115,112 @@
...
+                      },
+                      "https://ns.adobe.com/experience/customerJourneyManagement/messageInteraction/type": {
+                        "title": "Subscription Type",
+                        "type": "string",
+                        "description": "The type of subscription."
+                      }
+                    }
+                  }
+                },
+                "https://ns.adobe.com/experience/customerJourneyManagement/messageInteraction/consents": {
```
mjasrotia, line 153: Did you consider using the standard XDM "Consent and preferences Details" field group ? It will be great to use s a standard way to use consent signals consistently across the platform.

aarjain:
Thanks @mjasrotia for merging. However, we have some discussions going on, and we may need to make some breaking changes in this schema. What would you suggest in that case? Should we revert this from STAGE or can we make breaking changes, since this schema is only being used within CJM solution?

PR #1599, opened 2022-08-09
added segmentation support to xdm bus acct

prabhum2:
Please link to the issue #…


simon-marketo (approved):
Looks good to me.

mjasrotia:
@prabhum2 , Can we move this to the shared folder please as now its extending more than 1 class ? Thanks !

mjasrotia (approved):

PR #1602, opened 2022-08-12
Fix weather schema (uvIndex and sunsetTime)

mflorin:
Issue: https://github.com/adobe/xdm/issues/1603

Fixes errors related to `uvIndex` and `sunsetTime`: we're sending an object and a long value, but the schema expected
an enum and a string.


components/fieldgroups/shared/forecasted-weather.schema.json
```diff
@@ -324,6 +240,18 @@
           "description": "Distance in miles."
         }
       }
+    },
+    "uvIndex": {
+      "properties": {
+        "xdm:_name": {
```
prabhum2, line 246: what is the significance of "_" in "xdm:_name" ? Can this be simple name without "_" ?

mflorin: @prabhum2 - we need this in the schema to align with some data we're producing and unfortunately is much more expensive for us to adjust the data we're producing

prabhum2 (approved):

mjasrotia (approved):

Issue #1550, opened 2022-05-31
Simplification of media schema


lidiaist:
## What are the schemas that are affected by the issue

* advertisingdetails.schema.json
* advertisingpoddetails.schema.json
* chapterdetails.schema.json
* errordetails.schema.json
* playereventdetails.schema.json
* playerstatedata.schema.json
* qoedatadetails.schema.json
* sessiondetails.schema.json
* experienceevent-mediacollection.schema.json
* experienceevent-mediareporting.schema.json

## What are examples of products that are impacted by the issue

*  Adobe Media Analytics
Internal JIRA issue: https://jira.corp.adobe.com/browse/VA-4808

High-level changes:
* create datatypes to be used in 2 fieldgroups: mediacollection and mediareporting
* deprecate the media.schema.json and use the new schemas for a better experience for our clients

PR #1605, opened 2022-08-17
[VA-5145] Create mediaDetails datatype and downloaded fieldgroup

lidiaist:
This PR is ready for review. We created a new downloaded fieldgroup which uses an array of mediaDownloadedEvent datatypes, this fieldgroup will be used to send the events of a full session at once.


components/datatypes/mediacollectiondetails.schema.json
```diff
@@ -0,0 +1,93 @@
...
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Media collection details information",
+  "type": "object",
+  "meta:status": "experimental",
+  "meta:extensible": true,
+  "description": "Media collection details information.",
+  "definitions": {
+    "mediaCollectionDetails": {
+      "properties": {
+        "xdm:timeSeries": {
```
mjasrotia, line 18: What is the purpose of adding this here ? timestamp and eventType are pre-included in a schema when you use any experience event class based schemas.

lidiaist: We created the mediaCollectionDetails datatype to be used inside an array for the downloaded fieldgroup. Each of the mediaCollectionDetails in this array represents an individual event and needs to have specified the eventType and timestamp of the individual event (e.g. https://github.com/adobe/xdm/blob/c146e10808d2a1141c53f994867cd1151d8fb90a/components/fieldgroups/experience-event/experienceevent-media-collection-downloaded.example.1.json). The downloaded content represents a full media collection session with all events starting from sessionStart, ping, play, etc. to sessionComplete.

components/datatypes/mediaevent.schema.json
```diff
@@ -0,0 +1,35 @@
...
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/datatypes/mediaEvent",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Media event information",
+  "type": "object",
+  "meta:status": "experimental",
+  "meta:extensible": true,
+  "meta:extends": ["https://ns.adobe.com/xdm/data/time-series"],
```
mjasrotia, line 14: Adding this behavior will turn this into a class. I guess you just want to create a data type here right ? Please clarify.

components/fieldgroups/experience-event/experienceevent-media-collection-downloaded.schema.json
```diff
@@ -0,0 +1,39 @@
...
+  "meta:abstract": true,
+  "meta:tags": {
+    "industry": ["all"]
+  },
+  "meta:intendedToExtend": ["https://ns.adobe.com/xdm/context/experienceevent"],
+  "description": "Track interactions with media collection downloaded.",
+  "definitions": {
+    "experienceevent-mediaCollectionDownloaded": {
+      "properties": {
+        "xdm:events": {
```
mjasrotia, line 22: "event" as such has a very genetic meaning on the platform. It can easily be confused by users for any event. I would advise to prefix this with some context e.g : "xdm:mediaEvents" ? Also, I have already left a comment in the mediaEvent data type. It is referred here in the array. You will have to fix that as that is currently defined as a class and not an array.

components/datatypes/mediaevent.schema.json
lidiaist: yes, we need the mediaEvent datatype to have the format of e.g.:
```
{
      "eventType": "media.adBreakStart",
      "timestamp": "2022-04-13T15:51:05+03:00",
      "mediaCollection": {
        "playhead": 548,
        "sessionDetails": {
          "ID": "someID"
        },
        "advertisingPodDetails": {
          "friendlyName": "mid-roll",
          "index": 1,
          "second": 548
        }
}
```
so the timestamp and eventType from timeseries would be just as simple fields, not inside another datatype, if this is not possibile using `extends` one option would also be to add these fields manually.

components/fieldgroups/experience-event/experienceevent-media-collection-downloaded.schema.json
lidiaist: Good point. I added a new commit with this change of renaming `events` to `mediaEvents`. Thank you!

components/fieldgroups/experience-event/experienceevent-media-collection-downloaded.schema.json
```diff
@@ -0,0 +1,39 @@
+{
+  "meta:license": [
+    "Copyright 2022 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/mixins/experienceevent-media-collection-downloaded",
```
mjasrotia, line 8: Can you merge the below 3 field groups to 1 ?
media-collection
media-collection-downloaded
media-reporting

You can just move "xdm:mediaReporting" and "xdm:mediaEvents" objects to the media-collection field group.

lidiaist: Done. I created a fieldgroup of mediaAnalytics which contains mediaCollection, mediaReporting and mediaDownloadedEvents.

PR #1616, opened 2022-08-30
xdm changes to facilitate location events

iewilliam:
xdm changes necessary to ease sending location events into AEP:
1. Introducing a new experience event's location details field group
2. Adding location.entry/exit experience event types
3. Augment poi details datatype to be able to include a list of metadata kv pairs associated with the point of interest. 

use case:
AEP SDK will send location.entry/exit experience events to a dataset with the location details field group. 
Downstream services such as AJO reads the dataset and trigger various experiences based on poi details.


mjasrotia:
@iewilliam , we already have a field group which introduces the"placeContext" as a top level property. 
https://github.com/adobe/xdm/blob/master/components/fieldgroups/experience-event/experienceevent-environment-details.schema.json

Do we have plans to deprecate the existing property ? You can map your dataset to the existing property too. The existing placeContext refers to the same poi-detail datatype which you can still enhance by adding more properties. Adding the new event types is not a problem.

mjasrotia:
@iewilliam , please remove the location details field group as discussed.

components/datatypes/poi-detail.schema.json
```diff
@@ -144,6 +144,11 @@
           },
           "meta:titleId": "poi-detail##xdm:geoInteractionDetails##title##47521",
           "meta:descriptionId": "poi-detail##xdm:geoInteractionDetails##description##39951"
+        },
+        "xdm:metadata": {
+          "title": "POI metadata",
+          "description": "metadata assigned to the POI by the administrator of the POI definitions.",
+          "$ref": "https://ns.adobe.com/experience/analytics/keyedlist"
```
mjasrotia, line 151: @iewilliam , I noticed here that you are using a data type from Analytics here which is not recommended as we do not manage the functional changes of schema for internal Adobe solutions. It's very common to have standard XDM components used in the solution Field groups but not the other way around. Please clone the analytics data type under components/datatypes folder and then use that in your $ref.

iewilliam: done @mjasrotia

components/datatypes/keyedlist.schema.json
```diff
@@ -0,0 +1,37 @@
+{
+  "meta:license": [
+    "Copyright 2018 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/context/keyedlist",
```
mjasrotia, line 8: Please rename this to "https://ns.adobe.com/xdm/datatypes/keyedlist".

components/datatypes/keyvalue.schema.json
```diff
@@ -0,0 +1,41 @@
+{
+  "meta:license": [
+    "Copyright 2022 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/context/keyvalue",
```
mjasrotia, line 8: Please rename this to "https://ns.adobe.com/xdm/datatypes/keyvalue".

components/datatypes/poi-detail.schema.json
```diff
@@ -144,6 +144,11 @@
           },
           "meta:titleId": "poi-detail##xdm:geoInteractionDetails##title##47521",
           "meta:descriptionId": "poi-detail##xdm:geoInteractionDetails##description##39951"
+        },
+        "xdm:metadata": {
+          "title": "POI metadata",
+          "description": "metadata assigned to the POI by the administrator of the POI definitions.",
+          "$ref": "https://ns.adobe.com/xdm/context/keyedlist"
```
mjasrotia, line 151: Please rename this to "https://ns.adobe.com/xdm/datatypes/keyedlist".

iewilliam: @mjasrotia done.

mjasrotia: @iewilliam , I do not see the changed $id yet anywhere

iewilliam: oops, resubmitted.

mjasrotia (approved):

Issue #1519, opened 2022-04-20
Extend time-series eventType meta:enum for AJO web-channel inbound use cases


surbhi114:
## What are the schemas that are affected by the issue
components/behaviors/time-series.schema.json

## What are examples of products that are impacted by the issue
Adobe Experience Platform
Adobe Analytics
Adobe Target

PR #1619, opened 2022-09-01
CJM-33819 Rename eventtype Deliver to Send and add Dismiss

surbhi114:
Please link to the issue:
Jira:https://jira.corp.adobe.com/browse/CJM-33819
Github Issue: https://github.com/adobe/xdm/issues/1519

Summary:
Rename eventtype decisioning.propositionDeliver to decisioning.propositionSend and adds decisioning.propositionDismiss


donwalling (approved):
👍

shagreel (approved):

mjasrotia:
Was this a recently introduced event type ? I am wondering if there are any existing segments created based on this event type which will no longer evaluate if we change this. Could you please confirm before we can make this change ?

donwalling:
hi @mjasrotia yes, Deliver was recently introduced and is now being changed to Send. It is only being used in Stage environment right now, and will be changed once the xdm change is available. This feature is still pre-release, and the name change is needed so that we can progress to beta.

mjasrotia (approved):

PR #1621, opened 2022-09-01
changing the datatype namespace to xdm

shagreel:
When viewing the previous xdm change in the stage env I discovered that a datatype needs to use the `xdm` namespace otherwise the namespace is prepended to the datatype fields. I have made this change. If this merge could be added to last week's release that would be best so we do not introduce an invalid datatype in the prod env.


donwalling (approved):
+1

shagreel:
This is a known breaking change. No one is using the schema that this is changing.

shagreel:
This PR is ready for deployment.

mjasrotia:
Approving this after discussion with @shagreel that these properties are not yet used anywhere.

PR #1627, opened 2022-09-12
Removing subscriber address, as it is no longer needed

aarjain:
Please link to the issue #…


mjasrotia:
@aarjain , you are introducing a breaking change here. Removing this field from the schema will break existing datasets which have data ingested against this field (https://ns.adobe.com/experience/customerJourneyManagement/messageInteraction/subscriberContactAddress)

Please confirm if that's acceptable.

mjasrotia:
@aarjain , Is this PR still required ? I asked a question above and I haven't heard back yet.

mjasrotia:
Please re -open if required.

PR #1631, opened 2022-09-12
added propositionAction datatype

shagreel:
Please link to the issue #…

### DIFF
--- a/extensions/adobe/experience/decisioning/proposition-action.schema.json
+++ b/extensions/adobe/experience/decisioning/proposition-action.schema.json
@@ -1,0 +1,38 @@
+{
+  "meta:license": [
+    "Copyright 2022 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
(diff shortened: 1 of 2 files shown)


extensions/adobe/experience/decisioning/proposition-action.schema.json
```diff
@@ -0,0 +1,33 @@
...
+  "type": "object",
+  "meta:status": "stable",
+  "meta:extensible": true,
+  "meta:abstract": true,
+  "definitions": {
+    "proposition-action": {
+      "type": "object",
+      "properties": {
+        "xdm:id": {
+          "title": "id",
```
shagreel, line 21: @sbenedicadb Do mobile elements have both an id and label that should be tracked here?

sbenedicadb: not today - the UI allows the customer to specify what "action" to report when a user interacts with the message's (e.g.) button.  with that in mind, i'm not sure which one of `id` or `label` would be the correct one to use from the mobile perspective

pfransenadb: I think both can make sense in the long term, in the short term I'd populate id with the "action ID" -- in the future we may want to use label to contain the string label of the action itself (e.g. the actual value displayed on the button or other CTA).

mjasrotia (approved):

donwalling (commented):
👍

PR #1638, opened 2022-09-29
update workfront fieldgroups

rammatta:
move common field from fieldgroups into class object. 

We know that we are removing 2 fieldgroups and it is a breaking change. It is intentional and expected.


mjasrotia (approved):

mjasrotia:
Please re-open if required.

PR #1641, opened 2022-10-05
Updated frequency capping constraints schema to support repeat and custom events

jshaik-adobe:
https://jira.corp.adobe.com/browse/CJM-29979

1. Updated frequency capping constraints schema to support repeat & custom events
2. Deprecated unused attributes from frequency capping constraints schema

data type change for `xdm:repeat` is intentional .. the feature for frequency capping repeat is not live and we wanted to change the data type for `xdm:repeat` from `boolean` to `object`.

- [personalized-content-offer.schema.json](https://github.com/adobe/xdm/blob/master/extensions/adobe/experience/offer-management/personalized-content-offer.schema.json)
- [frequency-capping-constraints.schema.json](https://github.com/jshaik-adobe/xdm/blob/master/extensions/adobe/experience/decisioning/frequency-capping-constraints.schema.json)

- No impact on the existing products. Frequency capping constraints define additional rules based on prior offer proposition, either across all audiences or for a particular profile identity.


extensions/adobe/experience/decisioning/frequency-capping-constraints.schema.json
```diff
@@ -51,49 +51,57 @@
                   "placement": "To comply with offer placement."
                 }
               },
-              "xdm:placement": {
```
prabhum2, line 54: @jshaik-adobe Can you please confirm that there is no data set uploaded in platform associated to this field group referenced in any schema ?

jshaik-adobe: @prabhum2 , we don't have any data for deleted fields

mjasrotia: @jshaik-adobe , to be on the safer side, can we just deprecate these properties ? We generally do not delete properties from a schema which is marked as "stable".

jshaik-adobe: @mjasrotia deprecated unused properties

mjasrotia:
@jshaik-adobe , would you please  update your master branch and then re submit the PR ? It has some missing objects from the current master. Please refer the build failure. Thanks !

jshaik-adobe:
@mjasrotia updated master branch, current build failure is related to data type change for xdm:repeat which is intentional.

extensions/adobe/experience/decisioning/frequency-capping-constraints.schema.json
```diff
@@ -55,45 +55,83 @@
...
               },
+              "xdm:customQuery": {
+                "title": "Custom PQL query",
+                "description": "Custom PQL query to match events.",
+                "type": "string"
+              },
               "xdm:repeat": {
                 "title": "Repeat",
-                "description": "When true, new counter rule with updated time window will be created. Configure a capping to be repeated (e.g., max times every 1 week). ECS will support repeating config in the rules, scan rules daily, and create a new rule with new expiry when the rule is due for repeating",
-                "type": "boolean",
```
mjasrotia, line 95: Changing the type of a property is also a breaking change. If you want to change "repeat" from "boolean" to an "object", you will have to deprecate the existing "repeat" property and create another one with a different name as an object. Thanks.

jshaik-adobe: @mjasrotia there is no data created with `repeat` boolean value.  this filed was added before repeat feature implementation and after architectural design changes we decided to have the `repeat` as json object instead of boolean, Now the API changes with repeat as json object in Production and we have data with `repeat` json object.

mjasrotia:
@jshaik-adobe , I am approving this PR on your confirmation that the changed property has no data ingested and you understand the risks of changing the type of this property if there was any data ingested against it.

mjasrotia (approved):

PR #1642, opened 2022-10-11
SEG-319 Add order status attributes

kiranv926:
Please link to the issue #…


components/datatypes/data/order.schema.json
```diff
@@ -64,6 +71,46 @@
           },
           "meta:titleId": "order##xdm:orderType##title##56861",
           "meta:descriptionId": "order##xdm:orderType##description##62731"
+        },
+        "xdm:lastUpdatedAt": {
```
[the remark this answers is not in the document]
kiranv926: @prabhum2 Agree. Will change as per standards :)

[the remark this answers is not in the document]
kiranv926: @prabhum2 do we need any other changes or should be good to merge to the main?

components/datatypes/marketing/commerce.schema.json
```diff
@@ -119,6 +119,11 @@
           "$ref": "https://ns.adobe.com/xdm/data/measure",
           "description": "A product has been purchased instantly, potentially skipping the cart or checkout.",
           "meta:descriptionId": "commerce##xdm:instantPurchase##description##9211"
+        },
+        "xdm:billing": {
```
mjasrotia, line 123: Billing Address is generally a function of the Customer. We already have a billing address on one of the profile based field groups ( https://github.com/adobe/xdm/blob/master/components/fieldgroups/profile/profile-personal-details.schema.json) . Shouldn't that suffice ?
We have other addresses there as well like, mailing/shipping etc.

components/datatypes/data/order.schema.json
```diff
@@ -64,6 +71,46 @@
...
+        },
+        "xdm:createdDate": {
+          "title": "Created Date",
+          "type": "string",
+          "format": "date-time",
+          "description": "The time/date when a new order is created in the commerce system.",
+          "meta:titleId": "order##xdm:createdDate##title##86153",
+          "meta:descriptionId": "order##xdm:createdDate##description##47783"
+        },
+        "xdm:shipDate": {
```
mjasrotia, line 91: Please refer the "shipping" property in the field group https://github.com/adobe/xdm/blob/master/components/datatypes/marketing/commerce.schema.json

This property refers a "shipping" data type ( https://github.com/adobe/xdm/blob/master/components/datatypes/shipping.schema.json ) which has some shipping related information. Please add properties to this data type if there is a need to enhance this or add more shipping related properties for an order.

components/datatypes/data/order.schema.json
```diff
@@ -64,6 +71,46 @@
...
+          "title": "Ship Date",
+          "type": "string",
+          "format": "date-time",
+          "description": "The date when one or more items from an order is shipped. ",
+          "meta:titleId": "order##xdm:shipDate##title##86154",
+          "meta:descriptionId": "order##xdm:shipDate##description##47784"
+        },
+        "xdm:returnDate": {
```
mjasrotia, line 99: In traditional commerce systems, returns are generally modeled outside orders as they are pretty significant to have its own entity. We do not have modeled returns in AEP yet so, I would suggest to first of all create a proposition in a WIKI page to add "returns" properties in AEP. We can go through it and will figure out the best way to model it,

It might be modeled as its own data type and added to its own "returns" object rather than adding it under the same orders. We can discuss further.

components/datatypes/marketing/commerce.schema.json
shankar35619: Sure. We can re use the billing object within personal details.

components/datatypes/data/order.schema.json
shankar35619: I can put up a proposal together for introducing a returns datatype into the mix. If returns need to be independent of order as an object/datatype by itself, that should work for us too.

kiranv926:
Updated review comments as per suggestions. please let us know if we need anything else to be addressed?. thanks!

suranjanabasu:
@mjasrotia We have removed the returnDate field. After your discussion with @shankar35619 , we will add any fields needed  there in a separate PR. Can you please approve the PR. We have a Beta release due this month and need this PR merged. Thanks!

mjasrotia (approved):

PR #1643, opened 2022-10-19
Update media.schema.json

prabhum2:
This PR is to fix the bug that is introduced due to the change in pattern for video segment property

### DIFF
--- a/components/datatypes/media.schema.json
+++ b/components/datatypes/media.schema.json
@@ -236,7 +236,7 @@
                 "xdm:videoSegment": {
                   "title": "Video segment",
                   "type": "string",
-                  "pattern": "^[[0-9]+-[0-9]+]$",
+                  "pattern": "^[0-9]+-[0-9]+$",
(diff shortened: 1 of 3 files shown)


components/datatypes/media.schema.json
```diff
@@ -236,7 +236,7 @@
                 "xdm:videoSegment": {
                   "title": "Video segment",
                   "type": "string",
-                  "pattern": "^[[0-9]+-[0-9]+]$",
+                  "pattern": "^[0-9]+-[0-9]+$",
```
psnep, line 239: From our previous experience it won't work regardless of the pattern until the pattern is not removed. At some point as a test I created a field with a very simple pattern(to contain a digit) at that one didn't work either. So I presumed that it is a general problem regarding the pattern property of a field and it is not working as expected.

hanessian (changes requested):
Are any customers using the configuration of the old pattern? If so, then this will be be a breaking change. If not, we can go through with the change. Thanks!

PR #1649, opened 2022-11-01
Weather triggers should sit under xdm:weather

### LINKED ISSUE #1648: Weather triggers should sit under xdm:weather
- components/fieldgroups/shared/product-triggers
- components/fieldgroups/shared/severe-triggers
- components/fieldgroups/shared/relative-triggers

Weather integration in Konductor.

mflorin:
Weather triggers field groups moved under `xdm:weather`.

fixes #1648


flore77 (approved):
LGTM 👍

prabhum2:
@mflorin - changing the field name is a breaking change which will make any data set associated to a schema using this field group unusable . Please confirm that there is no existing data set in platform that is referencing this field group.

mflorin:
@prabhum2 - can you please detail a bit what "unusable" means? Does it mean the dataset is completely broken and we won't be able to reuse it or will it work again once data is sent using the new schema?

mjasrotia (approved):

PR #1650, opened 2022-11-17
[VA-5462] Remove droppedFramesCount XDM field

lidiaist:


lidiaist:
@prabhum2 please review. We realised we don't need 2 separate fields for collection and reporting for droppedFrames. If you could please move these changes into stage. Thank you!

mjasrotia (approved):

mjasrotia:
Created another PR with these changes.
https://github.com/adobe/xdm/pull/1658

PR #1652, opened 2022-11-30
Issue-1651 Adding consent policy id map v2 schema.

vedhera:
Please link to the issue #…


components/fieldgroups/profile/profile-consentResultsv2.schema.json
```diff
@@ -0,0 +1,107 @@
...
+  "title": "Consent policies evaluation results map",
+  "type": "object",
+  "meta:intendedToExtend": [
+    "https://ns.adobe.com/xdm/context/profile"
+  ],
+  "description": "Consent policies evaluation result for individual",
+  "definitions": {
+    "profile-consentResultsV2": {
+      "properties": {
+        "xdm:consentPoliciesIDMapV2": {
```
ankushkd, line 19: @mjasrotia is v2 accepted naming convention or should it be changed to something else?

mjasrotia: @ankushkd , having V2 kind of signifies that we would always have V1. Is there another suffix you guys can think of here ? How about "consentPoliciesMap" ?
cc: @vedhera

components/fieldgroups/profile/profile-consentResultsv2.schema.json
```diff
@@ -0,0 +1,89 @@
...
+      "description": "current status for a given consent policy",
+      "type": "string",
+      "enum": ["realized", "exited"],
+      "meta:enum": {
+        "realized": "Realized",
+        "exited": "Exited"
+      }
+    },
+    "profile-consentResultsV2": {
+      "type": "object",
```
mjasrotia, line 26: This is a definition. We don't need "type" here. Please remove

components/fieldgroups/profile/profile-consentResultsv2.schema.json
vedhera: I have renamed it to `consentPoliciesResultMap`. Although having v2 in the naming convention does not necessarily mean that v1 will always stick around. That's not how usually versioning works.

vedhera:
@mjasrotia Do we also need to remove the usage of referential pointer as below?
```javascript
  "xdm:status": {
                        "$ref": "#/definitions/consent-status"
                      },
```

components/fieldgroups/profile/profile-consentResultsv2.schema.json
vedhera: @mjasrotia Refer to the definition section here: http://json-schema.org/understanding-json-schema/structuring.html
It looks like the entity being referred to is an object (essentially as we are also defining `properties`), so IMO it needs to be present for completeness. LMK.

mjasrotia: Definitions are only look at as a wrapper by the schema registry. Any object property need to be inside the "properties" section. We can remove the type from the definition here.

components/fieldgroups/profile/profile-consentResultsv2.schema.json
```diff
@@ -0,0 +1,89 @@
...
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/mixins/profile-consentResultsV2",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Consent policies evaluation results map",
+  "type": "object",
+  "meta:intendedToExtend": ["https://ns.adobe.com/xdm/context/profile"],
+  "description": "Consent policies evaluation result for individual",
+  "definitions": {
+    "consent-status": {
```
mjasrotia, line 15: Also, as discussed over the slack thread, two things to note here:
1. Although this is a valid JSON schema, let's make this inline to the object where this is referred because, the UI does not yet support adding custom properties under an object which is defined as a local definition which is the case here. This was a recent issue reported and we would have to wait for UI to fix this. 
2. FYI, Any definition need to have a "properties"  section to wrap the fields defined in it. This change won't be necessary if we take care of the point# 1 above.

mjasrotia: Can we fix this as we are planning to push to INT/STG tomorrow (01/05) ?
cc: @ankushkd , @vedhera

components/fieldgroups/profile/profile-consentResultsv2.schema.json
mjasrotia: @vedhera , as you mentioned above, I don't see the change in the name yet. Are we keeping v2 in the name ?

mjasrotia (approved):

PR #1664, opened 2023-01-25
Adding Analytics context data

ninaceban:
Please link to the issue #…
https://jira.corp.adobe.com/browse/PLAT-140711

### DIFF
--- a/extensions/adobe/experience/analytics/context-data.schema.json
+++ b/extensions/adobe/experience/analytics/context-data.schema.json
@@ -1,6 +1,6 @@
 {
   "meta:license": [
-    "Copyright 2019 Adobe Systems Incorporated. All rights reserved.",
+    "Copyright 2023 Adobe Systems Incorporated. All rights reserved.",
     "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
     "you may not use this file except in compliance with the License. You may obtain a copy",
     "of the License at https://creativecommons.org/licenses/by/4.0/"
--- a/extensions/adobe/experience/analytics/context-data.schema.json
+++ b/extensions/adobe/experience/analytics/context-data.schema.json
@@ -12,12 +12,16 @@
   "description": "A free form field where Analytics context data will be captured.",
   "definitions": {
     "context-data-datatype": {
-      "title": "Analytics context data",
-      "type": "object",
-      "meta:xdmType": "map",
-      "additionalProperties": {
-        "type": "string",
-        "description": "Value of the property."
+      "properties": {
+        "xdm:context-data-datatype": {
+          "title": "Analytics context data",
+          "type": "object",
+          "meta:xdmType": "map",
+          "additionalProperties": {
+            "type": "string",
+            "description": "Value of the property."
+          }
+        }
       }
     }
   },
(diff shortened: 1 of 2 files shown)


extensions/adobe/experience/analytics/context-data.schema.json
```diff
@@ -0,0 +1,33 @@
...
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/experience/analytics/context-data",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "context-data",
+  "type": "object",
+  "description": "A free form field where Analytics context data will be captured.",
+  "definitions": {
```
mjasrotia, line 13: This definition should also have a "properties" object which will further have the map object you want to define.
Example:
`  "definitions": {
    "context-data-datatype": {
      "properties": {
        "xdm:<Map Object Name>": {
          "type": "object",
          "meta:xdmType": "map",
          "additionalProperties": {
            "type": "string",
			"description": "Your Description"
          }
        }
      }
    }
  }
`
an existing example you can refer is 
https://github.com/adobe/xdm/blob/master/components/fieldgroups/shared/identitymap.schema.json

Please make the changes.

ninaceban: thank you! I'll add the changes!

extensions/adobe/experience/analytics/context-data.schema.json
```diff
@@ -0,0 +1,37 @@
...
+    "context-data-datatype": {
+      "properties": {
+        "xdm:context-data-datatype": {
```
mjasrotia, line 16: We do not support "-" in the property names. Also, it seems like you are creating a map object to send context data so having a suffix of "datatype" in the object name would not make sense. I would suggest to change the property name to "xdm:contextData"

extensions/adobe/experience/analytics/context-data.schema.json
```diff
@@ -0,0 +1,37 @@
+{
+  "meta:license": [
+    "Copyright 2023 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/experience/analytics/context-data",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "context-data",
```
mjasrotia, line 10: Let's change this to "Context Data"

extensions/adobe/experience/analytics/context-data.schema.json
```diff
@@ -0,0 +1,37 @@
...
+  "type": "object",
+  "description": "A free form field where Analytics context data will be captured.",
```
mjasrotia, line 12: Can we add more information in this description regarding the map object and the examples of data you would typically send in this field ?

extensions/adobe/experience/analytics/context-data.schema.json
ninaceban: thank you, I'm adding that

PR #1670, opened 2023-02-16
[VA-5632] Add media reporting eventTypes

lidiaist:


lidiaist:
@mjasrotia please review, we added 5 more eventType necessary for media reporting, separate from the previous media eventTypes that we use for collection, we need to keep these separate to generate different report breakdowns in CJA (Customer Journey Analytics)

mjasrotia (approved):

PR #1676, opened 2023-02-22
[VA-5635] Add billing fields in xdm

lidiaist:


lidiaist:
@mjasrotia please review. We need to add 2 more fields used for billing by Media Analytics. These are the exact names that we use on our old API `pev3 and pccr`, does pev3 respect the naming convention or should we rename this? Thanks!

mjasrotia (approved):

PR #1677, opened 2023-02-23
category data for product list item

aniham:
https://jira.corp.adobe.com/browse/SEG-544

- category data array (as products can belong to multiple categories)
- category data contains name and path

### DIFF
--- a/components/datatypes/productlistitem.schema.json
+++ b/components/datatypes/productlistitem.schema.json
@@ -95,8 +95,8 @@
           "meta:titleId": "productlistitem##xdm:discountAmount##title##25051",
           "meta:descriptionId": "productlistitem##xdm:discountAmount##description##67611"
         },
-        "xdm:categoryData": {
-          "title": "Category data",
+        "xdm:productCategories": {
+          "title": "Product categories",
           "type": "array",
           "description": "List of category data this product falls under",
           "items": {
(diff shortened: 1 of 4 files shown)


components/datatypes/productlistitem.schema.json
```diff
@@ -94,6 +94,14 @@
           "description": "The difference between the regular price and the special price for the product.",
           "meta:titleId": "productlistitem##xdm:discountAmount##title##25051",
           "meta:descriptionId": "productlistitem##xdm:discountAmount##description##67611"
+        },
+        "xdm:categoryData": {
```
mjasrotia, line 98: can we rename this field to productCategory ? The final fields will look like below then:
category.ID
category.:name
category.path

categoryData.ID does not looks good semantically.

aniham: @mjasrotia thanks for the review!!

i renamed to `productCategories` since it's an array. is that ok or do you prefer `productCategory` still?

mjasrotia: @aniham , the plural name is much better but I realized that the fully qualified for these categories would not have this object name in that as its is an array of objects here. What I would like to suggest is to rename the datatype properties by adding a prefix of "category" like categoryID, categoryName, categoryPath.

This way, your payload should look like this:

`productCategories: [
  {
    categoryID:"C100",
    categoryName: "shoes".
    categoryPath: "Somepath"
  },
    {
    categoryID:"C200",
    categoryName: "Clothes".
    categoryPath: "Somepath"
  },
    {
    categoryID:"400",
    categoryName: "Houseware".
    categoryPath: "Somepath"
  }
]`

aniham: ✅

mjasrotia (approved):

Issue #1681, opened 2023-03-01
XDM schema for Target Classification datasets


surbhi114:
## What are the schemas that are affected by the issue
Its a brand new field group, so none.
## What are examples of products that are impacted by the issue
Adobe Target

PR #1682, opened 2023-03-01
TNT-44351 Create XDM schema for Target Classification datasets

surbhi114:
Jira:https://jira.corp.adobe.com/browse/TNT-44351
Github Issue: https://github.com/adobe/xdm/issues/1681

Summary:
A new XDM schema for Target Classification datasets containing a set of meta-data fields that will classify Target activities and experiences.

### DIFF
--- a/extensions/adobe/experience/target/classification.schema.json
+++ b/extensions/adobe/experience/target/classification.schema.json
@@ -13,73 +13,73 @@
   "definitions": {
     "classification": {
       "properties": {
-        "xdm:activityName": {
+        "https://ns.adobe.com/experience/target/classification/activityName": {
           "title": "Activity Name",
           "type": "string",
           "description": "The name of the Target activity."
         },
-        "xdm:experienceName": {
+        "https://ns.adobe.com/experience/target/classification/experienceName": {
           "title": "Experience Name",
           "type": "string",
           "description": "The name of the Target experience."
         },
-        "xdm:activityExperience": {
+        "https://ns.adobe.com/experience/target/classification/activityExperience": {
           "title": "Activity Experience",
           "type": "string",
           "description": "Target Activity name concatenated with experience name."
         },
-        "xdm:thirdId": {
+        "https://ns.adobe.com/experience/target/classification/thirdId": {
           "title": "Third Id",
           "type": "string",
           "description": "The third id of the classification record."
         },
-        "xdm:activityId": {
+        "https://ns.adobe.com/experience/target/classification/activityId": {
           "title": "Activity Id",
           "type": "string",
           "description": "The identity of Target activity."
         },
-        "xdm:experienceId": {
+        "https://ns.adobe.com/experience/target/classification/experienceId": {
           "title": "Experience Id",
           "type": "string",
           "description": "The identity of Target experience."
         },
-        "xdm:analyticsReporting": {
+        "https://ns.adobe.com/experience/target/classification/analyticsReporting": {
           "title": "Analytics Reporting",
           "type": "string",
           "description": "The reporting type associated with the Target activity."
         },
-        "xdm:activeDates": {
+        "https://ns.adobe.com/experience/target/classification/activeDates": {
           "title": "Active Dates",
           "type": "string",
           "description": "The date of Target activity activation."
         },
-        "xdm:controlExperience": {
+        "https://ns.adobe.com/experience/target/classification/controlExperience": {
           "title": "Control Experience",
           "type": "string",
           "description": "The experience set as control experience in the Target activity."
         },
-        "xdm:algorithmId": {
+        "https://ns.adobe.com/experience/target/classification/algorithmId": {
           "title": "Algorithm Id",
           "type": "string",
           "description": "The identity of the algorithm used in the Target activity."
         },
-        "xdm:controlvsTargeted": {
+        "https://ns.adobe.com/experience/target/classification/controlvsTargeted": {
           "title": "Control vs Targeted",
           "type": "string",
           "description": "The traffic allocation type for the Target activity."
         },
-        "xdm:trafficTypeId": {
+        "https://ns.adobe.com/experience/target/classification/trafficTypeId": {
           "title": "Traffic Type Id",
           "type": "string",
           "description": "The identity of the traffic type associated with Target activity."
         }
       },
       "required": [
-        "xdm:activityName",
-        "xdm:activityId",
-        "xdm:experienceName",
-        "xdm:experienceId",
-        "xdm:activityExperience"
+        "https://ns.adobe.com/experience/target/classification/activityName",
+        "https://ns.adobe.com/experience/target/classification/activityId",
+        "https://ns.adobe.com/experience/target/classification/experienceName",
+        "https://ns.adobe.com/experience/target/classification/experienceId",
+        "https://ns.adobe.com/experience/target/classification/activityExperience"
       ]
     }
   },
--- a/extensions/adobe/experience/target/classification.schema.json
+++ b/extensions/adobe/experience/target/classification.schema.json
@@ -1,0 +1,96 @@
+{
+  "meta:license": [
+    "Copyright 2023 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/experience/target/classification",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Adobe Target Classification Fields",
+  "type": "object",
+  "description": "A set of meta-data fields that will classify Target activities and experiences.",
+  "definitions": {
+    "classification": {
+      "properties": {
+        "xdm:activityName": {
+          "title": "Activity Name",
+          "type": "string",
+          "description": "The name of the Target activity."
+        },
+        "xdm:experienceName": {
+          "title": "Experience Name",
+          "type": "string",
+          "description": "The name of the Target experience."
+        },
+        "xdm:activityExperience": {
+          "title": "Activity Experience",
+          "type": "string",
+          "description": "Target Activity name concatenated with experience name."
+        },
+        "xdm:thirdId": {
+          "title": "Third Id",
+          "type": "string",
+          "description": "The third id of the classification record."
+        },
+        "xdm:activityId": {
+          "title": "Activity Id",
+          "type": "string",
+          "description": "The identity of Target activity."
+        },
+        "xdm:experienceId": {
+          "title": "Experience Id",
+          "type": "string",
+          "description": "The identity of Target experience."
+        },
+        "xdm:analyticsReporting": {
+          "title": "Analytics Reporting",
+          "type": "string",
+          "description": "The reporting type associated with the Target activity."
+        },
+        "xdm:activeDates": {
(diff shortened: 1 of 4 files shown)


extensions/adobe/experience/target/classification.schema.json
```diff
@@ -0,0 +1,97 @@
...
+                    "title": "Activity Name",
+                    "type": "string",
+                    "description": "The name of the Target activity."
+                },
+                "xdm:experienceName": {
+                    "title": "Experience Name",
+                    "type": "string",
+                    "description": "The name of the Target experience."
+                },
+                "xdm:activity>experience": {
```
shagreel, line 31: I don't think that '>' is a valid character in a name. You'll need to name it something like activityExperience. This is likely what is breaking the build.

extensions/adobe/experience/target/classification.schema.json
```diff
@@ -0,0 +1,97 @@
...
+    ],
+    "$id": "https://ns.adobe.com/experience/target/classification",
+    "$schema": "http://json-schema.org/draft-06/schema#",
+    "title": "Adobe Target Classification Fields",
+    "type": "object",
+    "description": "A set of meta-data fields that will classify Target activities and experiences.",
+    "definitions": {
+        "classification": {
+            "properties": {
+                "xdm:key": {
```
shagreel, line 16: Do you plan to put anything in the _id field? Will you duplicate the key or will you insert a uuid?

extensions/adobe/experience/target/classification.schema.json
```diff
@@ -0,0 +1,97 @@
...
+                "xdm:ExperienceId"
+            ]
+        }
+    },
+    "allOf": [
+        {
+            "$ref": "#/definitions/classification"
+        }
+    ],
+    "meta:status": "stable"
```
shagreel, line 96: You should mark this as experimental for this first PR. https://github.com/adobe/xdm/blob/master/CONTRIBUTING.md#specify-the-metastatus

extensions/adobe/experience/target/classification.schema.json
```diff
@@ -0,0 +1,97 @@
+{
```
shagreel, line 1: If you are intending this to be a class then you need to add
`"meta:extends": [
    "https://ns.adobe.com/xdm/data/record"
  ],`
If you are intending this to be a field group then you need to add
`"meta:intendedToExtend": [
    "https://ns.adobe.com/xdm/data/record"
  ],`

shagreel (changes requested):
You must extend the record data type or CJA will not be able to use the dataset.

extensions/adobe/experience/target/classification.schema.json
surbhi114: Removing Key, will add the correlationId as _id.

extensions/adobe/experience/target/classification.schema.json
```diff
@@ -0,0 +1,93 @@
...
+                    "description": "The traffic allocation type for the Target activity."
+                },
+                "xdm:trafficTypeId": {
+                    "title": "Traffic Type Id",
+                    "type": "string",
+                    "description": "The identity of the traffic type associated with Target activity."
+                }
+            },
+            "required": [
+                "xdm:Key",
```
shagreel, line 78: You will need to remove this since you removed the key field.

shagreel (changes requested):

extensions/adobe/experience/target/classification.schema.json
surbhi114: I missed adding this

shagreel:
@mjasrotia This new class should be ready to merge now.

extensions/adobe/experience/target/classification.schema.json
```diff
@@ -0,0 +1,96 @@
+{
+  "meta:license": [
+    "Copyright 2023 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/experience/target/classification",
```
mjasrotia, line 8: @surbhi114 , Are you defining a class here or just a data type ?

surbhi114: I would like it to be a class that can be used for creating schemas, and later on we can query the schema registry using the included class to get back the schema.

mjasrotia: Ok, In that case, I would suggest you to fully qualify the property namespaces. We only use xdm: for a property within an internal Adobe solution XDM object if that object is a data type. 

Here's an example of a fully qualified property namespace. Observe the properties in this link.

https://github.com/adobe/xdm/blob/master/extensions/adobe/experience/customerJourneyManagement/ajo-entity-mixins.schema.json

Please work within your team to come up with the right hierarchy for your use cases. 

Also, as a general rule, we try to keep only those properties in a class which are mandatory to define the entity the class represents. Rest of the properties should sit under respective field groups as per their logical groupings.

schemas/target/target-classification.example.1.json
```diff
@@ -0,0 +1,8 @@
+{
+  "@id": "123456:2:0:400001",
+  "xdm:activityName": "My Activity A",
+  "xdm:activityId": "123456",
+  "xdm:experienceName": "My Experience 2",
+  "xdm:experienceId": "2",
+  "xdm:activityExperience": "My Activity A > My Experience 2"
+}
```
shagreel, line 8: Because this activity has a 4th value in the id, I would expect that the algorithm fields are also filled out.

shagreel (approved):

schemas/target/target-classification.schema.json
```diff
@@ -0,0 +1,98 @@
...
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Adobe Target Classification Fields",
+  "description": "A set of meta-data fields that will classify Target activities and experiences.",
+  "type": "object",
+  "meta:extensible": false,
+  "meta:abstract": false,
+  "meta:extends": ["https://ns.adobe.com/xdm/data/record"],
+  "definitions": {
+    "target": {
+      "properties": {
```
mjasrotia, line 18: As we discussed during our call last week, could you please add a top level object field here which will give some context to the nested properties below ? something like
`targetClassification: {
   activityName:
  experienceName
  etc.....
}
`

schemas/target/target-classification.schema.json
```diff
@@ -0,0 +1,98 @@
...
+          "title": "Experience Name",
+          "type": "string",
+          "description": "The name of the Target experience."
+        },
+        "xdm:activityExperience": {
+          "title": "Activity Experience",
+          "type": "string",
+          "description": "Target Activity name concatenated with experience name."
+        },
+        "xdm:thirdId": {
```
mjasrotia, line 34: Id ==> ID

schemas/target/target-classification.schema.json
```diff
@@ -0,0 +1,98 @@
...
+          "title": "Third Id",
+          "type": "string",
+          "description": "The third id of the classification record."
+        },
+        "xdm:activityId": {
```
mjasrotia, line 39: Id ==> ID

schemas/target/target-classification.schema.json
```diff
@@ -0,0 +1,98 @@
...
+          "title": "Experience Id",
+          "type": "string",
+          "description": "The identity of Target experience."
+        },
+        "xdm:analyticsReporting": {
+          "title": "Analytics Reporting",
+          "type": "string",
+          "description": "The reporting type associated with the Target activity."
+        },
+        "xdm:activeDates": {
```
mjasrotia, line 54: If its not an array, then why plurals ? Please rename this to activityDate

Also, please add a format="date-time"

schemas/target/target-classification.schema.json
```diff
@@ -0,0 +1,98 @@
...
+          "title": "Active Dates",
+          "type": "string",
+          "description": "The date of Target activity activation."
+        },
+        "xdm:controlExperience": {
```
mjasrotia, line 59: Should this be a boolean of its just a flag whether the experience is control or not ?

schemas/target/target-classification.schema.json
```diff
@@ -0,0 +1,98 @@
...
+          "title": "Control Experience",
+          "type": "string",
+          "description": "The experience set as control experience in the Target activity."
+        },
+        "xdm:algorithmId": {
```
mjasrotia, line 64: Id ==> ID

schemas/target/target-classification.schema.json
```diff
@@ -0,0 +1,98 @@
...
+          "title": "Algorithm Id",
+          "type": "string",
+          "description": "The identity of the algorithm used in the Target activity."
+        },
+        "xdm:controlvsTargeted": {
+          "title": "Control vs Targeted",
+          "type": "string",
+          "description": "The traffic allocation type for the Target activity."
+        },
+        "xdm:trafficTypeId": {
```
mjasrotia, line 74: Id ==> ID

schemas/target/target-classification.schema.json
surbhi114: The top level object is "target", if that seems too generic, I can rename it to targetClassification

schemas/target/target-classification.schema.json
surbhi114: The string is a combination of start date and end date, separated by '|', hence activedates plural.

schemas/target/target-classification.schema.json
surbhi114: The values are "0"/"1", thats how Analytics gets the data, unsure if should change it for CJA4T. @shagreel any pointers here?

schemas/target/target-classification.schema.json
```diff
@@ -0,0 +1,98 @@
...
+          "title": "Control Experience",
+          "type": "string",
+          "description": "The experience set as control experience in the Target activity."
+        },
+        "xdm:algorithmId": {
+          "title": "Algorithm Id",
+          "type": "string",
+          "description": "The identity of the algorithm used in the Target activity."
+        },
+        "xdm:controlvsTargeted": {
```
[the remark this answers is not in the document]
surbhi114: No, this should be a string with values like "Targeted" and "control".

schemas/target/target-classification.schema.json
surbhi114: Renamed it to targetClassification

schemas/target/target-classification.schema.json
shagreel: I think it would be fine to change. I don't know of any plans to convert the current analytics report to CJA. Customers will have to build reports manually. So small changes to make the data clear would be fine.

schemas/target/target-classification.schema.json
shagreel: We could create two new fields to separate the two dates. It would make them more usable.

mjasrotia: agreed with @shagreel . Having semantically well defined  dates is much better than concatenated strings which make them less usable downstream as dates.

mjasrotia: If the decision as per slack discussion is to keep it like this then I would recommend to rename this to activityStartAndEndDate

schemas/target/target-classification.schema.json
surbhi114: We decided on the slack chat that we keep this intact.

mjasrotia (approved):

PR #1685, opened 2023-03-09
refund info for order

aniham:
https://jira.corp.adobe.com/browse/SEG-545

- refund data for order
- refund data for product list item (line-item level returns/adjustments)


components/datatypes/data/refunditem.schema.json
```diff
@@ -0,0 +1,69 @@
+{
```
mjasrotia, line 1: Please move this to the one folder above 
.../components/datatypes/refunditem.schema.json
All the example files will also move along.

aniham: @mjasrotia done. i had it in the same directory as `payment` which i see as kind of the inverse/equivalent of it. should we move `payment` out too?

mjasrotia: I can move them later. Few old data types are in that directory and somehow hardcoded in the code so, can't move without testing. Any new ones, we tend to create under the datatype folder itself.

components/datatypes/refunditem.schema.json
```diff
@@ -0,0 +1,69 @@
...
+            "return": "Return",
+            "price_adjustment": "Price Adjustment"
+          }
+        },
+        "xdm:refundPaymentType": {
+          "title": "Refund Payment Type",
+          "type": "string",
+          "description": "The method of payment for this order. Enumerated, custom values allowed.",
+          "default": "other",
+          "meta:enum": {
```
mjasrotia, line 41: If you have such a long list of meta:enums here , do think it will be a good idea to also add enums to restrict the data to this list of values. Late on we can always add more if required. starting with an open ended list would later make it impossible to restrict as it's a breaking change.
Having a well defined list would also help in segmentation queries.

mjasrotia: Same is true for all other meta:enums

aniham: @mjasrotia i'm probably misunderstanding this comment but doesn't the enum ensure these are the only accept values? ware you saying add `enums` for other fields? if so, which fields?

mjasrotia: @aniham , meta:enums is just the suggested display names for the enums. Actual enforcement is done through "enum" like below:

"enum": [
                        "not_provided",
                        "pending",
                        "in",
                        "out"
                    ],
                    "meta:enum": {
                        "not_provided": "Not provided",
                        "pending": "Pending verification",
                        "in": "Opt-in: the user explicitly consents to receiving messages.",
                        "out": "Opt-out: the user declines to receive any messages on this channel."
                    }

aniham: aaah gotcha!! thank you @mjasrotia and fixed!!

mjasrotia: Could you please fix the lint issue ?

aniham: sorry! ✅

mjasrotia (approved):

components/datatypes/refunditem.schema.json
mjasrotia: @aniham , I updated the branch from the master and I see some conflicts. Could you please resolve the conflicts on your PR ?

PR #1693, opened 2023-03-27
[VA-5758] Separate chapterDetails reporting and collection schemas

rdecea:
Please link to the issue #…


rdecea:
Details that sit behind this refactoring are covered in this decision record: https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=analytics&title=XDM+Media+Model+for+collection+vs+reporting

PR #1697, opened 2023-03-30
Removing unused fields from Web Referrer 

ninaceban:
Link to the issue - [PDCL-10206](https://jira.corp.adobe.com/browse/PDCL-10206)


jfkhoury (approved):

jfkhoury:
Details from the ticket:

`We have agreed with the Analytics team to work on a new field group to send the buffered events from the previous page, to be more Platform specific. Since these fields will not be used to send previous link interaction with the page-view event, it's better to remove them to keep the schema as clean as possible.`

mjasrotia:
@ninaceban , we don't allow breaking changes in our schemas. If you no longer need some properties, please deprecate those properties rather than delete them.

mjasrotia:
closing because it has been open from very long.

PR #1703, opened 2023-04-03
[PREFACTORING] Separate media details in collection and reporting

rdecea:
Issue: [#1550](https://github.com/adobe/xdm/issues/1550)


rdecea:
@mjasrotia this is a pre-factoring commit.. It doesn't make much sense on its own. It will make more sense after subsequent PRs, or checking out [our documented decision](https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=analytics&title=XDM+Media+Model+for+collection+vs+reporting)

PR #1716, opened 2023-04-19
Extract sessionDetails.pccr as a reporting-only field

rdecea:
Please link to the issue #…


components/datatypes/sessiondetails.schema.json
```diff
@@ -338,11 +333,15 @@
...
+          "title": "Pccr",
+          "type": "boolean",
+          "description": "Indicates that a redirect occurred."
+        }
+      }
     }
-  },
-  "allOf": [
-    {
-      "$ref": "#/definitions/sessionDetails"
```
mjasrotia, line 345: Need to add both the local definitions here

rdecea: Done

mjasrotia (approved):

PR #1724, opened 2023-05-08
XDM -1723 updating DirectMail Feedback

nikshar:
Please link to the issue #…
[XDM -1723 updating DirectMail Feedback](https://github.com/adobe/xdm/issues/1723)


nikshar:
@chetanmeh @sawhneyh @mjasrotia : Please take a look

extensions/adobe/experience/customerJourneyManagement/messageprofile.schema.json
```diff
@@ -89,6 +89,26 @@
...
+            "sftp": "SFTP Routing Destination",
+            "azure": "Azure Routing Destination "
+          }
+        },
+        "https://ns.adobe.com/experience/customerJourneyManagement/directMailChannelContext/directMailFileName": {
+          "title": "Direct Mail FileName",
+          "type": "string",
+          "description": "Direct Mail FileName"
+        },
+        "https://ns.adobe.com/experience/customerJourneyManagement/directMailChannelContext/directMaildestinationId": {
```
mjasrotia, line 108: directMaildestinationId ==> directMaildestinationID

nikshar: Thanks @mjasrotia 
Updated

sawhneyh: @nikshar : It seems fields under directMailChannelContext are not event specific but action specific. I don't think we should be ingesting action level fields in user specific events.

We should review whether we can use entity dataset for ingesting these fields.

cc:  @chetanmeh

nikshar:
@anshchauhan :  I've updated the entity schema for Direct mail Channel
Please take a look

extensions/adobe/experience/customerJourneyManagement/ajo-entity-mixins.schema.json
```diff
@@ -208,6 +208,28 @@
...
+          "type": "object",
+          "description": "Direct Mail Channel Specific Fields.",
+          "properties": {
+            "https://ns.adobe.com/experience/customerJourneyManagement/entities/routingDestination": {
+              "title": "Direct Mail Routing Destination",
+              "type": "string",
+              "description": "Direct Mail Routing Destination, non-personalized"
+            },
+            "https://ns.adobe.com/experience/customerJourneyManagement/entities/directMailFileName": {
+              "title": "Direct Mail FileName",
```
sawhneyh, line 223: @nikshar: Is this property a function of batch recurrence or a property of messagePublication?

As per my understanding we add entry in entityDataset at messagePublication, so if this property is at batch recurrence level we won't be able to override this value or preserve value at batch recurrence.

cc: @chetanmeh

nikshar: @sawhneyh: for DM we'll ingest this at export time
at message publication time we don't ingest this value.
this is a batch level property

sawhneyh: In that case, are we changing the key for this dataset also. Currently key for this datadet is a function of MessagePublicationId and VariantId.

mjasrotia:
closing because it has been open from very long. Please re-open if required

PR #1727, opened 2023-05-12
add currencyCode to shipping.schema.json

websterlove:
[https://jira.corp.adobe.com/browse/SEG-763](https://jira.corp.adobe.com/browse/SEG-763)

The shippingAmount field should have an associated currencyCode

### DIFF
--- a/components/datatypes/shipping.schema.json
+++ b/components/datatypes/shipping.schema.json
@@ -23,6 +23,13 @@
           "type": "number",
           "description": "The amount the customer had to pay for shipping."
         },
+        "xdm:currencyCode": {
+          "title": "Currency code",
+          "type": "string",
+          "examples": ["USD", "EUR"],
+          "pattern": "^[A-Z]{3}$",
+          "description": "The ISO 4217 alphabetic currency code used for pricing the product."
+        },
         "xdm:shippingDestination": {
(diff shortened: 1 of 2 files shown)


components/datatypes/shipping.schema.json
```diff
@@ -23,6 +23,13 @@
           "type": "number",
           "description": "The amount the customer had to pay for shipping."
         },
+        "xdm:currencyCode": {
```
mjasrotia, line 26: @websterlove , we do have a currency data type defined already which you can use in whatever context you intend to. 
https://github.com/adobe/xdm/blob/master/components/datatypes/currency.schema.json
Please use this existing data type wherever you want to add currency.

websterlove: Hi @mjasrotia - I was following the pattern from our other field groups, e.g. [https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=ACBI&title=Power+BI+Access+Process](https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=ACBI&title=Power+BI+Access+Process). How do I determine which way is correct in which situation? And also, since we already have that other format in our schemas, it seems simpler for users to see the same format throughout in the data we send.

mjasrotia (approved):

PR #1728, opened 2023-05-15
Product-specific identity deletes for Data Hygiene

perlmonger42:
[PLAT-151491](https://jira.corp.adobe.com/browse/PLAT-151491)

Add a new field, `targetProducts`, which is an array of strings intended to contain some subset of `["datalake", "identity", "profile"]`. Products that consume the HygieneOps dataset should ignore batches that contain a `targetProducts` array unless their product name appears in the list.

The contents of the array are represented as a strings, and are not constrained by an enumeration. This will allow new products to be supported later without modifying the XDM.

### DIFF
--- a/schemas/hygiene/aep-hygiene-ops-record.schema.json
+++ b/schemas/hygiene/aep-hygiene-ops-record.schema.json
@@ -49,13 +49,13 @@
           "meta:titleId": "aep-hygiene-ops##xdm:workorderID##title##7131",
           "meta:descriptionId": "aep-hygiene-ops##xdm:workorderID##description##50431"
         },
-        "xdm:targetProducts": {
-          "title": "Products that should process this request.",
-          "description": "Array of products to which these requests apply. If this field is absent, all products should process these requests.",
+        "xdm:targetServices": {
+          "title": "Services that should process this request.",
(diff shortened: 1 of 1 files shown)


schemas/hygiene/aep-hygiene-ops-record.schema.json
```diff
@@ -49,6 +49,16 @@
           "meta:titleId": "aep-hygiene-ops##xdm:workorderID##title##7131",
           "meta:descriptionId": "aep-hygiene-ops##xdm:workorderID##description##50431"
         },
+        "xdm:targetProducts": {
```
mjasrotia, line 52: I would suggest to rename this to targetServices as it is more intuitive of what we mean here.

perlmonger42: Perhaps that would make sense for some teams, but it would confuse my team. We use consistently use "product" throughout our code, documentation, and JSON API's to refer to the various product teams we work with, such as `identity`, `profile`, and `datalake`.

perlmonger42: After consulting with my team, we've decided to accept your suggestion.

mjasrotia (approved):

PR #1730, opened 2023-05-16
add discountAmount to order.schema.json

websterlove:
[https://jira.corp.adobe.com/browse/SEG-764](https://jira.corp.adobe.com/browse/SEG-764)

Discounts can also be on the entire order amount (not just on individual items within the order)

### DIFF
--- a/components/datatypes/data/order.schema.json
+++ b/components/datatypes/data/order.schema.json
@@ -61,6 +61,11 @@
           "meta:titleId": "order##xdm:taxAmount##title##81954",
           "meta:descriptionId": "order##xdm:taxAmount##description##51205"
         },
+        "xdm:discountAmount": {
+          "title": "Discount Amount",
+          "type": "number",
+          "description": "The difference between the regular price and the special price for the entire order."
+        },
         "xdm:priceTotal": {
           "title": "Price Total",
           "type": "number",
--- a/components/datatypes/data/order.schema.json
+++ b/components/datatypes/data/order.schema.json
@@ -61,10 +61,10 @@
           "meta:titleId": "order##xdm:taxAmount##title##81954",
           "meta:descriptionId": "order##xdm:taxAmount##description##51205"
         },
-        "xdm:discountAmount": {
+        "xdm:DiscountAmount": {
           "title": "Discount Amount",
           "type": "number",
-          "description": "The difference between the regular price and the special price for the entire order."
+          "description": "The difference between the regular price and the special price - applied to the entire order, rather than to individual products."
         },
         "xdm:priceTotal": {
           "title": "Price Total",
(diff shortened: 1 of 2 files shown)


components/datatypes/data/order.schema.json
```diff
@@ -61,6 +61,11 @@
           "meta:titleId": "order##xdm:taxAmount##title##81954",
           "meta:descriptionId": "order##xdm:taxAmount##description##51205"
         },
+        "xdm:discountAmount": {
```
mjasrotia, line 64: @websterlove , in the order data type there already another property defined for payments. That property refers to a data type "paymentItems". Do you think that the new property "discountAmount" belongs there ?

https://github.com/adobe/xdm/blob/master/components/datatypes/data/paymentitem.schema.json

mjasrotia: Oh ok, if it represents total discount amount at the order header level and not at the individual line item level, then it should belong here. I would suggest to rename this to totalDiscountAmount in that case to be very clear. Agreed ?

websterlove: If we rename it, I think orderDiscountAmount might be clearer. I used discountAmount to match the field name used in [https://github.com/adobe/xdm/blob/master/components/datatypes/productlistitem.schema.json](https://github.com/adobe/xdm/blob/master/components/datatypes/productlistitem.schema.json) (for the item-level discounts), thinking that consistency would be a good idea. Which option do you think is best?

mjasrotia: I am good with orderDiscountAmount too as long it conveys the idea :) Thanks !

websterlove: Had another thought while making this edit - do you think it would work better to adjust the field description to be more explicitly about the order-level (rather than item-level) discount, and leave the field name more consistent with the other fields? (For example, priceTotal is the same field name here and in productListItem.)

mjasrotia: As a general rule of thumb, we should always keep the descriptions very detailed and intuitive. Please go ahead and change the description accordingly.

components/datatypes/data/order.schema.json
```diff
@@ -61,6 +61,11 @@
           "meta:titleId": "order##xdm:taxAmount##title##81954",
           "meta:descriptionId": "order##xdm:taxAmount##description##51205"
         },
+        "xdm:DiscountAmount": {
```
mjasrotia, line 64: One last thing, we tend to keep property names start with a small case "discountAmount"

websterlove: Oops, yeah, that was a typo

mjasrotia (approved):

PR #1732, opened 2023-05-16
[CJM-48273] add decisioning asset data type

usankara:
###  PR Summary
- Add a data type to represent assets used in decisioning.
- Data type is not expected to create a hierarchy for itself in AEP UI view.
- JIRA - https://jira.corp.adobe.com/browse/CJM-48273
@mjasrotia , @guayang-adobe - please review.


extensions/adobe/experience/decisioning/decisioning-asset.schema.json
```diff
@@ -0,0 +1,61 @@
...
+          "type": "string",
+          "description": "source URL of the asset. eg) URL of an image asset http://cdn.com/myimage.png"
+        },
+        "xdm:destinationURL": {
+          "title": "destinationURL",
+          "type": "string",
+          "description": "certain assets like images can have a destination, when clicked upon. eg) http://ecommerce.com/product"
+        }
+      },
+      "required": ["xdm:repository", "xdm:sourceURL"]
```
mjasrotia, line 49: @usankara , let's put the required in the field group where you intend to use it. This data type could be used elsewhere too.

extensions/adobe/experience/decisioning/decisioning-asset.schema.json
```diff
@@ -0,0 +1,61 @@
...
+    }
+  },
+  "allOf": [
+    {
+      "$ref": "https://ns.adobe.com/xdm/common/extensible#/definitions/@context"
```
mjasrotia, line 54: Let's remove this . We don't need this

extensions/adobe/experience/decisioning/decisioning-asset.schema.json
usankara: @mjasrotia , ok , sure. do we need to change the datatype name to "asset" instead of "decisioning-asset" and move to a common directory at higher level (say https://github.com/adobe/xdm/tree/master/components/datatypes) ..?

extensions/adobe/experience/decisioning/decisioning-asset.schema.json
usankara: ok, sure

extensions/adobe/experience/decisioning/decisioning-asset.schema.json
mjasrotia: asset in general could be very generic and can be interpreted in a million ways, document as an asset, service as an asset , application as an asset and so on. To me , it deserves its own class with an assetID, name and then corresponidng field groups defining different assets. Let's keep this one here for your use cases :)

guayang-adobe:
Is there additional field needed for AEM use case? Like AEM asset id? Not sure how/if AEM asset picker can show existing selection for editing case.

guayang-adobe (approved):

mjasrotia (approved):

usankara:
Thank you @mjasrotia, @guayang-adobe for reviewing the PR!

PR #1733, opened 2023-05-17
[VA-5837] Add cdn field in xdm schema in mediaCollection and mediaReporting

lidiaist:


lidiaist:
@mjasrotia please review, this PR adds a field on the sessionDetails datatype which will be needed for both mediaCollection and mediaReporting

PR #1734, opened 2023-05-22
add billing info to commerce.schema.json

websterlove:
[https://jira.corp.adobe.com/browse/SEG-765](https://jira.corp.adobe.com/browse/SEG-765)

Include billing info in a parallel location to shipping info (in commerce schema), for easier navigation by users


components/datatypes/marketing/commerce.schema.json
```diff
@@ -115,6 +115,10 @@
           "description": "Shipping details for one or more products.",
           "meta:descriptionId": "commerce##xdm:shipping##description##26111"
         },
+        "xdm:billing": {
```
mjasrotia, line 118: If you are just adding a billing address, you can just refer the address data type here 
"https://ns.adobe.com/xdm/common/address",

There is no need to create the billing data type yet unless you plan to add more billing related information in that data type. 

Also, by using the standard address type, you would avoid adding one unnecessary hierarchy here.

mjasrotia: Agreed, even there are more fields expected for future under billing, we should still define billing address as its own object referring to the address data type. In future, we can have more properties added to the billing data type if required.

mjasrotia (approved):

PR #1736, opened 2023-05-24
[VA-5970] Change bitrateAverageBucket XDM type to String

lidiaist:
As we are doing the final end to end tests, we discovered we need to have this field of bitrateAverageBucket defined as a String, otherwise, we get platform errors due to our reporting logic.
I see for the build it says changing the type of this field would be a breaking change, this is why we prefer to do this before making the schemas stable, so far clients would not be impacted yet.
This is a field we use only on mediaReporting, so clients don't specify this field themselves we just send it on mediaReporting (after computing it internally) in the final reports from the platform.


mjasrotia (approved):

mjasrotia:
Approving because its an experimental data type and is still going through testing by the media analytics team. They also confirmed that this field is only use iinternally

PR #1737, opened 2023-05-31
Adding new field groups for describing point of interest details with location.entry/exit experience events

iewilliam:
This PR introduces 2 new experience event field groups for attaching location details to location.entry/exit events:
- Places Service field group: for attaching basic information about the point of interest (POI) where the event took place. This is simply the POI's unique ID in Places Service. 
- Places Service Advanced field group: for attaching more detailed information about the point of interest. So far this includes the POI's name and its metadata in Places Service. 

In most cases the user would include Places Service field group only in an experience event dataset. If the user would like to create rules/segments/etc that requires matching POIs with certain name/metadata, then the Places Service Advanced field group could also be added to the dataset.


components/fieldgroups/experience-event/experienceevent-places-service-advanced.schema.json
```diff
@@ -0,0 +1,47 @@
+{
+  "meta:license": [
```
mjasrotia, line 2: If these two field groups are functionally same and the "advanced" version is just adding some properties on top of the other one, we should combine them into one and just enhance the description of the property.

mjasrotia: Also, we do have a data type already existing which contains the point of interest details.

https://github.com/adobe/xdm/blob/master/components/datatypes/placecontext.schema.json

This is used inside "Environment Details" Field Group. Please see if you can extend these field groups to add your additional attributes.

mjasrotia:
Closing this because it has been here since May . Please re-open if required

PR #1738, opened 2023-06-01
add scoping info to commerce.schema.json

websterlove:
[https://jira.corp.adobe.com/browse/SEG-828](https://jira.corp.adobe.com/browse/SEG-828)

Commerce scoping info allows merchants to differentiate where each of their events took place across multiple store views/stores/websites within an Adobe Commerce environment


websterlove:
@aniham @suranjanabasu Please check if you think the fields I've added here match your understanding of what we need to capture. In particular, I'd like to make sure the field names and descriptions are accurate and clear.

websterlove:
After further investigation I'm told that both environment ID and environment name are unique per merchant. Given that, I'd prefer to use environment name, since it's more human-readable. But it looks like Order Service currently only includes env ID (not sure about the data available to storefront events - Ani?). My proposal is: for now, stick with env ID, to get implementation done more quickly, but add env name later once we can make that data also available

components/datatypes/commercescope.example.1.json
```diff
@@ -0,0 +1,6 @@
+{
+  "xdm:environment": "27a15179-5545-4d23-e654-1b4ac0b42bd3",
```
aniham, line 2: should be `environmentId`

components/datatypes/commercescope.example.1.json
```diff
@@ -0,0 +1,6 @@
...
+  "xdm:website": "base",
```
aniham, line 3: should be `websiteCode`

components/datatypes/commercescope.example.1.json
```diff
@@ -0,0 +1,6 @@
...
+  "xdm:store": "main_website_store",
```
aniham, line 4: should be `storeCode`

components/datatypes/commercescope.example.1.json
```diff
@@ -0,0 +1,6 @@
...
+  "xdm:storeView": "default"
```
aniham, line 5: should be `storeViewCode`

components/datatypes/commercescope.schema.json
```diff
@@ -0,0 +1,45 @@
...
+        "xdm:store": {
+          "title": "Store Code",
+          "type": "string",
+          "description": "The store code. Many stores may exist per website"
+        },
+        "xdm:storeView": {
+          "title": "Store View Code",
+          "type": "string",
+          "description": "The store view code. Many store views may exist per store"
+        }
```
aniham, line 35: `environmentId`, `websiteCode`, `storeCode`, `storeViewCode`

aniham (changes requested):
@websterlove sorry for not clarifying earlier. i listed out the correct names for the fields as they are used by commerce to uniquely identify a client

to your question on environment name vs id - it should be id which is the unique identifier

hope this helps!!

aniham (approved):

mjasrotia:
@websterlove , I am closing this as this was taken care of in the below PR
\https://github.com/adobe/xdm/pull/1747

PR #1739, opened 2023-06-07
[VA-5978] Make mediaReporting fields not required in xdm schema

lidiaist:

In the process of preparing for making our media xdm schemas stable we need the following 2 commits:

Context:
Initially we wanted to make a clear separation between mediaCollection and mediaReporting schemas but unfortunately in order to not duplicate fields we need to use the same datatype (mediaDetails) for both.
We documented the Manage related fields usage in the AEP UI in this particular case such that clients know which fields are used for collection and which for reporting (Set up the schema in Adobe Experience Platform https://experienceleague.adobe.com/docs/media-analytics/using/implementation/implementation-edge.html?lang=en) but we would like to give the clients the possibility of using the schemas without a mandatory step of hiding fields, therefore in this PR we remove the required fields which are used on mediaReporting in our schemas, and only required mediaCollection fields remain.

First commit: 
- from sessionDetails remove the required check for ID because this is mediaReporting only field and clients only complete mediaCollection so if they choose not to hide mediaReporting fields from the schema, they should still be able to send us requests 
- from chapterDetails remove the required check for isStarted, isCompleted, timePlayed because these are mediaReporting only fields and clients only complete mediaCollection so if they choose not to hide mediaReporting fields from the schema, they should still be able to send us requests 
- from advertisingPodDetails remove the required check for ID

Second commit: Reverts https://github.com/adobe/xdm/pull/1716 the extraction of pccr field in only mediaReporting as these changes did not work (the AEP schema UI now does not show the pccr field anywhere not in mediaColecton and not in mediaReporting)

### DIFF
--- a/components/datatypes/advertisingpoddetails.example.1.json
+++ b/components/datatypes/advertisingpoddetails.example.1.json
@@ -1,5 +1,4 @@
 {
   "xdm:index": 2,
-  "xdm:offset": 123,
-  "xdm:ID": "123"
+  "xdm:offset": 123
 }
--- a/components/datatypes/advertisingpoddetails.schema.json
+++ b/components/datatypes/advertisingpoddetails.schema.json
@@ -36,7 +36,7 @@
           "description": "The offset of the ad break inside the content, in seconds."
         }
       },
-      "required": ["xdm:index", "xdm:offset", "xdm:ID"]
+      "required": ["xdm:index", "xdm:offset"]
     }
   },
   "allOf": [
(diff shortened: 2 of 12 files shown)


lidiaist:
I readded the sessionDetails.ID field, just removed the required check.
So there should be no breaking changes with this PR because we only remove required checks and add back the pccr field.

mjasrotia (approved):

lidiaist:
The 3rd commit removes the required check for advertisingPodDetails.index which is only present on mediaCollection and not present on mediaReporting, so this is the reversed case to the first commit which contained fields not present on collection but present on reporting.

PR #1761, opened 2023-07-27
Context Data map object for Analytics Usage

mjasrotia:
Please link to the issue #…


extensions/adobe/experience/analytics/context-data.schema.json
```diff
@@ -0,0 +1,37 @@
...
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/experience/analytics/context-data",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "meta:extensible": true,
+  "meta:abstract": true,
+  "meta:intendedToExtend": ["https://ns.adobe.com/xdm/context/experienceevent"],
+  "title": "Context Data",
```
clubycoder, line 13: I would change this to "Adobe Analytics Context Data" so we can tell users to search for "Adobe Analytics" and add both  "Adobe Analytics ExperienceEvent Full Extension" and "Adobe Analytics Context Data".

extensions/adobe/experience/analytics/context-data.schema.json
```diff
@@ -0,0 +1,37 @@
...
+  "type": "object",
+  "description": "A free form field where Analytics context data will be captured.",
+  "definitions": {
+    "context-data": {
+      "properties": {
+        "https://ns.adobe.com/experience/analytics/contextData": {
+          "title": "Context Data Map",
```
clubycoder, line 20: I would just call the field "Context Data" to match what people know it as.

extensions/adobe/experience/analytics/context-data.schema.json
```diff
@@ -0,0 +1,37 @@
...
+    "Copyright 2023 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/experience/analytics/context-data",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "meta:extensible": true,
+  "meta:abstract": true,
+  "meta:intendedToExtend": ["https://ns.adobe.com/xdm/context/experienceevent"],
```
clubycoder, line 12: Can you have field-groups that are restricted to only be used when you have already added another field group?  Should this be changed to [https://ns.adobe.com/experience/analytics/experienceevent-all](https://github.com/adobe/xdm/blob/master/extensions/adobe/experience/analytics/experienceevent-all.schema.json) so you have to include the Adobe Analytics schema first before you can use it?

clubycoder (changes requested):
A couple of suggestions and a question?

extensions/adobe/experience/analytics/context-data.schema.json
mjasrotia: Will do

extensions/adobe/experience/analytics/context-data.schema.json
mjasrotia: Will do

extensions/adobe/experience/analytics/context-data.schema.json
mjasrotia: I see what you are saying. You know what ? Let me move this property to the experienceevent-all field group. That way, it will always appear when someone adds the experienceevent-all field group to the schema.

PR #1763, opened 2023-08-04
JIRA:CORE-89552 Adding consentResults-v2 to profile union schema

patnaiku:
https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=DMSArchitecture&title=Profile+Consent+v2#ProfileConsentv2-ProposedDesignforConsentv2 

Summary of changes:
Add profile-consentResults-v2 to all Profile union schemas

### DIFF
--- a/extensions/adobe/experience/profile/profile-all.schema.json
+++ b/extensions/adobe/experience/profile/profile-all.schema.json
@@ -132,7 +132,5 @@
       "$ref": "https://ns.adobe.com/xdm/mixins/profile-consentResults-v2"
     }
   ],
-  "meta:status": "stable",
-  "meta:titleId": "profile-all##title##8161",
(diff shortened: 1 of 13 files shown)


extensions/adobe/experience/profile/profile-all.schema.json
```diff
@@ -108,6 +108,86 @@
               }
             }
           }
+        },
```
mjasrotia, line 111: @patnaiku , I know you followed the previous way to adding the consent map to this field group. We can just add the field group ID ("$ref": "https://ns.adobe.com/xdm/mixins/profile-consentResults-v2)  to the allOf section of this field group. Remove the inline property. That way, it will ensure any change in the field group for consent will be propagated automatically to this.

mjasrotia (approved):

PR #1773, opened 2023-09-01
Commerce returns

websterlove:
[SEG-985](https://jira.corp.adobe.com/browse/SEG-985)

### DIFF
--- a/components/datatypes/return.schema.json
+++ b/components/datatypes/return.schema.json
@@ -1,0 +1,35 @@
+{
+  "meta:license": [
+    "Copyright 2017 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/datatypes/return",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Return",
+  "type": "object",
+  "description": "The RMA (Return Merchandise Authorization) issued.",
+  "definitions": {
+    "return": {
+      "properties": {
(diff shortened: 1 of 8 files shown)


components/datatypes/return.schema.json
```diff
@@ -0,0 +1,35 @@
...
+  ],
+  "$id": "https://ns.adobe.com/xdm/datatypes/return",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Return",
+  "type": "object",
+  "description": "The RMA (Return Merchandise Authorization) issued.",
+  "definitions": {
+    "return": {
+      "properties": {
+        "xdm:returnID": {
```
mjasrotia, line 16: Generally, returns are processed against an existing order. The way it is proposed , you will always have to add returns information in the context of an order with the order information in the same payload. What if someone just sends in return event data with no payload for the original order ? We won't be able to link that return with the order. 

I would suggest to add puchaseID in the return header so that we can handle both cases 

Returns coming along with the original  order
Returns coming on their own

websterlove: Thanks. To make sure I understand you right - your suggestion is to add the order ID into this return schema (say, as orderPurchaseID), alongside returnID, so that it's available even if someone uses this datatype independently of the order schema?

What would that mean for when a return is in fact sent along with other order info? A duplicated purchase ID value? Or would it just be adding another option for where to put the purchase ID, and that data could go either or both places (order.purchaseID, order.returns.orderPurchaseID)?

websterlove: @mjasrotia Just making sure GitHub told you about my reply here - thanks :)

mjasrotia: @websterlove , if the complete order is sent along with the return, the source system can decide to send purchaseID in the returns or not but the original order itslelf will have that available. 

Let's add "puchaseID" to be consistent with the original purchaseID

mjasrotia (approved):

PR #1786, opened 2023-10-16
B2B Custom Object Base Class

chrispanella:
This is for B2BCDP on AEP. We need to create a new base class to support Marketo Custom Objects. This is experimental and needs to be on non-production only.


mjasrotia:
@chrispanella , since this is a new B2B object you are trying to add, please work with the XDM and UPS B2B product managers to make sure the business requirements are documented and understood well. B2B has a special handling in UPS and I would also urge to get an approval from the ARB before putting in a PR. Thanks !

chrispanella:
@mjasrotia We are working on a PoC with PM and engineering leadership. This is just needed in non-prod at the moment to ensure it works as UPS has stated it should. I can check with getting approval from the ARB, will ping you directly.

PR #1787, opened 2023-10-16
Add privacyDelete and privacyAccess actions to HygieneOps

perlmonger42:
Add new actions to the HygieneOps dataset to support Privacy Service operations.

https://jira.corp.adobe.com/browse/PLAT-170201 "XDM: Update Hygiene Ops Schema to Support Privacy Access"
https://jira.corp.adobe.com/browse/PLAT-170199 "XDM: Update Hygiene Ops Schema to Support Privacy Delete"

### DIFF
--- a/schemas/hygiene/aep-hygiene-ops-record.schema.json
+++ b/schemas/hygiene/aep-hygiene-ops-record.schema.json
@@ -27,12 +27,14 @@
           "title": "Name of the hygiene operation to be applied by downstream consumers",
           "description": "Name of the hygiene operation to be applied by downstream consumers.",
           "type": "string",
-          "enum": ["deleteIdentity", "updateField"],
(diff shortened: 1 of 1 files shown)


schemas/hygiene/aep-hygiene-ops-record.schema.json
```diff
@@ -122,10 +129,201 @@
...
+              "description": "Represents data-subject requests to have their data deleted",
+              "type": "object",
+              "properties": {
+                "xdm:identities": {
+                  "title": "Set of identities",
+                  "description": "Set of identities whose data must be deleted",
+                  "type": "array",
+                  "items": {
+                    "type": "object",
+                    "additionalProperties": false,
```
mjasrotia, line 144: @perlmonger42 , let's remove all the additionalProperties instances in this PR as we do not have to explicitly set them as false because we should not allow additionalProperties anywhere outside map objects

PR #1789, opened 2023-11-08
DINT-1160: add commerce.requisitionListDeletes xdm

NadiyaS:
Jira ticket https://jira.corp.adobe.com/browse/DINT-1160


shiftedreality (approved):

components/datatypes/marketing/commerce.schema.json
```diff
@@ -144,6 +144,11 @@
           "$ref": "https://ns.adobe.com/xdm/datatypes/requisitionlist",
           "description": "The properties of requisition list created by customer."
         },
+        "xdm:commerce.requisitionListDeletes": {
+          "title": "Delete Requisition List",
+          "$ref": "https://ns.adobe.com/xdm/datatypes/requisitionlist",
```
websterlove, line 149: Isn't this the field that should be a 1 when the event sent is deleteRequisitionList? So it should ref "https://ns.adobe.com/xdm/data/measure" like the "xdm:requisitionListOpens" field does (line 127)?

websterlove: If I'm understanding that correctly, I'd also personally prefer to have it located right under "requisitionListOpens" or at least under "requisitionListRemovals", rather than in the middle of the fields that don't correspond to event types

websterlove (changes requested):

components/datatypes/marketing/commerce.schema.json
```diff
@@ -129,6 +129,11 @@
           "$ref": "https://ns.adobe.com/xdm/data/measure",
           "description": "Initializations of a new Requisition List."
         },
+        "xdm:commerce.requisitionListDeletes": {
```
websterlove, line 132: Didn't catch this yesterday, sorry - should just be "xdm:requisitionListDeletes" - no need for "commerce."

websterlove (changes requested):

websterlove (approved):

mjasrotia (approved):

PR #1790, opened 2023-11-09
Adding a version field to the experienceevent-all schema

apatnaik14:


extensions/adobe/experience/analytics/experienceevent-all.schema.json
```diff
@@ -10113,6 +10113,13 @@
             "type": "string",
             "description": "Context Value."
           }
+        },
+        "https://ns.adobe.com/experience/analytics/version": {
```
clubycoder, line 10117: @apatnaik14 , Since this field will show up in the customer experiences, I would suggest being a little more descriptive for what this field is.  Maybe change the name to `recordVersion` with a description `Adobe Analytics internal event record version. This value will change when records are updated.  An example of this is a GDPR request.`

It may also be worth it to put this under a structure where we could put other internal, not for customer use in the future.  Maybe under `https://ns.adobe.com/experience/analytics/internal`, and you can give that struct a description that calls out that these fields are for internal processing.

clubycoder (changes requested):
A couple suggestions on the field and and description.

extensions/adobe/experience/analytics/experienceevent-all.schema.json
apatnaik14: Sounds good @clubycoder ! I'll go ahead and make those changes right away.

apatnaik14: @clubycoder I have updated the name and description on the version field, and added it under the /internal structure as well. Please do let me know if this looks a bit better. I can totally tweak the description and title for the internal structure and the field further as well.

clubycoder (approved):
👍

mjasrotia (approved):

shonesadler (approved):
LGTM, thanks @apatnaik14

Issue #1795, opened 2023-12-05
Add Field Group for Marketo Engage B2B data


ChrisRobbers:
## What are the schemas that are affected by the issue
This field group is meant to be used in the context of an [XDM Profile schema](https://github.com/adobe/xdm/blob/master/docs/reference/classes/profile.schema.md).

## What are examples of products that are impacted by the issue
Marketo Measure Ultimate and Marketo Engage integration will use this new field group to bring Marketo Engage data into Marketo Measure.

PR #1797, opened 2023-12-11
Add Marketo Engage Person Details extension.

ChrisRobbers:
This change is to add a new Marketo Engage-specific extension that will be used to decorate XDM Profile person records with fields that are unique to Marketo Engage. These fields will be used in Marketo Measure Ultimate to de-anonymize Marketo Measure-sourced web activity data with Marketo Engage Munchkin ID cookies.
Issue: #1795


components/fieldgroups/profile/marketo-engage-person-details.schema.json
```diff
@@ -0,0 +1,41 @@
...
+  "meta:abstract": true,
+  "meta:intendedToExtend": ["https://ns.adobe.com/xdm/context/profile"],
+  "description": "This field group is used to capture Marketo Engage cookie IDs for the B2B Persons.",
+  "definitions": {
+    "marketo-engage-person-details": {
+      "properties": {
+        "xdm:marketoEngage": {
+          "type": "object",
+          "properties": {
+            "xdm:cookieIDs": {
```
anandphatak, line 25: Can you explain the use case for storing the cookie ID in a profile fragment, since this class is intended to extend profile class?

ChrisRobbers: Hi, yes, our product owner has written this document that describes the purpose: https://wiki.corp.adobe.com/x/8g0Utg

components/fieldgroups/profile/marketo-engage-person-details.schema.json
```diff
@@ -0,0 +1,41 @@
...
+    "b2bSchema": true
+  },
+  "meta:extensible": true,
+  "meta:abstract": true,
+  "meta:intendedToExtend": ["https://ns.adobe.com/xdm/context/profile"],
+  "description": "This field group is used to capture Marketo Engage cookie IDs for the B2B Persons.",
+  "definitions": {
+    "marketo-engage-person-details": {
+      "properties": {
+        "xdm:marketoEngage": {
```
anandphatak, line 22: Can you move this under `"xdm:b2b"` category?

ChrisRobbers: Like this?
```
  "definitions": {
    "marketo-engage-person-details": {
      "properties": {
        "xdm:b2b": {
          "type": "object",
          "properties": {
            "xdm:marketoEngage": {
              "type": "object",
              "properties": {
                "xdm:cookieIDs": {
                  "title": "Cookie IDs",
                  "description": "Comma separated Marketo Engage cookie IDs.",
                  "type": "string"
                }
              }
            }
          }
        }
      }
    }
  }
```

ChrisRobbers: @anandphatak will it be an issue that the `XDM Business Person Details` field group already declares `xdm:b2b` at the same level? https://github.com/adobe/xdm/blob/master/components/fieldgroups/profile/b2b-person-details.schema.json#L22

We are unsure if this will cause a conflict, or if the underlying objects/fields would simply be merged together.

ChrisRobbers: I implemented the change, please let me know if it looks good to you.

anandphatak: Yes they will get merged.

anandphatak (approved):

mjasrotia:
@ChrisRobbers , I have got few questions for regarding this requirement.
the "cookieIDs" field you are adding, will it be required for for all existing and new B2B customers or is it a requirement for only a selected ones which use marketo engage.

For B2B customers, there is a pre-provisioning script that is executed which sets all the schemas, their namespaces, identities and relationships at the time of deployment of a B2B sandbox. If this "cookieIDs" field is supposed to part of every B2B implementation then it should be added to that pre-provisioning script as well. Please reach out to the B2B product manager to understand the process.

Additionally,  if this is a needed property for all B2B implementations, can't we just add this to the existing field group "https://github.com/adobe/xdm/blob/master/components/fieldgroups/profile/b2b-person-details.schema.json". You can add it under the definition "b2b-person-details" as a child node to the ""xdm:b2b" object.  

Is there a business need to have this property defined under "marketoEngage" object ? If that is the case, then let's move this field group under this folder "https://github.com/adobe/xdm/tree/master/extensions/adobe/b2b" You can use "https://github.com/adobe/xdm/blob/master/extensions/adobe/b2b/bizible/bizible-person-details.schema.json" field group and its folder structure as reference 

cc: @anandphatak

ChrisRobbers:
> @ChrisRobbers , I have got few questions regarding this requirement. the "cookieIDs" field you are adding, will it be required for for all existing and new B2B customers or is it a requirement for only a selected ones which use marketo engage.
> 
This will be an optional field group for customers who use Marketo Engage. It is intended to supply extra metadata to Marketo Measure Ultimate that will be used to supplement customer's existing integrations from Marketo Engage -> Marketo Measure. It will not be a requirement for any B2B customer, new or existing, but something they will optionally provide in their datasets which we integrate via a Marketo Measure Destination connector.

> For B2B customers, there is a pre-provisioning script that is executed which sets all the schemas, their namespaces, identities and relationships at the time of deployment of a B2B sandbox. If this "cookieIDs" field is supposed to part of every B2B implementation then it should be added to that pre-provisioning script as well. Please reach out to the B2B product manager to understand the process.
> 
I don't think this should be required as per my description above.

> Additionally, if this is a needed property for all B2B implementations, can't we just add this to the existing field group "https://github.com/adobe/xdm/blob/master/components/fieldgroups/profile/b2b-person-details.schema.json". You can add it under the definition "b2b-person-details" as a child node to the ""xdm:b2b" object.
> 
> Is there a business need to have this property defined under "marketoEngage" object ? If that is the case, then let's move this field group under this folder "https://github.com/adobe/xdm/tree/master/extensions/adobe/b2b" You can use "https://github.com/adobe/xdm/blob/master/extensions/adobe/b2b/bizible/bizible-person-details.schema.json" field group and its folder structure as reference
> 
The thinking is that this is the first of potentially more fields that are specific to Marketo Engage, thus the choice to group them under a "marketoEngage" object. We could have just created a field named "marketoEngageCookieIDs" at the "b2b" object level, but this will allow the potential for more fields to be grouped in the future, if that makes sense.

> cc: @anandphatak

ChrisRobbers:
@mjasrotia given the above info, please let me know if I still should move the files to the extensions path. Thanks.

mjasrotia:
@ChrisRobbers , considering that this is a customer specific requirement and since it would not be added to the prte-provisioning script, I think we should move it under the extensions folder and you can take bizible as an example

mjasrotia:
Adding additional notes here from slack thread.
- This data model addition is not customer-specific but solution/vendor-specific. It’s unique to the Marketo->AEP integration. Sounds like it still fits the bill as an extension field. Please correct me if I were wrong.
             You are right, we should model it under extensions. Let's do that.
- Can extension fields be added to a schema and be used just like regular fields from a field group?
             Extension field groups are also available to customers for their schemas. 
- This needs to be added to Marketo extension not Bizible extension.
             I just suggested Bizible as a reference. please create the field group under the marketo folder (see screenshot below). 
             Let's keep the top level field as "marketo" to be used for all future marketo specific fields. We can add more details in 
              the description of the field to provide information about marketo engage. 
             Example :  the field name should be defined as "marketo:cookieIDs" 
- This is an extension to Person Profile class.
             That's fine, we can always have extension field groups extending the profile or any other class.

cc: @anandphatak , @ChrisRobbers

mjasrotia:
@ChrisRobbers , please refresh your master to resolve the conflicts. See the build failure message

ChrisRobbers:
> @ChrisRobbers , please refresh your master to resolve the conflicts. See the build failure message

I have merged in master. Is there anything else you need, or can this be merged?

mjasrotia (approved):

PR #1801, opened 2024-01-10
[PDCL-11518] - Change sunsetTime type to a date-time formatted string

mflorin:
The `sunsetTime` field of the `currentWeather` fieldset must be a date-time formatted string.

### DIFF
--- a/components/fieldgroups/shared/current-weather.schema.json
+++ b/components/fieldgroups/shared/current-weather.schema.json
@@ -124,7 +124,8 @@
                   "$ref": "#/definitions/speed"
                 },
                 "xdm:sunsetTime": {
-                  "type": "integer",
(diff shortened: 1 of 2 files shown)


components/fieldgroups/shared/current-weather.schema.json
```diff
@@ -124,7 +124,8 @@
                   "$ref": "#/definitions/speed"
                 },
                 "xdm:sunsetTime": {
-                  "type": "integer",
+                  "type": "string",
```
mjasrotia, line 127: Changing a data type of an existing field is breaking change. This is not allowed
cc:  @anandphatak

mflorin:
We will find an alternative approach for this. Closing this PR.

PR #1803, opened 2024-01-11
add channel info for commerce

websterlove:
[DINT-1245](https://jira.corp.adobe.com/browse/DINT-1245)

### DIFF
--- a/components/datatypes/marketing/commerce.schema.json
+++ b/components/datatypes/marketing/commerce.schema.json
@@ -154,10 +154,10 @@
           "$ref": "https://ns.adobe.com/xdm/datatypes/commercescope",
           "description": "Commerce scope identifiers of where an event occurred (store view, store, website etc.)."
         },
-        "xdm:channel": {
-          "title": "Channel Info",
-          "$ref": "https://ns.adobe.com/xdm/channels/channel",
-          "description": "Info on the modes through which a shopper is provided content and interacts with the store."
+        "xdm:commerceChannel": {
+          "title": "Commerce Channel",
+          "type": "string",
+          "description": "Mode through which a shopper is provided content and interacts with the store."
         }
       }
     }
--- a/components/datatypes/marketing/commerce.schema.json
+++ b/components/datatypes/marketing/commerce.schema.json
@@ -153,6 +153,11 @@
           "title": "Scope",
           "$ref": "https://ns.adobe.com/xdm/datatypes/commercescope",
           "description": "Commerce scope identifiers of where an event occurred (store view, store, website etc.)."
+        },
+        "xdm:channel": {
+          "title": "Channel Info",
+          "$ref": "https://ns.adobe.com/xdm/channels/channel",
+          "description": "Info on the modes through which a shopper is provided content and interacts with the store."
         }
       }
     }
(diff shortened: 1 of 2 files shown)


components/datatypes/marketing/commerce.schema.json
```diff
@@ -153,6 +153,11 @@
           "title": "Scope",
           "$ref": "https://ns.adobe.com/xdm/datatypes/commercescope",
           "description": "Commerce scope identifiers of where an event occurred (store view, store, website etc.)."
+        },
+        "xdm:channel": {
```
mjasrotia, line 157: Hey @websterlove , we don't need to add this node under commerce. We already have a field group "Channel Details" which you can add to the schema. That is using the channel data type already.

mjasrotia: cc: @anandphatak

websterlove: Thanks for taking a look, @mjasrotia. My reasoning in adding the datatype node under our commerce node was to make the data location more logical for our users to find, and also to avoid requiring them to add another field group. That way they don't have to make retroactive changes to their schemas, but instead will get the changes automatically once the PR is merged and pushed out, which means they won't get errors when we start sending data to these fields. Am I missing/forgetting anything about how this works, that would make using the Channel Details field group the better option in this case?

websterlove: @mjasrotia Are you okay with merging this, given my explanation above? (Are you still the one who would merge it or should I ping someone else?)

mjasrotia: @websterlove , I understand your reasoning and that's a good reasoning. The thing I am concerned about this approach is that we are opening a can of worms to re create similar fields at different levels thereby adding more confusion for future customers. We could have had the same argument to bring productListItems node under commerce and it will make perfect sense but its outside the customers know how to use that. 

Similarily, regarding channel, our customers do know that there is an existing channel FG to add to the schemas but if we add another node under commerce, the existing customers who are already using the channel field group will start seeing this new node appear under commerce all of a sudden. I am sure you understand where I am getting from here.

Had it been a custom field group, we could have taken an exception as it does not impact any other customer but since this is OOTB field group, there are some tradeoffs we need do. I hope I am making some sense here.

cc: @anandphatak

mjasrotia: My apologies for the delay. I am handling many things at the same time.

components/datatypes/marketing/commerce.schema.json
```diff
@@ -153,6 +153,11 @@
           "title": "Scope",
           "$ref": "https://ns.adobe.com/xdm/datatypes/commercescope",
           "description": "Commerce scope identifiers of where an event occurred (store view, store, website etc.)."
+        },
+        "xdm:commerceChannel": {
+          "title": "Commerce Channel",
+          "type": "string",
```
websterlove, line 159: @mjasrotia Here's an example of what I was thinking of, from the Slack conversation. This would be similar to the promotionID field near the top of this file - just a string field directly under the commerce node. We might also do a commerceChannel node and put a couple of fields under it, if there are multiple relevant values we anticipate needing, but I wanted to indicate the general idea first. This shouldn't show up under a tenant namespace as far as I understand.

CC @anandphatak

websterlove: @mjasrotia Will you have time to look at this soon? It's now blocking development and release for this value

PR #1804, opened 2024-01-12
feat: add first draft of Content Asset and Content Experience

ilievlad73:
Please link to the issue #…


components/classes/content-asset.schema.json
```diff
@@ -0,0 +1,149 @@
...
+    },
+    "content-asset-featurization": {
+      "properties": {
+        "xdm:contentFeaturization": {
+          "title": "Content Featurization",
+          "description": "Content featurization.",
+          "type": "object",
+          "properties": {
+            "xdm:orientation": {
+              "title": "Asset Orientation",
```
anandphatak, line 54: Do you want it to be plain string, or an enum? Do you expect any other values besides `portrait`, `landscape` and `square`?

components/classes/content-asset.schema.json
```diff
@@ -0,0 +1,149 @@
...
+  "title": "Content Asset",
+  "type": "object",
+  "meta:extensible": false,
+  "meta:abstract": true,
+  "meta:extends": ["https://ns.adobe.com/xdm/data/record"],
+  "description": "Content asset such as image or video.",
+  "definitions": {
+    "content-asset": {
+      "properties": {
+        "xdm:assetIdentityID": {
```
anandphatak, line 19: Do you want to term it as `assetPerceptionID` than `assetIdentityID` ? Identity and ID somehow denote to the same term. There also an `assetID` in the schema.

components/classes/content-experience.schema.json
```diff
@@ -0,0 +1,193 @@
...
+              "description": "Experience emojis count."
+            },
+            "xdm:textQuotesCount": {
+              "title": "Experience Text Quotes Count",
+              "type": "integer",
+              "description": "Experience text quotes count."
+            },
+            "xdm:wordsCountPerSentence": {
+              "title": "Experience Words Count Per Sentence",
+              "type": "number",
```
anandphatak, line 107: Why is this field not an integer?

components/classes/content-asset.schema.json
ilievlad73: This will be an enum. I will add the enum with: portrait, landscape, square.

components/classes/content-experience.schema.json
ilievlad73: this will be the value of count(words)/count(sentences) from the experience.
Won't be round to an integer in any way tho, so we will expect float values.
Similar to stopWordsRation.

components/classes/content-asset.example.1.json
```diff
@@ -0,0 +1,30 @@
+{
+  "xdm:assetPerceptionID": "1c93aa01-f183-a3f1-bda6-a9c96e66afea",
+  "xdm:assetID": "https://domain/media_random_id.jpeg",
+  "xdm:assetName": "jump.png",
+  "xdm:assetDescription": "A man is performing a mid-air jump on a skateboard at an indoor skate park",
+  "xdm:assetType": "image",
+  "xdm:contentFeaturization": {
+    "properties": {
```
anandphatak, line 8: There can't be a properties node here; could you pl. remove it?

components/classes/content-experience.example.1.json
```diff
@@ -0,0 +1,29 @@
+{
+  "xdm:experienceID": "https://domain/experiments/faq.2023-12-04",
+  "xdm:experienceChannel": "web",
+  "xdm:contentFeaturization": {
+    "properties": {
```
anandphatak, line 5: Same comment as above - remove the "properties" node

components/classes/content-experience.schema.json
anandphatak: Ok. Do you want to add a "minimum": 0 constraint to prevent a negative value? You would probably want to add such a constraint to all "count" fields

components/classes/content-asset.schema.json
```diff
@@ -0,0 +1,221 @@
...
+    "Copyright 2024 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/classes/content-asset",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Content Asset",
+  "type": "object",
+  "meta:extensible": false,
```
anandphatak, line 12: What is the purpose of marking this class as ` "meta:extensible": false`? Marking it false means that you can't create any schema  that extends this class; so how would you use it? Same question of the other class in this PR. Understanding your use case would help us providing guidance.

ilievlad73: This is a mistake, thanks for catching this! I will update it

anandphatak (approved):

PR #1812, opened 2024-01-30
feat: add content classes thumbnail

ilievlad73:
Please link to the issue #…


components/classes/content-asset.schema.json
```diff
@@ -40,6 +40,11 @@
           "title": "Asset Type",
           "type": "string",
           "description": "Asset type such as image or video."
+        },
+        "xdm:assetThumbnailURL": {
+          "title": "Asset thumbnail URL",
+          "type": "string",
```
anandphatak, line 46: Seems like a URL, so do you want to add "format": "uri" or "format": "uri-reference" qualifier to this field?

components/classes/content-experience.schema.json
```diff
@@ -25,6 +25,11 @@
           "title": "Experience Channel",
           "type": "string",
           "description": "Experience channel such as web, mobile or email."
+        },
+        "xdm:experienceThumbnailURL": {
+          "title": "Experience thumbnail URL",
+          "type": "string",
```
anandphatak, line 31: Same comment as above

anandphatak (approved):

Issue #1818, opened 2024-02-15
Mark descriptorTimeSeriesGranularity as stable and add timezone


clubycoder:
## What are the schemas that are affected by the issue

The descriptorTimeSeriesGranularity descriptor that is used with the Summary Metrics Class.

## What are examples of products that are impacted by the issue

Ingestion of paid media data that is aligned to a tiemzone other than UTC.

PR #1819, opened 2024-02-15
[Issue #1818] Mark descriptorTimeSeriesGranularity as stable and add optional timezone property

clubycoder:
This is for issue #1818 

The timezone was added the same way as Place Context that is used in Experience Events:
https://github.com/adobe/xdm/blob/43bf2a1e7e0657621d392307ca9a2a1ce7948665/components/datatypes/placecontext.schema.json#L25


anandphatak:
Change looks good. Do you want to consider updating the example file? https://github.com/adobe/xdm/blob/master/schemas/descriptors/time-series/descriptorTimeSeriesGranularity.example.1.json

clubycoder:
> Change looks good. Do you want to consider updating the example file? https://github.com/adobe/xdm/blob/master/schemas/descriptors/time-series/descriptorTimeSeriesGranularity.example.1.json

Good call...added.

anandphatak (approved):

Issue #1826, opened 2024-02-28
Add "tokens" string array to _experience.decisioning.propositionAction object


krishnanbalasubramanyam:
## What are the schemas that are affected by the issue
proposition-action.schema.json

## What are examples of products that are impacted by the issue
AJO Inbound

PR #1827, opened 2024-02-28
Add "tokens" string array to _experience.decisioning.propositionAction object

krishnanbalasubramanyam:
Please link to the issue #1826 

The tokens are required for AJO Inbound decision item tracking.

### DIFF
--- a/extensions/adobe/experience/decisioning/proposition-action.schema.json
+++ b/extensions/adobe/experience/decisioning/proposition-action.schema.json
@@ -26,6 +26,14 @@
           "title": "label",
           "description": "The customer generated display name of the element with which the end-user interacted.",
           "type": "string"
+        },
+        "xdm:tokens": {
+          "title": "tokens",
+          "type": "array",
+          "description": "Application-generated metadata associated with the element with which the end-user interacted.",
+          "items": {
+            "type": "string"
+          }
         }
       }
     }


extensions/adobe/experience/decisioning/proposition-action.schema.json
```diff
@@ -26,6 +26,14 @@
           "title": "label",
           "description": "The customer generated display name of the element with which the end-user interacted.",
           "type": "string"
+        },
+        "xdm:tokens": {
+          "title": "tokens",
+          "type": "array",
+          "description": "An array of decision item tokens for tracking",
```
ewiener, line 33: Let’s use a more solution-agnostic description consistent with the description for id and labels, e.g. “Application-generated metadata associated with the element with which the end-user interacted.”

krishnanbalasubramanyam: Done.

extensions/adobe/experience/decisioning/proposition-action.schema.json
```diff
@@ -26,6 +26,14 @@
           "title": "label",
           "description": "The customer generated display name of the element with which the end-user interacted.",
           "type": "string"
+        },
+        "xdm:tokens": {
```
anandphatak, line 30: Can this field be named a bit more semantically meaningful?

krishnanbalasubramanyam: Tokens refer to the proposition-interaction tokens. The field "tokens" is enclosed with proposition interaction, just like "label" and "id" that is already present.

anandphatak (approved):

Issue #1828, opened 2024-02-28
Move the trackingToken from option-detail.schema.json to decisionitem.schema.json


kenadobe:
## What are the schemas that are affected by the issue
extensions/adobe/experience/decisioning/option-detail.schema.json
extensions/adobe/experience/decisioning/decisionitem.schema.json

Move the _xdm:trackingToken_ from option-detail.schema.json to decisionitem.schema.json as _xdm:_trackingToken_.
The reason it is prefix with "_" is because this is a placeholder attribute so that customer can find the attribute in AJO Code based experience editor but it will never populate with value. If this is not allowed, i will change it back to xdm:trackingToken.

## What are examples of products that are impacted by the issue
This is for the upcoming new CBE feature which is not being used in current product.

PR #1829, opened 2024-02-28
Feature/cjm 62503 move the xdm:trackingToken from option-detail.schema.json to decisionitem.schema.json

kenadobe:
#1828

### DIFF
--- a/extensions/adobe/experience/decisioning/decisionitem.schema.json
+++ b/extensions/adobe/experience/decisioning/decisionitem.schema.json
@@ -69,6 +69,11 @@
           "$ref": "https://ns.adobe.com/experience/decisioning/calendar-constraint-details",
           "title": "Item Calendar Constraints",
           "description": "Defines the date range during which this decision item is eligible. This constraint may not be strictly enforced always."
+        },
+        "xdm:_trackingToken": {
+          "title": "Tracking Token",
+          "type": "string",
+          "description": "A placeholder attribute to be added to the message to indicate the item interaction will be tracked. The value is meant to be populated by the AJO and customer should not populate with any value."
         }
       }
     }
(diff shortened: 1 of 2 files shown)


usankara (approved):
LGTM!

extensions/adobe/experience/decisioning/decisionitem.schema.json
```diff
@@ -69,6 +69,11 @@
           "$ref": "https://ns.adobe.com/experience/decisioning/calendar-constraint-details",
           "title": "Item Calendar Constraints",
           "description": "Defines the date range during which this decision item is eligible. This constraint may not be strictly enforced always."
+        },
+        "xdm:_trackingToken": {
```
usankara, line 73: missed to notice the path -  let's double check this.

kenadobe: The team decided to change the path to be consistent with all other attributes. I have updated it to "https://ns.adobe.com/experience/decisioning/decisionitem/trackingToken"

kenadobe:
I updated the trackingToken path to be consistent with all other attributes in the schema and also removed the "_" prefix from the trackingToken attribute name.

extensions/adobe/experience/decisioning/option-detail.schema.json
```diff
@@ -104,6 +104,7 @@
           "description": "Option selection type. The selection type of which the item was derived from."
         },
         "xdm:trackingToken": {
+          "meta:status": "deprecated",
```
anandphatak, line 107: @kenadobe , just FYI, marking a field `deprecated` merely means that it will not show up in the UI. But someone can still ingest data in this field.

anandphatak (approved):

PR #1832, opened 2024-03-11
Add phone numbers associated with billing and shipping addresses

websterlove:
[SEG-1302](https://jira.corp.adobe.com/browse/SEG-1302)
For e-commerce, phone numbers are collected together with billing and shipping addresses. These do not map to the existing phone number fields in the Personal Contact Details field group, so this PR adds new fields for these values.


websterlove:
@anandphatak Could you take a quick look at this soon? I'd like to make sure I can make any required changes in time for this PR to be in the next Staging release. Thanks!

anandphatak:
The shipping and billing phone numbers better fit in the commerce field group; but since this field group already had billing and shipping address, we will approve this change.

PR #1837, opened 2024-03-25
Braze Currents Users Event field group updates

isturdy:
This PR includes the latest updates from the Braze Currents Users Events schemas.

### DIFF
--- a/extensions/braze/braze-currents-users-event.schema.json
+++ b/extensions/braze/braze-currents-users-event.schema.json
@@ -20,12 +20,22 @@
         "braze:appID": {
           "title": "App ID",
           "type": "string",
-          "description": "API ID of the app this action occurred on."
+          "description": "API ID of the app on which this event occurred."
         },
         "braze:timezone": {
           "title": "Timezone",
           "type": "string",
           "description": "Timezone of the user."
+        },
+        "braze:appGroupID": {
+          "title": "App Group ID",
+          "type": "string",
+          "description": "API ID of the app group this user belongs to."
+        },
+        "braze:source": {
+          "title": "Source",
+          "type": "string",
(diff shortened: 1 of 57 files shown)


extensions/braze/braze-currents-users-event.schema.json
```diff
@@ -27,29 +27,70 @@
...
+                "P": "Prefer not to say"
+              }
+            },
+            "braze:phoneNumber": {
+              "title": "Phone Number",
+              "type": "string",
+              "description": "Phone number of the user in e164 format."
             }
-          },
-          "required": [
```
anandphatak, line 50: Just want to make sure you want to remove the `required` attribute. Once removed we won't be able to add it back in the future.

isturdy: We do--it isn't required in all events this schema covers, just in the first set we implemented.

extensions/braze/braze-currents-users-event.schema.json
```diff
@@ -240,30 +445,143 @@
...
           }
         },
+        "braze:session": {
+          "title": "Session",
+          "type": "object",
+          "properties": {
+            "braze:duration": {
+              "title": "Duration",
+              "type": "number",
+              "description": "Duration of the session."
```
anandphatak, line 563: Is there an expected time unit? Then maybe it should be included in the description (like days/hours/minutes/seconds/milliseconds). You could also consider changing the field name to say `durationMilliseconds` or `durationSeconds` etc.

isturdy: That's a good idea. Changing the field name is complicated on our end, but I'll update the documentation once I have an answer.

anandphatak (approved):

PR #1843, opened 2024-04-15
B2BAEP-5638: add extra field to account-org dataTypes.

sttwarrior:
https://jira.corp.adobe.com/browse/B2BAEP-5638

Add websiteDomain for b2b intent use cases


sttwarrior:
@anandphatak I don't have access to request a review, but it'll be great if you can take a look. I'm not sure if I need to add the meta fields or it'll be generated by the system.

anandphatak (approved):

PR #1850, opened 2024-06-07
Added  class extending from record as named-record see example  with …

rmunix:
…'@id' and 'xdm:name' named-record.example.1.json

Please link to the issue #…


rmunix:
Closing

PR #1890, opened 2024-09-12
Add new media analytics FG to “Adobe Analytics ExperienceEvent Template”

rdecea:
Please link to the issue #…


rdecea:
Closed this PR; It will be replaced by[this one](https://github.com/adobe/xdm/pull/1917) which adds less number of fields to the temaplate.

Issue #1907, opened 2024-10-29
Modify Media Analytics XDM schema to enable ADC migration


RaduBPavel:
## What are the schemas that are affected by the issue
- `advertisingpoddetails.schema.json`
- `chapterdetails.schema.json`
- `sessiondetails.schema.json`

## What are examples of products that are impacted by the issue
- Adobe Media Analytics

Internal JIRA issue: https://jira.corp.adobe.com/browse/VA-6928

High level changes:
- remove the required property from some fields found in the Advertising Pod Details and Chapter Details schemas
- add a new field in the Session Details schema:`genreList`

andrei2102:
The genreList field addition is an attempt to fix a schema issue without deprecating the media schema and creating a new stable version. The genre already in the schema is a comma separated list and xdm users can't use the gendre values to create breakdowns in CJA.

PR #1908, opened 2024-10-29
[VA-6928] Modify XDM schema to enable ADC migration

RaduBPavel:
Issue #1907


andrei2102 (approved):
the changes are inline with what the media product needs

anandphatak (approved):

PR #1930, opened 2025-01-21
AN-364299:ACA/AGS Shared Global classes

mike2184:
Global classes to be used by GenStudio and Content Analytics for asset and experience schemas.

https://jira.corp.adobe.com/browse/AN-364299

https://wiki.corp.adobe.com/pages/viewpage.action?pageId=3371485681


components/classes/asset.schema.json
```diff
@@ -0,0 +1,91 @@
...
+            "xdm:ID": {
+              "title": "ID",
+              "type": "string",
+              "description": "Unique identifier of the asset."
+            },
+            "xdm:idSource": {
+              "title": "idSource",
+              "type": "string",
+              "description": "The source that created the id for this asset.",
+              "enum": [
```
mike2184, line 43: @aniham @jdelbick - Are there other id sources that should be added to this enum list?

components/classes/asset.schema.json
```diff
@@ -0,0 +1,91 @@
...
+              "title": "Version",
+              "type": "string",
+              "description": "Asset version."
+            },
+            "xdm:perceptionID": {
+              "title": "Perception ID",
+              "type": "string",
+              "description": "Unique identifier of the asset perception."
+            },
+            "xdm:ID": {
```
mike2184, line 34: @anandphatak - Should we have this ID or should we just use the _id inherited from /xdm/data/record

components/datatypes/auditing/auditable.schema.json
```diff
@@ -43,6 +43,12 @@
           "description": "The last dataset files in Catalog which has modified the record. At creation time, `modifiedByBatchID` is set as `createdByBatchID`.",
           "meta:titleId": "auditable##xdm:modifiedByBatchID##title##45371",
           "meta:descriptionId": "auditable##xdm:modifiedByBatchID##description##21001"
+        },
```
mike2184, line 46: @anandphatak - I noticed this auditable data type already exists but is not used in very many places. Would it be best to update this or should we create a new one with this new property?

components/classes/asset.schema.json
mike2184: @anandphatak - Looks like the build linter does not like this enum. What is the correct way to define this?

components/classes/asset.schema.json
anandphatak: You can use inherited _id if the value it will contain is a unique identifier.

components/classes/asset.schema.json
```diff
@@ -0,0 +1,85 @@
...
+              "title": "Name",
+              "type": "string",
+              "description": "Asset name."
+            },
+            "xdm:description": {
+              "title": "Description",
+              "type": "string",
+              "description": "Asset description."
+            },
+            "xdm:type": {
```
anandphatak, line 54: I'll recommend to change the field name to "xdm:assetType" because "type" is a keyword in JSON schema and we don't want unexpected results

components/classes/experience.schema.json
```diff
@@ -0,0 +1,80 @@
+{
+  "meta:license": [
+    "Copyright 2025 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/classes/experience",
```
anandphatak, line 8: Recommend changing the ID. `https://ns.adobe.com/xdm/classes/experience` is very generic. Could you come up with something that is unique to your use case?

components/classes/experience.schema.json
```diff
@@ -0,0 +1,80 @@
...
+  "title": "Experience",
+  "type": "object",
+  "meta:extensible": true,
+  "meta:abstract": true,
+  "meta:extends": ["https://ns.adobe.com/xdm/data/record"],
+  "description": "Digital experience such as a website or social media post.",
+  "definitions": {
+    "experience": {
+      "properties": {
+        "xdm:experience": {
```
anandphatak, line 19: I will recommend changing this field name to something specific to your use case; "xdm:experience" is very generic

components/datatypes/auditing/auditable.schema.json
anandphatak: This datatype is actually used in many places. I am curious, why would you want to include `expirationDate` as a part of audit log? Audit fields have a different purpose, which is to track who created or modified a record. Expiration date field gives me an impression that you want to add some logic to delete the record..

components/classes/asset.schema.json
mike2184: This should be the unique identifier so we will use the record _id

components/datatypes/auditing/auditable.schema.json
mike2184: Sounds like this does not belong here. I will put this in a different datatype as we would like to use some different names for some of the other audit properties as well.

components/classes/digital-asset.schema.json
```diff
@@ -0,0 +1,80 @@
...
+          "title": "Digital Asset",
+          "description": "Digital asset such as an image or video.",
+          "type": "object",
+          "properties": {
+            "xdm:version": {
+              "title": "Version",
+              "type": "string",
+              "description": "Asset version."
+            },
+            "xdm:perceptionId": {
```
anandphatak, line 29: following XDM coding guidelines, please change the field name from `xdm:perceptionId` to `xdm:perceptionID`

components/datatypes/contentaudit.schema.json
```diff
@@ -0,0 +1,61 @@
...
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/xdm/common/contentaudit",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Content Audit",
+  "type": "object",
+  "meta:abstract": true,
+  "description": "Content audit.",
```
anandphatak, line 13: Could you provide a more meaningful description? (that will help in the field discovery)

components/datatypes/featurizationattributes.schema.json
```diff
@@ -0,0 +1,56 @@
+{
+  "meta:license": [
+    "Copyright 2017 Adobe Systems Incorporated. All rights reserved.",
```
anandphatak, line 3:
```suggestion
    "Copyright 2025 Adobe Systems Incorporated. All rights reserved.",
```

anandphatak (approved):

components/classes/digital-experience.schema.json
```diff
@@ -0,0 +1,75 @@
...
+            "xdm:description": {
+              "title": "Description",
+              "type": "string",
+              "description": "Experience description."
+            },
+            "xdm:channel": {
+              "title": "Channel",
+              "type": "string",
+              "description": "Channel such as web or mobile."
+            },
```
aniham, line 48: this can go away - nested inside the source ID

components/datatypes/contentaudit.schema.json
```diff
@@ -0,0 +1,61 @@
...
+            "xdm:lastUpdatedDate": {
+              "title": "Last Updated Date",
+              "description": "Date when this record was last modified.",
+              "type": "string",
+              "format": "date-time"
+            },
+            "xdm:lastUpdatedBy": {
+              "title": "Last Updated By",
+              "description": "User who last modified this record.",
+              "type": "string"
```
aniham, line 42: should this description be "user" or "service"

components/classes/digital-asset.schema.json
```diff
@@ -0,0 +1,80 @@
...
+          "title": "Digital Asset",
+          "description": "Digital asset such as an image or video.",
+          "type": "object",
+          "properties": {
+            "xdm:version": {
+              "title": "Version",
+              "type": "string",
+              "description": "Asset version."
+            },
+            "xdm:perceptionID": {
```
aniham, line 29: i think here we agreed on having additional IDs (one of them perception ID):

```
      "additionalIDs": [
        {
          "platform": "Meta",
          "platformID": "Meta_account2_campaign1_adGroup1_ad1_experience1_mock2"
        },
{
          "platform": "CR",
          "platformID": "perceptionID1234"
        }
      ],
```

components/datatypes/contentaudit.schema.json
```diff
@@ -0,0 +1,61 @@
...
+            },
+            "xdm:expirationDate": {
+              "title": "Expiration Date",
+              "description": "Date when this record will expire.",
+              "type": "string",
+              "format": "date-time"
+            }
```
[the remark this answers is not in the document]
mike2184: It is not standard for everything in XDM as far as I know, but this was something we wanted to include for data retention purposes.

components/classes/digital-experience.schema.json
```diff
@@ -0,0 +1,75 @@
...
+          "title": "Digital Experience",
+          "description": "Digital experience such as a website or social media post.",
+          "type": "object",
+          "properties": {
+            "xdm:version": {
+              "title": "Version",
+              "type": "string",
+              "description": "Experience version."
+            },
+            "xdm:idSource": {
```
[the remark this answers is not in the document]
mike2184: We originally talked about doing that but the platformID would be the exact same value as the _id so it would be redundant. With the current idSource property being a string of ACA or AGS we can know which platform it came from and then just read the _id attribute to get the value.

components/datatypes/contentaudit.schema.json
mike2184: https://github.com/adobe/xdm/pull/1942/files

components/classes/digital-asset.schema.json
mike2184: https://github.com/adobe/xdm/pull/1942/files

PR #1932, opened 2025-01-29
add STO summary dataset schema for reporting

czhongadobe:
https://jira.corp.adobe.com/browse/CJM-87117

### DIFF
--- a/extensions/adobe/experience/intelligentServices/event-journeyai-sendtimeoptimization-summary.schema.json
+++ b/extensions/adobe/experience/intelligentServices/event-journeyai-sendtimeoptimization-summary.schema.json
@@ -68,5 +68,5 @@
       "$ref": "#/definitions/journeyai-send-time-optimization-summary"
     }
   ],
-  "meta:status": "stable"
+  "meta:status": "experimental"
 }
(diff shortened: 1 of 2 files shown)


extensions/adobe/experience/intelligentServices/event-journeyai-sendtimeoptimization-summary.schema.json
```diff
@@ -0,0 +1,72 @@
...
+        }
+      }
+    }
+  },
+  "allOf": [
+    {
+      "$ref": "#/definitions/journeyai-send-time-optimization-summary"
+    }
+  ],
+  "meta:status": "stable"
```
anandphatak, line 71: Since this is a new field group, could you mark it as experimental? That will expose it only to the beta customers. After testing and feedback you can mark it stable.

czhongadobe: Could you be more specific about "expose it only to the beta customers"? Are there some tags in the AEP to distinguish the beta customers, so that only a specific group of customers/orgs can access this XDM while others cannot? If so, how could I get this settings?

anandphatak (approved):

PR #1942, opened 2025-02-11
AN-368516:Add additional ids to allow stitching across different solutions and platforms (AGS vs ACA, Meta, Google, etc)

mike2184:
https://jira.corp.adobe.com/browse/AN-368516

### DIFF
--- a/components/classes/digital-asset.example.1.json
+++ b/components/classes/digital-asset.example.1.json
@@ -4,10 +4,28 @@
     "xdm:perceptionID": "1c93aa01-f183-a3f1-bda6-a9c96e66afea",
     "xdm:_id": "https://domain/media_random_id.jpeg",
     "xdm:idSource": "ACA",
+    "xdm:assetSource": {
+      "xdm:platform": "ACA",
+      "xdm:sourceID": "https://domain/media_random_id.jpeg"
+    },
     "xdm:name": "jump.png",
     "xdm:description": "A man is performing a mid-air jump on a skateboard at an indoor skate park",
     "xdm:assetType": "image",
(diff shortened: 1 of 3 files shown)


components/classes/digital-asset.example.1.json
```diff
@@ -7,7 +7,17 @@
     "xdm:name": "jump.png",
     "xdm:description": "A man is performing a mid-air jump on a skateboard at an indoor skate park",
     "xdm:assetType": "image",
-    "xdm:thumbnailURL": "https://domain/thumbnail.jpeg"
+    "xdm:thumbnailURL": "https://domain/thumbnail.jpeg",
+    "xdm:additionalIDs": [
+      {
+        "xdm:platform": "ACA",
```
aniham, line 13: don't know if we want to mess with enums here for the `platform` value? fine by me not to, but probably more correct to have as enum

aniham:
couple notes, @mike2184  (posted on slack too)

- `perceptionID` should be part of `additionalIDs` imo with platform asset_pool or content_registry
- `idSource `— i had that one in the schema with both ID and source, similar to `additionalIDs`  - are you ok with nesting?
```source: {
  id: ...
  platform: ...
}```
- `thumbnailURL` looks good but i think we should also have sourceURL (will be whatever asset pool gives us on ingestion)
- `assetType` - i think that might be better removed to type and then the `asset` and `experience` schemas become identical

components/classes/digital-asset.example.1.json
mike2184: We discussed this in our teams meeting

anandphatak (approved):

PR #1963, opened 2025-04-10
AMM-2405 : Add Standard Factor Field Group

bvirk3:
* Introduced a new JSON schema for factor standard fields, defining properties such as factor name, value, type, value type, and Granularity.
* Added Default values for Factor Type and Value Type fields
* Added extensive example JSON files demonstrating the usage of the factor standard fields.


components/fieldgroups/factor/factor-standard-fields.schema.json
```diff
@@ -0,0 +1,134 @@
...
+            "Sales Volume",
+            "Customer Satisfaction Score"
+          ],
+          "meta:titleId": "factor-standard-fields##xdm:factorName##title##98732",
+          "meta:descriptionId": "factor-standard-fields##xdm:factorName##description##99700"
+        },
+        "xdm:factorValue": {
+          "title": "Factor Value",
+          "type": "number",
+          "meta:xdmType": "double",
```
anandphatak: `"meta:xdmType": "double"` is not supported; just `"type":"number"` is sufficient

bvirk3: Made the required change. Thx!

anandphatak (approved):

PR #1973, opened 2025-05-12
Add proxy type field

luiliescu:
As of https://jira.corp.adobe.com/browse/PDCL-13533
A new field called proxytype needs to be added which indicates the type of vpn used by clients.


anandphatak (commented):
Approved

PR #1986, opened 2025-06-13
CJM-93844: fixing nesting level

medhakumari:
Please link to the issue #…

### DIFF
--- a/extensions/adobe/experience/decisioning/offerItem.schema.json
+++ b/extensions/adobe/experience/decisioning/offerItem.schema.json
@@ -124,6 +124,7 @@
           "description": "Content References to be associated with offer",
           "type": "object",
           "xdm:type": "object",
+          "meta:status": "deprecated",
           "meta:xdmType": "map",
           "additionalProperties": {
             "title": "Reference Details",
--- a/extensions/adobe/experience/decisioning/offerItem.schema.json
+++ b/extensions/adobe/experience/decisioning/offerItem.schema.json
@@ -124,7 +124,14 @@
           "description": "Content References to be associated with offer",
           "type": "object",
           "xdm:type": "object",
-          "meta:status": "deprecated",
+          "$ref": "https://ns.adobe.com/experience/decisioning/content-references",
+          "meta:status": "deprecated"
+        },
+        "https://ns.adobe.com/experience/decisioning/offeritem/contentReferencesMap": {
+          "title": "Content References",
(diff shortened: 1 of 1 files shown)


usankara (approved):

extensions/adobe/experience/decisioning/offerItem.schema.json
```diff
@@ -122,7 +122,14 @@
         "https://ns.adobe.com/experience/decisioning/offeritem/contentReferences": {
           "title": "Content References",
           "description": "Content References to be associated with offer",
-          "$ref": "https://ns.adobe.com/experience/decisioning/content-references"
```
anandphatak, line 125: Unfortunately, deleting an existing field is a breaking change and cannot be allowed. You can deprecate the field instead.

medhakumari: @anandphatak can we sync on this ?
Basically I want the format 
_experience
decisioning
offerItem
contentReferences <Map[Key, FragmentMetadata]>

Also without merging how can I ensure the structure of how xdm is going to look like?

anandphatak: @medhakumari You will need to make the existing field as `"meta:status": "deprecated"`, and introduce a new field (with a different name) with correct format.

medhakumari: https://github.com/adobe/xdm/blob/ff3a5f3c19c78f34ddff5e86a13c7e5a900a2535/extensions/adobe/experience/decisioning/fragment-metadata.schema.json#L4

extensions/adobe/experience/decisioning/offerItem.schema.json
```diff
@@ -122,7 +122,27 @@
...
```
anandphatak, line 125: Pl. put this line back:
 `"$ref": "https://ns.adobe.com/experience/decisioning/content-references"`
 And then add ` "meta:status": "deprecated"`

extensions/adobe/experience/decisioning/offerItem.schema.json
```diff
@@ -122,7 +122,27 @@
...
+          "type": "object",
```
anandphatak: Remove rest of the lines - from 125 to 133

anandphatak (approved):

PR #1995, opened 2025-07-09
Timestamp Descriptor changes

rrohit25:
Please link to the issue #…


schemas/descriptors/timestamp/descriptorTimestamp.schema.json
```diff
@@ -0,0 +1,51 @@
...
+          "title": "Type",
+          "type": "string",
+          "const": "xdm:descriptorTimestamp",
+          "meta:titleId": "descriptorTimestamp##@type##title##12345"
+        },
+        "xdm:sourceProperty": {
+          "title": "Source Property",
+          "oneOf": [
+            { "type": "string" },
+            { "type": "array", "items": { "type": "string" } }
```
anandphatak: Why would you need an array of properties for timestamp descriptor?

schemas/descriptors/timestamp/descriptorTimestamp.schema.json
```diff
@@ -0,0 +1,44 @@
...
+  "meta:abstract": false,
+  "type": "object",
+  "description": "Identifies a root-level field containing the event timestamp for time-series schemas. The field must be of type Date-Time and required in the source schema.",
+  "definitions": {
+    "descriptor": {
+      "properties": {
+        "@type": {
+          "const": "xdm:descriptorTimestamp"
+        },
+        "xdm:sourceProperty": {
```
anandphatak: This field is already inherited from `"$ref": "https://ns.adobe.com/xdm/common/descriptors/schemadescriptor#/definitions/descriptor"` so you don't need to define it here again. As an example, take a look at `descriptorPrimaryKey.schema.json`

anandphatak (approved):

PR #2001, opened 2025-07-21
Additional fields to support dynamic capping threshold

kmmraju:
We currently support frequency capping using a static threshold to limit offers. However, there is a new requirement to support a dynamic capping threshold that utilizes a PQL rule to determine its value. This request comes from Universal Studios, which wants to leverage ELS to provide the dynamic threshold capping value from a specified dataset.

This PR introduced new fields to support dynamic capping threshold for Offeritem.
JIRA : https://jira.corp.adobe.com/browse/CJM-102540


jabeerulla (approved):
Looks good to me

guayang-adobe (approved):

anandphatak (approved):

PR #2029, opened 2025-08-25
[feature-1979-conversational-experiences] - adding in conciergeID, agentID, and skillID fields

rbuck23:
Please link to the issue #…

PR #2030, opened 2025-08-29
[feature-1979-conversational-experiences] - adding in conversational signals

rbuck23:
https://jira.corp.adobe.com/browse/PLAT-244621

This also includes code from https://github.com/adobe/xdm/pull/2029 so we can close that one if we want to do it all together. Alternatively, I can separate these two PRs.


clubycoder (approved):
👍

anandphatak (approved):

PR #2032, opened 2025-09-01
Adds Lifecycle schema for AJO entities

surbhi114:
https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=CJM&title=Enhanced+customer+alerting+-+Campaign


extensions/adobe/experience/lifecycle/lifecycle.schema.json
```diff
@@ -0,0 +1,97 @@
...
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$id": "https://ns.adobe.com/experience/lifecycle",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Lifecycle Fields",
+  "type": "object",
+  "meta:extensible": false,
+  "meta:abstract": true,
+  "meta:intendedToExtend": ["https://ns.adobe.com/xdm/context/experienceevent"],
```
shagreel: You will need to be able to extend the step events class as well.

extensions/adobe/experience/lifecycle/lifecycle.schema.json
```diff
@@ -0,0 +1,97 @@
...
+          "type": "string",
+          "description": "Unique identifier for the batch instance."
+        },
+        "https://ns.adobe.com/experience/lifecycle/recurrenceIndex": {
+          "title": "Recurrence Index",
+          "type": "integer",
+          "description": "Index value representing the recurrence index of a recurring AJO entity."
+        }
+      },
+      "required": [
```
shagreel: These should not be required. This field group will be included with other field groups on an experience event and not all events will be life cycle events. Therefore these can not be required.

extensions/adobe/experience/lifecycle/lifecycle.schema.json
```diff
@@ -0,0 +1,97 @@
...
+          "title": "Source Type",
+          "type": "string",
+          "description": "Type of the AJO entity (e.g., campaign, journey)."
+        },
+        "https://ns.adobe.com/experience/lifecycle/sourceName": {
+          "title": "Source Name",
+          "type": "string",
+          "description": "Human-readable name of the AJO entity."
+        },
+        "https://ns.adobe.com/experience/lifecycle/isRecurrent": {
```
shagreel: Remove this field

extensions/adobe/experience/lifecycle/lifecycle.schema.json
```diff
@@ -0,0 +1,97 @@
...
+          "title": "Is Recurrent",
+          "type": "boolean",
+          "description": "Indicates if this lifecycle event is part of a recurring AJO entity."
+        },
+        "https://ns.adobe.com/experience/lifecycle/batchInstanceId": {
```
shagreel: Remove this field

extensions/adobe/experience/lifecycle/lifecycle.schema.json
```diff
@@ -0,0 +1,97 @@
...
+          "title": "Batch Instance ID",
+          "type": "string",
+          "description": "Unique identifier for the batch instance."
+        },
+        "https://ns.adobe.com/experience/lifecycle/recurrenceIndex": {
```
shagreel: Remove this field

shagreel (changes requested):

extensions/adobe/experience/lifecycle/lifecycle.schema.json
surbhi114: if the lifecycle schema is gonna be present directly inside _experience, why do we need step events class? Asking because stepEvents is _experience.journeyOrchestration.stepEvents and this one is intended to be _experience.lifecycle

extensions/adobe/experience/lifecycle/lifecycle.schema.json
surbhi114: Dont we need this for recurrent journeys and campaigns. For example, a recurrent Campaign would complete delivery and restart delivery -> there would be 2 separate events fired for this scenario with different recurrenceIndex?

extensions/adobe/experience/lifecycle/lifecycle.schema.json
shagreel: Are you creating a new dataset or adding this to the current step events dataset?

extensions/adobe/experience/lifecycle/lifecycle.schema.json
shagreel: I didn't see any of these things in the examples so i assumed they were added on accident. If they should be here, that's fine. I just didn't have context for them.

extensions/adobe/experience/lifecycle/lifecycle.schema.json
surbhi114: Adding this to current step events dataset. I assumed that just adding the experienceevent context would be sufficient, but I guess not. Will address it.

surbhi114: https://ns.adobe.com/experience/journeyOrchestration/stepEvents/journey to be added

surbhi114: https://ns.adobe.com/experience/journeyOrchestration/stepEvents/journeyStepEvent to be added

shagreel (approved):

PR #2050, opened 2025-10-14
Adding referringSource to channel details for OOTB click-thru tracking

clubycoder:
This is for BRANDCON-1985 & AN-397475
We want to create a standard for referring source that we will use for OOTB Brand Concierge click-through tracking.  A WebSDK tag will be added that will automatically grab whatever is in the page URL source query-string param and store it in this location in the ExperienceEvent.

### DIFF
--- a/components/datatypes/channels/channel.schema.json
+++ b/components/datatypes/channels/channel.schema.json
@@ -124,6 +124,11 @@
           },
           "description": "The types of locations (virtual places) that this channel consists of and can deliver content to.",
           "meta:descriptionId": "channel##xdm:locationTypes##description##91781"
+        },
+        "xdm:referringSource": {
+          "title": "Referring Source",
+          "description": "Referring source to this channel",
+          "type": "string"
         }
       }
     }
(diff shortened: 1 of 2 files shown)


components/fieldgroups/experience-event/experienceevent-channel-referrer.schema.json
```diff
@@ -0,0 +1,46 @@
...
+  "title": "Channel Referral Details",
+  "type": "object",
+  "meta:extensible": true,
+  "meta:abstract": true,
+  "meta:intendedToExtend": ["https://ns.adobe.com/xdm/context/experienceevent"],
+  "description": "Channel referral information such as the referring source.",
+  "definitions": {
+    "experienceevent-channel-referral": {
+      "properties": {
+        "xdm:channel": {
```
anandphatak: There is already a field named "xdm:channel" in the "Channel Details" field group that is intended to extend experience event and this will collide. Could you use a different name for this field?

clubycoder: I intended for it to be the same channel structure and just add a new field under it.  channel is still an object and has the sane name and description.  shouldn't they merge together if you use both field groupd?

anandphatak: As discussed offline let's move this field to the existing "Channel Details" fieldgroup.

Issue #2060, opened 2025-11-07
Add Support for Paid Media normalized schema


ejsuncy:
Add datatypes, fieldgroups, and schemas to support cross-network paid media concepts

PR #2061, opened 2025-11-08
#2060 Add reusable datatypes and examples supporting Paid Media

ejsuncy:
#2060 will involve multiple PRs to keep the review effort manageable. 

This PR adds the initial base datatypes that will be reused across multiple schemas in future PRs to support Paid Media concepts.

Here is the JIRA ticket [AN-421524](https://jira.corp.adobe.com/browse/AN-421524)

### DIFF
--- a/components/datatypes/paid-media/paid-media-additional-field.schema.json
+++ b/components/datatypes/paid-media/paid-media-additional-field.schema.json
@@ -10,34 +10,43 @@
   "title": "Paid Media Additional Field",
   "type": "object",
   "description": "Flexible field structure for storing platform-specific or custom fields not defined in core schema.  Supports multiple value types to accommodate different data types from paid media network APIs.",
-  "properties": {
-    "xdm:fieldName": {
-      "title": "Field Name",
-      "type": "string",
-      "description": "Name of the additional field"
-    },
-    "xdm:stringValue": {
-      "title": "String Value",
-      "type": "string",
-      "description": "String value of the field"
-    },
-    "xdm:numberValue": {
-      "title": "Number Value",
-      "type": "number",
-      "description": "Numeric value of the field"
-    },
-    "xdm:booleanValue": {
-      "title": "Boolean Value",
-      "type": "boolean",
-      "description": "Boolean value of the field"
-    },
-    "xdm:dateValue": {
-      "title": "Date Value",
-      "type": "string",
-      "format": "date-time",
-      "description": "Date/time value of the field"
+  "definitions": {
+    "additionalField": {
+      "properties": {
+        "xdm:fieldName": {
+          "title": "Field Name",
+          "type": "string",
+          "description": "Name of the additional field"
+        },
+        "xdm:stringValue": {
+          "title": "String Value",
+          "type": "string",
+          "description": "String value of the field"
+        },
+        "xdm:numberValue": {
+          "title": "Number Value",
+          "type": "number",
+          "description": "Numeric value of the field"
+        },
+        "xdm:booleanValue": {
+          "title": "Boolean Value",
+          "type": "boolean",
+          "description": "Boolean value of the field"
+        },
+        "xdm:dateValue": {
+          "title": "Date Value",
+          "type": "string",
+          "format": "date-time",
+          "description": "Date/time value of the field"
+        }
+      },
+      "required": ["xdm:fieldName"]
     }
   },
-  "required": ["xdm:fieldName"],
+  "allOf": [
+    {
+      "$ref": "#/definitions/additionalField"
+    }
+  ],
   "meta:status": "experimental"
 }
--- a/components/datatypes/paid-media/paid-media-url-tracking.schema.json
+++ b/components/datatypes/paid-media/paid-media-url-tracking.schema.json
@@ -10,37 +10,46 @@
   "title": "Paid Media URL Tracking",
   "type": "object",
   "description": "URL tracking configuration for paid media campaigns, ad groups, and ads. Captures tracking templates and custom parameters used to track user clicks and conversions.",
-  "properties": {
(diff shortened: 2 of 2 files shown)


components/datatypes/paid-media/paid-media-additional-field.schema.json
```diff
@@ -0,0 +1,43 @@
...
+      "description": "Name of the additional field"
+    },
+    "xdm:stringValue": {
+      "title": "String Value",
+      "type": "string",
+      "description": "String value of the field"
+    },
+    "xdm:numberValue": {
+      "title": "Number Value",
+      "$ref": "https://ns.adobe.com/xdm/data/measure",
```
anandphatak: Any reason to use the data type Measure here? Do you also need the "id" field specified in the Measure? If not, then you could use "type": "number" rather than $ref to Measure data type; that way you won't need to have a nested "value" field under "numberValue".

components/datatypes/paid-media/paid-media-creative.schema.json
```diff
@@ -0,0 +1,187 @@
...
+              }
+            }
+          }
+        },
+        "xdm:assets": {
+          "type": "array",
+          "title": "Creative Assets",
+          "description": "Media assets used in the creative",
+          "items": {
+            "$ref": "#/definitions/asset"
```
anandphatak: Do you reuse the "asset" block defined in the schema in the future? Wanted to find out why it was taken out from the items and defined as a separate block.

components/datatypes/paid-media/paid-media-creative.schema.json
```diff
@@ -0,0 +1,187 @@
...
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "$id": "https://ns.adobe.com/xdm/datatypes/paid-media-creative",
+  "title": "Paid Media Creative",
+  "type": "object",
+  "description": "Data type representing creative assets and content for paid media ads",
+  "definitions": {
+    "creative": {
+      "properties": {
```
anandphatak, line 15: I'd recommend to keep these fields inside an object that acts like a namespace, for example "xdm:paidMediaCreative". You can choose an appropriate name that does not clash; some of the fields defined in this datatype clash with other datatypes in global XDM (for example "xdm:createiveID", "xdm:body", "xdm:destinationURL".) By nesting inside an object will avoid collisions

components/datatypes/paid-media/paid-media-additional-field.schema.json
rayyang15: we thought we might need follow ADC schema on Analytics side. You are right, it turns out we don't need "id" field for paidmedia schema. I changed it back to number type.

rayyang15:
> Could you pl. also add the JIRA ticket number in the PR title? Also I think you need to update the PR number in the description.

Here is the JIRA ticket [AN-421524](https://jira.corp.adobe.com/browse/AN-421524) I don't have permission to change it. 
https://github.com/adobe/xdm/issues/2060 is issue number in description, which is confusing.

components/datatypes/paid-media/paid-media-creative.schema.json
rayyang15: I don't think we might reuse "asset" in the schema, should I move it in items?

components/datatypes/paid-media/paid-media-creative.schema.json
rayyang15: added xdm:paidMediaCreative namespace to avoid collisions.

components/datatypes/paid-media/paid-media-url-tracking.schema.json
```diff
@@ -0,0 +1,46 @@
...
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "$id": "https://ns.adobe.com/xdm/datatypes/paid-media-url-tracking",
+  "title": "Paid Media URL Tracking",
+  "type": "object",
+  "description": "URL tracking configuration for paid media campaigns, ad groups, and ads. Captures tracking templates and custom parameters used to track user clicks and conversions.",
+  "properties": {
```
anandphatak: "definitions" block is missing

components/datatypes/paid-media/paid-media-additional-field.schema.json
```diff
@@ -0,0 +1,43 @@
...
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
+    "of the License at https://creativecommons.org/licenses/by/4.0/"
+  ],
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "$id": "https://ns.adobe.com/xdm/datatypes/paid-media-additional-field",
+  "title": "Paid Media Additional Field",
+  "type": "object",
+  "description": "Flexible field structure for storing platform-specific or custom fields not defined in core schema.  Supports multiple value types to accommodate different data types from paid media network APIs.",
+  "properties": {
```
anandphatak: "definitions" block is missing

rayyang15: done

components/datatypes/paid-media/paid-media-url-tracking.schema.json
rayyang15: done

PR #2068, opened 2025-11-25
Loyalty Challenge Event Schema (CJM-125879)

iewilliam:
This PR introduces an experience event field group for loyalty challenge events, which are events to be processed by the Loyalty Challenge system to advance end users' progress in the loyalty challenges that they have signed up for. 

**Challenges** is an AJO functionality to help customers offer loyalty challenges to their end users and keep track of their progress in them. Completing the challenge would earn the end user some kind of reward such as loyalty points or a coupon for a future discount and such. A few examples of loyalty challenges:
- Visit the store and buy something 3 days in a row
- Buy a certain category of product items (sodas, breakfast items, etc) 3 times
- Spend a total of X dollars over the next week
- Use feature xyz in a photo editing software product n times

Each **challenge** is modeled as a list of **tasks** to complete. 
Each **task** has:
- Numerical target goal (buy 3 sodas, visit 3 times, spend $45.99)
- List of item tags/categories that satisfies the task (sodas, breakfast items, etc)
- List of item tags/categories that are excluded from satisfying the task

The primary purpose of this schema's fields are for evaluation against the challenges'/tasks' rules. The fields in this schema are as follows:
- transactionId - transaction Id for the event. 
- locationId - location Id for the event, this could be for example a store Id. Some challenge tasks for example can require the end user to visit 3 different stores (for example gas stations.)
- localTimezoneOffset - timezone offset of the event/location where the event happens. For example, some challenges may require the end user to do purchases in the morning. 
- itemList - essentially the shopping cart. Includes the items' tags, price, and quantity. Most challenge tasks has item SKU inclusion/exclusion rules or minimum quantity/spend rules. 

Customers will typically transform their Point of Sale Events to this schema before sending them to the Loyalty Challenge system.


extensions/adobe/experience/ajo/loyalty/loyalty-challenge-event-details.schema.json
```diff
@@ -0,0 +1,95 @@
...
+  "description": "Experience event field group for loyalty events to be processed by Loyalty Challenge Service.",
+  "definitions": {
+    "loyalty-challenge-event-details": {
+      "properties": {
+        "https://ns.adobe.com/experience/ajo/loyalty/challenge": {
+          "title": "Challenge",
+          "type": "object",
+          "description": "Loyalty challenge event details.",
+          "properties": {
+            "https://ns.adobe.com/experience/ajo/loyalty/itemList": {
```
anandphatak: Pl. change field name to "xdm:itemList", and every subsequent field in this hierarchy and its children. Only the outermost field should include the `https://ns.adobe.com/experience/ajo/loyalty` part, which is `https://ns.adobe.com/experience/ajo/loyalty/challenge` in this case.

PR #2105, opened 2026-03-07
AN-439651:Cvf:Fix dateTime and Long fields

dcwardell7:
Uploading data to Cvf datasets caused issues for fields of type long and dateTime.

I've updated field definitions to accurately reflect their type.


waykar-prashant (approved):
LGTM

PR #2111, opened 2026-03-18
Mark all the loyalty field groups as stable (CJM-115123)

iewilliam:
Mark all the loyalty field groups as stable so that they will be available in all orgs.


waykar-prashant (approved):
LGTM

PR #2122, opened 2026-03-27
Adding more fields to Loyalty Profile field group (CJM-115123)

iewilliam:
Add more fields to loyalty profile schema to support more loyalty use cases around challenges and rewards, as requested. Also fix tier upgradeDate to be of type date-time.

### Challenges
A loyalty challenge is a time bound challenge that usually has tasks for the user to finish with a reward for finishing on time. A typical example would be spend more than $20 over the next week to receive 100 extra loyalty points.

This PR adds a "challenges" field that lists such loyalty challenges that the user has signed up for and tracks their progress in them.

### Rewards
This PR also adds a rewards section that tracks different reward channels for the user:

- Coupons: Tracks coupons that has been assigned to the user. these are usually time bound and has some kind of redemption limit. 
- Badges: achievement badges earned by the user through various activities.
- Giveaways: These are very much like coupons but to be redeemed against a partner.
- Referrals: Referral rewards earned by the user for referring others to the loyalty program

### DIFF
--- a/components/fieldgroups/profile/profile-loyalty-details.example.2.json
+++ b/components/fieldgroups/profile/profile-loyalty-details.example.2.json
@@ -6,7 +6,7 @@
     "xdm:status": "active",
     "xdm:tier": "silver",
     "xdm:tierExpiryDate": "2025-12-31T23:59:59Z",
-    "xdm:upgradeDate": "2022-07-01T10:00:00Z",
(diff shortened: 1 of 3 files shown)


waykar-prashant (approved):
LGTM

waykar-prashant (approved):
LGTM

PR #2133, opened 2026-04-07
Fix spendInMicroCurrency and bidAmountMicro to use 64-bit long type

### LINKED ISSUE #2132: [Paid Media] spendInMicroCurrency and bidAmountMicro overflow 32-bit integer limit
## Problem

`xdm:spendInMicroCurrency` and `xdm:bidAmountMicro` in the [Core Paid Media Cost Metrics](https://github.com/adobe/xdm/blob/master/components/fieldgroups/paid-media/core-paid-media-cost-metrics.schema.json) field group are defined as `type: "integer"` without a `meta:xdmType`, which causes the platform to treat them as 32-bit integers (max: 2,147,483,647).

Micro currency values are computed as `amount * 1,000,000`, so even a modest spend of ~$2,437 overflows:

```
The value: 2436580000 for field: paidMedia.costMetrics.spendInMicroCurrency
in source type: Long does not fit in type: Integer.
```

Real-world spend values routinely exceed the Int32 ceiling, making these fields unusable at scale.

## Fix

Add `"meta:xdmType": "long"` to both fields to explicitly declare them as 64-bit integers.

Related internal ticket: PLAT-279178

ejsuncy:
## What the change intends to do

Fixes an integer overflow bug in the Core Paid Media Cost Metrics field group. `xdm:spendInMicroCurrency` and `xdm:bidAmountMicro` were defined as `type: "integer"` without `meta:xdmType`, causing the platform to treat them as 32-bit integers (max 2,147,483,647). Since micro currency values are computed as `amount × 1,000,000`, even a ~$2,437 spend overflows and produces an ingestion error.

This change adds `"meta:xdmType": "long"` to both fields, explicitly declaring them as 64-bit integers.

## Breaking changes

None — this is a widening type change (Int32 → Int64). Existing valid values remain valid.

Fixes #2132


waykar-prashant (approved):
LGTM

PR #2143, opened 2026-04-16
CJM-141189: Add optional variation field to external-fragment schema

pradyumnm1:
https://jira.corp.adobe.com/browse/CJM-141189
Adding variation support in External fragment schema to enable customer to use different variants of AEM fragments.


waykar-prashant (approved):
LGTM

PR #2144, opened 2026-04-20
PLAT-280667 Add additional fields in volume metadata table schema

jerryjch:
In the new meta-data.json file provided by the Agent Composer, there are two additional fields at the 'sources' level for source entry:
asset_url
asset_type

Currently the volume metadata schema (for the metadata dataset) does not have these two fields, and the syncMetadata does not fetch and sycn these two fields.

The plan at the moment is to add these two fields to the xdm schema for the metadata dataset. Then we will up the syncMetadata to fetch and store these two fields for the source entry.


waykar-prashant (approved):
LGTM

jerryjch:
Hi, @waykar-prashant I see a Revert PR.
Is this PR reverted?

PR #2167, opened 2026-05-13
Cjm 124855 add supplemental id and supplemental namespace in journey step events schema

DongchaoXu:
Please link to the issue #…


waykar-prashant (approved):
LGTM

PR #2172, opened 2026-05-15
CJM-135243 - Update decisioning-content datatype to use xdm: shorthand field names

kmmraju:
Ticket: https://jira.corp.adobe.com/browse/CJM-135243

Replace full namespace URLs (https://ns.adobe.com/experience/decisioning/*) with standard xdm: shorthand aliases for contentType and content fields in schema definition and example files. This data type is not yet used in stage or prod, so the change is safe.

As part of the testing, we found that using full namespace URLs adds additional roots before the actual fields. In this case, the "experience/deciding" root is prefixed to the content & contentType fields, which is undesired.


waykar-prashant (approved):
LGTM

PR #2177, opened 2026-05-20
CJM-146745: Add messageInteraction tags and label to messaging-channel-context fieldgroup

pushkark08:
CJM-146745

Adding two new fields to the newly created field group


srikanth-bathineni (approved):

waykar-prashant (approved):
LGTM

PR #2180, opened 2026-05-23
Add custom-channel execution metrics mixin for AJO monitoring

mukullamba04:
Adds two new XDM mixins under `extensions/adobe/experience/customerJourneyManagement/` to support execution telemetry for Adobe Journey Optimizer Custom Channels and Integrations:

- `custom-channel-execution-metrics` — extends ExperienceEvent for Custom Channel message feedback events.
- `integration-execution-metrics` — extends ExperienceEvent for Integration execution events (separate event family from message feedback).

## Design context

Two new XDM mixins for Adobe Journey Optimizer (AJO) Custom Channels monitoring. AJO Custom Channels are customer-defined outbound HTTP integrations (e.g. Viber, Slack-as-channel). Each delivery emits a `message.feedback` event; this mixin extends those events with per-call HTTP execution detail (endpoint URL template, method, request timestamp, response time, status code, error code, error reason, throttle-applicable flag, and a Phase-2-reserved waitTime field). Fields stored as templates (with `{{placeholder}}` syntax un-substituted) rather than runtime-interpolated values to control downstream cardinality. Examples cover the success and HTTP-5xx cases.

## Why two mixins

The two pipelines have different shapes:

- Custom Channel events flow through message feedback and have no response cache, so the mixin omits `cacheHit` / `cacheApplicable`.
- Integration events emit on a separate event stream and carry required identity fields (`integrationId`, `integrationName`, `messageExecutionID`), so they live in their own mixin rather than being overloaded onto message feedback.

## Field design notes

- `statusCode` is a raw HTTP integer to preserve drill-down (e.g. 502 vs 504).
- `errorCode` is a free-form string (no enum constraint) so producers can evolve the vocabulary without an XDM schema bump.
- `throttleApplicable` mirrors the `cacheApplicable` config-level-vs-runtime semantic — distinguishes "throttling configured for this call" from "this specific call was throttled".
- `waitTime` is reserved for Phase 2 throttle-wait instrumentation and will be null in Phase 1 emissions.

## Examples

6 example JSON files are included (2 per mixin):
- Custom channel: success + HTTP 5XX failure.
- Integration: success + cacheHit.

## Test plan

- npm test
- npm run lint
- npm run incompatibility-check
- npm run xed-validation
- Verified all 6 examples validate against their mixin schemas locally.


mukullamba04:
Closed in favor of #2181 — clean rewrite with all session decisions consolidated (field renames, URL-template form, free-form errorCode, integration mixin deferred to Phase 1B).

PR #2193, opened 2026-06-02
[PLAT-285474] Define the schema of whatsapp conversation that will be used by downstream consumers (CDP, CJA)

dilipbiswal:
https://jira.corp.adobe.com/browse/PLAT-285474

We have tested using this schema and verified that we can land the whats-app originated data in the lake.

### DIFF
--- a/components/fieldgroups/whatsapp/whatsapp-webhook.schema.json
+++ b/components/fieldgroups/whatsapp/whatsapp-webhook.schema.json
@@ -1,0 +1,827 @@
+{
+  "meta:license": [
+    "Copyright 2026 Adobe Systems Incorporated. All rights reserved.",
+    "This work is licensed under a Creative Commons Attribution 4.0 International (CC BY 4.0) license",
+    "you may not use this file except in compliance with the License. You may obtain a copy",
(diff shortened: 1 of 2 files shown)


dilipbiswal:
cc @anandphatak @waykar-prashant

components/fieldgroups/whatsapp/whatsapp-webhook.schema.json
```diff
@@ -0,0 +1,805 @@
...
+                                    "properties": {
+                                      "xdm:id": {
+                                        "title": "Conversation ID",
+                                        "type": "string"
+                                      },
+                                      "xdm:origin": {
+                                        "title": "Origin",
+                                        "type": "object",
+                                        "properties": {
+                                          "xdm:type": {
```
waykar-prashant, line 718: Should this be an enum?

dilipbiswal: @waykar-prashant can we stay a bit flexible here ? Coz, i don't know the full range of values that can appear on this field. wdyt ?

waykar-prashant (approved):
Rest LGTM

PR #2189, opened 2026-05-29
CJM-148039: Custom Channel + Integration monitoring XDM field groups (Phase 1)

mukullamba04:
## Summary

**Phase 1 scope: Custom Channels + Integration monitoring foundation.** Adds three additive XDM field groups under `extensions/adobe/experience/customerJourneyManagement/`, capturing per-call HTTP execution detail and channel/integration identity for AJO Custom Channel sends.

This is the schema layer for epic **CJM-119729**, paired with:
- [Adobe-CJM/message-runtime#5998](https://github.com/Adobe-CJM/message-runtime/pull/5998) — Scala model + error vocabulary + ChannelConfigMeta
- [Adobe-CJM/message-streaming-worker#96](https://github.com/Adobe-CJM/message-streaming-worker/pull/96) — feedback event producer

Design source-of-truth: wiki 3886615741.

### `customExecutionMetrics` — per-call HTTP-dispatch telemetry (channel-agnostic)

| Field | Semantics |
|---|---|
| `executionType` | Free-form discriminator selecting the companion context group. Phase-1 value `channel`. |
| `endpoint` | URL template (un-interpolated) — cardinality control |
| `method` | HTTP verb |
| `requestTimestamp` | ISO-8601 at HTTP boundary |
| `responseTime` | ms at HTTP boundary only (auth excluded) |
| `waitTime` | Phase-2 reserved (not populated today) |
| `statusCode` | Raw integer HTTP status |
| `throttleApplicable` | Config-applied throttling flag |
| `cacheApplicable` / `cacheHit` | Config-/runtime-level cache signals; channel paths emit `cacheApplicable=false` |
| `errorCode` | Free-form string; producer vocabulary in `ExecutionErrorCode` |
| `errorReason` | Human-readable detail |

### `customChannelContext` — Layer-1/2/3 Custom Channel identity (PRD-aligned names)

| Field | PRD term |
|---|---|
| `customChannelID` / `customChannelVersion` | Custom Channel (Layer 1) |
| `customChannelAuthID` / `customChannelAuthName` | API Credentials (Layer 2) |
| `customChannelConfigurationID` / `customChannelConfigurationName` | Custom Channel Configuration (Layer 3) |

### `integrationContext` — flat integration identity

| Field | Semantics |
|---|---|
| `integrationID` | Integration definition UUID — primary reporting filter |
| `integrationName` | Customer-chosen display name |

## Scope

[...]

PR #2194, opened 2026-06-03
CJM-148040: finalize custom-channel schema — slim customExecutionMetrics to per-recipient results, remove customChannelContext + integrationContext + throttleRate

mukullamba04:
## Summary

Follow-up to #2189 / #2196 / #2198 (all merged). Brings the Phase-1 custom-channel monitoring schema to its finalized state:

1. **Removes** the two just-added context field groups (`customChannelContext`, `integrationContext`) — per-instance channel identity rides on the standard `messageProfile.channel._id` instead.
2. **Slims `customExecutionMetrics`** to the four *per-recipient* delivery results we actually emit (`responseTime`, `statusCode`, `errorCode`, `errorReason`); execution-level config lives on the AJO entity.
3. **Removes `entities/throttleRate`** from `ajo-entity-mixins` `channelDetails` (added by #2198 this release) — deferred until per-call throttle telemetry exists to pair with it.

No consumer impact: every removed field/group is unconsumed today (zero producers, zero consumers). The context groups + `customExecutionMetrics` extras are `experimental`; `throttleRate` is on the stable entity mixin but was added this same release by #2198 and never shipped or consumed, so this pulls it back before it binds.

## Motivation

A delivery-feedback event is **per-recipient, per-attempt**, so only data that *varies by recipient* belongs on it (HTTP status, latency, error). Execution-level config that is *constant across recipients* — endpoint, HTTP method, throttling, channel name — lives once on the AJO entity and is fetched by reporting, not stamped onto every feedback event. This split came out of the runtime design review, and it also drops feedback fields that duplicate the already-stable `message-delivery-feedback` field group.

`throttleRate` is deferred because it carries no reporting value on its own: it has no consumer, and the only way to make it meaningful — correlating it against per-call throttle wait-time — needs telemetry we don't collect in Phase-1. It returns later alongside that.

## Changes

- **Removed** `custom-channel-context` and `integration-context` schemas + examples.
- `custom-execution-metrics.schema.json`:
  - **Kept** (per-recipient results emitted in Phase-1): `responseTime`, `statusCode`, `errorCode`, `errorReason`.
  - **Removed** (execution-level config → AJO entity): `endpoint`, `method`, `throttleApplicable`, `executionType`.
  - **Removed** (response-cache signals — a Data Integration feature; custom channels never cache, Caveat E; returns additively in Phase-1B): `cacheApplicable`, `cacheHit`.
  - **Removed** `requestTimestamp` (duplicates the stable `messageDeliveryfeedback.requestTimestamp`) and `waitTime` (Phase-2 throttle-wait; not populated in Phase-1).
  - `errorCode` carries the **raw** HTTP status (`HTTP_404`, `HTTP_503`), not a `HTTP_4XX`/`HTTP_5XX` bucket; no `meta:enum` (open strings; sample values in examples).
- `ajo-entity-mixins.schema.json` + example — **removed** `entities/throttleRate` (deferred).
- Examples updated to the slimmed shape.

## Validation

- `npm test` — **2407 passing**
- `npm run lint` — clean → `lint-json` CI job green
- `npm run incompatibility-check` — flags **only the intentional removals** above (all unconsumed). This is the last step of the `build` CI job, so the `build` check shows **red by design**; the test + xed-validation steps inside `build` pass. Merging means accepting this advisory flag (or deprecating in place instead of removing).

## Breaking changes

No consumer impact — every removed field is unconsumed (zero producers/consumers). `incompatibility-check` flags the removals as it flags any removal; expected here. The context groups + `customExecutionMetrics` extras are `experimental`; `throttleRate` is on the stable `ajo-entity-mixins` but was added this same release (#2198) and is unshipped/unconsumed.

<details>
<summary>🔍 Per-recipient vs execution-level (the split) + reuse</summary>

A feedback event fires once per recipient per attempt. So:
- **Per-recipient results** — `statusCode`, `responseTime`, `errorCode` / `errorReason` — differ by recipient/attempt and can only live on the event. These stay.
- **Execution-level config** — endpoint (URL template), HTTP method, throttling, channel name — is identical for every recipient of an execution. Keeping it here would duplicate it on every event; it belongs on the AJO entity (one row per execution/channel), which reporting already joins. So it's removed here and carried there (`channelName`/`endpoint`/`httpMethod` via the entity-mixin changes in #2196/#2198).

`requestTimestamp` is dropped under the same reuse-over-reinvent principle: the stable `message-delivery-feedback` field group already defines `requestTimestamp` (and `messageFailure.reason`), so the producer populates those rather than duplicating here.

`cacheApplicable` / `cacheHit` / `waitTime` are also removed: response-caching is a Data Integration feature custom channels never use (Caveat E), and `waitTime` is Phase-2 throttle instrumentation not populated in Phase-1. Both return additively when needed.
</details>


waykar-prashant (approved):
LGTM

PR #2211, opened 2026-06-17
[PLAT-285474][FOLLOW-UP] Fix the casing

dilipbiswal:
Fixes WhatsApp webhook ingestion failures on Stage by correcting field names in the experimental WhatsApp Business Messaging Webhook field group to match the source webhook
  payload structure. Records were failing DCVS validation because required fields were being dropped during mapping.

  Motivation

  WhatsApp webhook ingestion on Stage was failing with DCVS validation errors (required key [metadata] not found). The root cause was a mismatch between the field group's
  property names and the native WhatsApp webhook payload field names, causing mapped fields to be dropped before validation. This fix aligns the field group with the actual
  source payload so that ingestion works with a single top-level mapping without field-level transformations.
  
  Changes

  - components/fieldgroups/whatsapp/whatsapp-webhook.schema.json — corrected property names to match the WhatsApp Cloud API webhook payload structure; removed required
  constraints to allow raw landing of any valid webhook payload
  - components/fieldgroups/whatsapp/whatsapp-webhook.example.1.json — updated example to reflect corrected field names

  Validation

  - npm test — 2406 passing, 0 failing
  - npm run lint — clean
  - npm run validate — zero new failures vs master baseline
  - npm run incompatibility-check — flags property renames as breaking; schema is experimental and no data was successfully landed with the previous field names

  Breaking changes

  Property renames in an experimental schema. No active consumers — ingestion was failing with the previous field names.

### DIFF
--- a/components/fieldgroups/whatsapp/whatsapp-webhook.example.1.json
+++ b/components/fieldgroups/whatsapp/whatsapp-webhook.example.1.json
@@ -8,14 +8,14 @@
           {
             "xdm:field": "messages",
             "xdm:value": {
-              "xdm:messagingProduct": "whatsapp",
+              "xdm:messaging_product": "whatsapp",
               "xdm:metadata": {
-                "xdm:displayPhoneNumber": "+15551234567",
-                "xdm:phoneNumberId": "987654321"
+                "xdm:display_phone_number": "+15551234567",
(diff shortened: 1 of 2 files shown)


waykar-prashant (approved):
LGTM

PR #2229, opened 2026-08-06
Add Send Time Optimization (STO) result fields to messageProfile

naveenpatil1:
## Summary

Adds Send Time Optimization (STO) result fields to the CJM `messageProfile` definition. Non-breaking, additive only.

## Motivation

Downstream STO reporting needs the per-lead STO result and computed send time alongside a message execution. `messageProfile` already exposes `isSendTimeOptimized` to indicate whether STO applied to a message, so the STO result fields are added alongside it there rather than introducing a new object elsewhere.

## Changes

- `extensions/adobe/experience/customerJourneyManagement/messageprofile.schema.json` — new `sendTimeOptimizationType` (integer soft enum, documented via `meta:enum`) and `optimizedSendTime` (date-time string) fields alongside the existing `isSendTimeOptimized`.
- `extensions/adobe/experience/customerJourneyManagement/messageprofile.example.1.json` — example populated for the new fields.

## Validation

- `npm test` — 2408 passing
- `npm run lint` — clean (prettier reports all files use Prettier code style)
- `npm run validate` — zero new failures vs `master` baseline (`messageprofile` is not among the pre-existing failures)
- `npm run incompatibility-check` — clean

## Breaking changes

None.


vedhera (approved):
LGTM!, just a qq We removed the boolean flag?

Prashanthi-Ram (approved):

PR #2207, opened 2026-06-12
Generic, extensible signals for the agentic Conversation Event field group

rbuck23:
## Summary

Closes #2206

Adds a generic, extensible `xdm:attributesMap` to the agentic Conversation Event field group so producers can emit new conversation signals **without a schema change**. **Non-breaking / additive** — the existing `xdm:attributes` signals (`subjects`/`intents`/`tones`/`sentiment`) are deprecated but retained for backward compatibility.

## Motivation

Today `xdm:signals[].xdm:attributes` is a closed set of four named signal types. Every new signal (language, urgency, topic, PII detection, …) needs a schema edit. Agentic/conversational producers emit an open-ended, producer-specific set of derived signals, so a map keyed by signal id is a better fit.

## Changes

- `components/fieldgroups/agentic/conversation-event.schema.json`
  - Add `xdm:attributesMap` — `meta:xdmType: "map"` keyed by signal id; value is an inlined typed signal envelope (`xdm:type` discriminator + `xdm:values[]` of `stringValue` / `numberValue` / `booleanValue` / `confidence` / `qualifiers`). `xdm:type` is a hard enum, fully documented in `meta:enum`.
  - Deprecate the legacy `xdm:attributes` container (`meta:status: deprecated` + description), retained for backward compatibility.
- `components/fieldgroups/agentic/conversation-event.example.4.json`
  - New example demonstrating the map form, including a new `language` signal to show extensibility.

## Validation

- `npm test` — 2408 passing
- `npm run lint` — clean (prettier)
- `npm run validate` — +1 new pass (example.4), **zero new failures** vs the `master` baseline (862/58 with this change vs 861/58 on a clean tree; the 58 are pre-existing, unrelated)
- `npm run incompatibility-check` — clean

## Breaking changes

None. The change is additive; the legacy `xdm:attributes` fields are deprecated, not removed.

Issue #2206, opened 2026-06-12
Generic, extensible signals for the agentic Conversation Event field group


rbuck23:
## Problem

The agentic Conversation Event field group (`components/fieldgroups/agentic/conversation-event.schema.json`) models conversation signals under `xdm:signals[].xdm:attributes` as a **closed set** of four named properties: `xdm:subjects`, `xdm:intents`, `xdm:tones`, and `xdm:sentiment`.

Any additional signal a producer wants to emit — language, urgency, topic, PII detection, etc. — requires a schema change. That does not scale for agentic/conversational use cases where the set of derived signals is open-ended and producer-specific.

## Proposal

Add a sibling `xdm:attributesMap` to each `xdm:signals[]` element — an XDM map (`meta:xdmType: "map"`) keyed by **signal id**, whose value is a typed signal envelope:

- `xdm:type` — `string` | `number` | `boolean` discriminator (documented in `meta:enum`)
- `xdm:values[]` of `{ xdm:stringValue, xdm:numberValue, xdm:booleanValue, xdm:confidence, xdm:qualifiers[] }`

The `xdm:type` discriminator tells consumers which typed value field is populated, satisfying XDM's concrete-type rule while still allowing one schema to carry heterogeneous signals.

This lets producers emit new signal types **without a schema change**. The four existing named signals are deprecated (kept for backward compatibility), so the change is **additive / non-breaking**.

## Scope

- Add `xdm:attributesMap` to the field group (signal value shape inlined under `additionalProperties`).
- Deprecate the legacy `xdm:attributes` container.
- Add `conversation-event.example.4.json` demonstrating the map form, including a new `language` signal.

rbuck23:
Re-opening: PR #2207 (which closed this) is being reverted because the generic `xdm:attributesMap` form is causing a downstream problem. Tracking here so the extensible-signals need can be re-addressed correctly.

rbuck23:
PR #2207 (which closed this) is being reverted in #2230 — the generic `xdm:attributesMap` form is causing a downstream problem. Flagging for a maintainer to re-open this issue so the extensible-signals need can be re-addressed. (I don't have permission to re-open it directly.)

GirishPore:
Reopened as per @rbuck23 is request in this thread https://adobedx.slack.com/archives/C7DQFNVGX/p1786039035888529?thread_ts=1753994238.029039&cid=C7DQFNVGX

PR #2231, opened 2026-08-06
Revert #2207 — remove experimental xdm:attributesMap (violates map-depth guardrail, breaking prod provisioning)

rbuck23:
## Summary

Reverts #2207 (merge commit `87fcade2e`). Re-opens #2206.

This backs out the generic, extensible `xdm:attributesMap` that #2207 added to the
agentic Conversation Event field group. **It is breaking schema provisioning for
conversational agent products (e.g. Brand Concierge) in production** and needs to come
out.

### Why — downstream problem

A platform-side guardrail was deployed that restricts XDM **map nesting depth to 1
level** for most orgs (deeper maps were causing problems in downstream systems; only a
small set of allow-listed orgs may go deeper). Provisioning a schema that contains a
nested map now fails with:

> `Map field exceeds maximum allowed depth for this organization: 1`

#2207's `xdm:attributesMap` is a `meta:xdmType: "map"` whose value envelope contains
`xdm:values[].xdm:metadata` — itself another `meta:xdmType: "map"`. That is a map nested
inside a map (depth 2), which exceeds the new limit. #2207 passed validation on stage
*before* the guardrail landed and then shipped to prod, so the break only surfaced in
production provisioning — affecting everything conversational-related.

The resolution agreed in the discussion thread was to roll back #2207 (rather than relax
the platform constraint). Thread:
https://adobedx.slack.com/archives/C7DQFNVGX/p1786031625889299?thread_ts=1753994238.029039&cid=C7DQFNVGX

Reverting removes both the outer `xdm:attributesMap` and the inner `xdm:metadata` map,
returning the field group to its pre-#2207 (single-level) shape. `xdm:attributesMap`
carried `meta:status: experimental` and was never promoted to `stable`, so removing it
does not break a released stable consumer contract.

## Changes

- `components/fieldgroups/agentic/conversation-event.schema.json` — remove the
  `xdm:attributesMap` map property and its inlined signal envelope; restore
  `xdm:attributes` (drop the `deprecated` status and the "(DEPRECATED …)" description).
- `components/fieldgroups/agentic/conversation-event.example.4.json` — deleted; it only
  demonstrated the reverted map form.

## Validation

- `npm test` — 2407 passing (one fewer than the reverted state; the removed `example.4`
  accounts for the difference)
- `npm run lint` — clean (prettier)
- `npm run validate` — zero new failures vs the `master` baseline (58 pre-existing,
  unrelated to this change)
- `npm run incompatibility-check` — flags removal of `xdm:attributesMap` as breaking.
  **This is expected and intended**: it is a deliberate revert of an additive property,
  and the property was `experimental` (never `stable`), so no released stable contract
  is affected.

## Breaking changes

Removal of the `experimental` `xdm:attributesMap` property (intentional — this is a
revert to unblock production provisioning). No `stable` fields are affected; the legacy
`xdm:attributes` signals are restored unchanged.


rbuck23:
### Note on the failing `xed-validation` CI job

The `xed-validation` job fails at the **"restricted property changes"** step
(`bin/xed-validation/check_properties.js`), with:

```
Error: PR contains disallowed changes to minimum property!
Error: PR contains disallowed changes to maximum property!
```

**This failure is expected and inherent to this revert.** That check diffs the PR's
generated XED against `master` and rejects any removal of a `minimum`/`maximum` bound.
This PR reverts #2207, which removes the `experimental` `xdm:attributesMap` block — and
that block contained:

```json
"xdm:confidence": { "type": "number", "minimum": 0, "maximum": 1 }
```

So the two errors are simply the deletion of that field's `minimum: 0` / `maximum: 1`
bounds. There is no way to faithfully revert #2207 without removing them. This is the
XED-layer counterpart of the `incompatibility-check` flag already documented in the PR
description — both are backward-compatibility guards that fire on any removal, and
neither has an exception for a deliberate revert of an `experimental`, never-`stable`
property.

Since this is an emergency rollback to unblock production provisioning (the nested map
violates the platform's map-nesting-depth guardrail — see the PR description), the
`xed-validation` failure is a false positive here and a maintainer override is
appropriate for merge.

For the record, the `TypeError: Cannot create property 'meta:xdmField' on string
'Properties'` earlier in the log is unrelated, pre-existing noise — it reproduces
identically on the freshly cloned `master` baseline (`tempmaster`) and does not gate the
build.

Prashanthi-Ram (approved):

PR #2238, opened 2026-08-18
Replace conversation signals map with an array of named signals

### LINKED ISSUE #2237: Conversation Event: replace signal attributesMap map with an array of named signals
## Problem

The `Conversation Event` field group (`components/fieldgroups/agentic/conversation-event.schema.json`) models per-scope derived signals as `xdm:attributesMap`, an XDM `map<string, ConversationSignal>` keyed by signal id (`intents`, `sentiment`, `budget`, …).

Because JSON has no map type, batch ingestion reads the map as a struct and then runs a struct→map conversion. When producers populate only the fields relevant to each signal, each key's value is inferred with a **different** shape (e.g. `subjects` has no `confidence`, `intents` does; `budget` is a number, `tones` a string). Reconciling those heterogeneous per-key structs into one map value trips the conversion, producing errors such as:

- `... cannot be converted to a map because a property ... numberValue with varying types (long, number) ...`
- `Missing definition for property 'confidence' at ... attributesMap.<key>.values ...`

Maps are also limited to a single level of depth in AEP, which already forced `xdm:metadata` to be modeled as a key/value array rather than a nested map.

## Proposal

Replace `xdm:attributesMap` with `xdm:signalAttributes`, an **array** of signal objects where the former map key becomes an `xdm:name` field. An array has one uniform element shape by construction, so it ingests cleanly from both JSON and Parquet, needs no struct→map conversion, and is not subject to the map-depth limit — while remaining fully extensible (producers add any `xdm:name` with no schema change).

`xdm:attributesMap` is `meta:status: experimental` and was only recently introduced (#2232), so this replaces it directly rather than carrying a deprecation shim.

rbuck23:
## Summary

Closes #2237

Replaces the experimental `xdm:attributesMap` (`map<string, ConversationSignal>`) in the Conversation Event field group with `xdm:signalAttributes`, an **array** of signal objects where the former map key is carried on an `xdm:name` field. **Breaking** — `xdm:attributesMap` is removed (it was `meta:status: experimental`), not deprecated.

## Motivation

JSON has no map type, so AEP batch ingestion reads `attributesMap` as a struct and runs a struct→map conversion. When producers populate only the fields relevant to each signal, each key infers a different value shape, and reconciling them into one map value fails (e.g. `... cannot be converted to a map because a property ... numberValue with varying types (long, number)`, and `Missing definition for property 'confidence' at ... attributesMap.<key>.values`). Maps are also limited to a single level of depth in AEP, which had already forced `xdm:metadata` into a key/value array.

An array has one uniform element shape by construction: it ingests cleanly from JSON and Parquet, needs no struct→map conversion, isn't subject to the map-depth limit, and stays fully extensible (any `xdm:name`, no schema change). See #2237.

## Changes

- `components/fieldgroups/agentic/conversation-event.schema.json` — `xdm:attributesMap` (map) → `xdm:signalAttributes` (array); add `xdm:name` to the signal object; drop the now-obsolete map-depth note on `xdm:metadata`. Value shape (`xdm:type`, `xdm:values[]` with `stringValue`/`numberValue`/`booleanValue`/`confidence`/`qualifiers`/`metadata`) is unchanged.
- `components/fieldgroups/agentic/conversation-event.example.4.json` — converted to `xdm:signalAttributes`.
- `CHANGELOG.md` — records the breaking change.

## Validation

- `npm test` — 2408 passing, 0 failing
- `npm run lint` — clean (prettier applied)
- `npm run incompatibility-check` — **flags the removal of `xdm:attributesMap`, as expected for this breaking change** (the tool baselines against master, which still has the experimental field). Called out here and in `CHANGELOG.md` for maintainer sign-off.

## Breaking changes

Removes `xdm:attributesMap`. It was `meta:status: experimental` and introduced only recently (#2232), so it is replaced directly rather than deprecated. Recorded in `CHANGELOG.md`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 Breaking Changes
 
-- #2237 Conversation Event field group: reworked the extensible conversation signals into a flat array. Each `xdm:signals[]` entry is now a single signal carrying a required `xdm:scope` plus `xdm:name`, `xdm:type`, and `xdm:values` — replacing both the experimental `xdm:attributesMap` (a `map` keyed by signal id) and the intermediate per-scope grouping layer. A flat array of one uniform shape ingests cleanly from JSON and Parquet, avoiding the struct→map conversion and single-level map-depth limits the map incurred. `xdm:attributesMap` is removed; the legacy `xdm:attributes` object is retained as `deprecated`.
+- #2237 Conversation Event field group: reworked the extensible conversation signals into a flat array. Each `xdm:signals[]` entry is now a single signal carrying `xdm:scope`, `xdm:name`, `xdm:type`, and `xdm:values` — replacing both the experimental `xdm:attributesMap` (a `map` keyed by signal id) and the intermediate per-scope grouping layer. A flat array of one uniform shape ingests cleanly from JSON and Parquet, avoiding the struct→map conversion and single-level map-depth limits the map incurred. `xdm:attributesMap` is removed; the legacy `xdm:attributes` object is retained as `deprecated`.
 
 Fixed Issues
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 Breaking Changes
 
-- #2206 Conversation Event field group: replaced the experimental `xdm:attributesMap` (a `map` of signal id → signal value) with `xdm:signalAttributes`, an array of named signal objects where `xdm:name` carries the value that used to be the map key. An array of a single uniform value shape ingests cleanly from JSON and Parquet, avoiding the struct→map conversion and single-level map-depth limits the map incurred. `xdm:attributesMap` was removed rather than deprecated.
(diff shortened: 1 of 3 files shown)


clubycoder (approved):
👍

Prashanthi-Ram (approved):

rbuck23:
retest

rbuck23:
### CI note for reviewers — two red checks are expected for this intentional breaking change

This PR intentionally removes the experimental `xdm:attributesMap` and replaces it with a flat array of signals on `xdm:signals[]`. Two CI checks go red as a direct result; both are understood:

**1. `incompatibility-check` — expected (the real breaking change).**
The tool baselines against `master`, which still has `xdm:attributesMap` (`meta:status: experimental`, added in #2232). It only exempts `deprecated` fields, not `experimental` ones, so the removal is flagged. This is the deliberate breaking change described in the PR body and `CHANGELOG.md`, and needs a maintainer to accept it.

**2. `xed-validation` restricted-property check — false positive.**
`bin/xed-validation/check_properties.js` errors with:
```
Error: PR contains disallowed changes to minimum property!
Error: PR contains disallowed changes to maximum property!
```
No bound was actually removed or loosened. The check is a line-level diff heuristic: it sees the `minimum:0`/`maximum:1` on `confidence` disappear from its old path under `attributesMap` and doesn't recognize the **identical** bounds re-added at the new flat path. After this PR:
- `confidence` still has `minimum: 0` / `maximum: 1` — now at `conversation.signals[].values[].confidence`
- `sentiment.value` still has `minimum: -1` / `maximum: 1` — unchanged (in the retained, deprecated `xdm:attributes`)

(The `TypeError: Cannot create property 'meta:xdmField' on string 'Properties'` earlier in the log is pre-existing noise — it also reproduces on the clean `tempmaster` clone and does not gate.)

Requesting a maintainer override for both checks. Happy to adjust if you'd prefer the non-breaking route (retain `xdm:attributesMap` as `meta:status: deprecated` alongside the new flat fields), which turns both checks green.

PR #2251, opened 2026-09-02
Stabilize Conversation Event signal fields

### LINKED ISSUE #2250: Conversation Event: promote experimental signal fields to stable
## Summary

In the Conversation Event field group (`components/fieldgroups/agentic/conversation-event.schema.json`), three per-signal fields on each `xdm:signals[]` entry are still marked `"meta:status": "experimental"`:

- `xdm:name` (Signal Name)
- `xdm:type` (Signal Value Type)
- `xdm:values` (Signal Values)

However, the field group itself is already `"meta:status": "stable"`, and the sibling fields on the same signal entry (e.g. `xdm:scope`, and the nested `xdm:stringValue`/`xdm:numberValue`/`xdm:booleanValue`/`xdm:confidence`/`xdm:qualifiers`/`xdm:metadata`) carry no per-field status. The flat per-signal model these three fields belong to has settled, so the leftover `experimental` markers are now inconsistent with the rest of the field group.

## Proposal

Remove the `"meta:status": "experimental"` annotation from `xdm:name`, `xdm:type`, and `xdm:values` so they inherit the field group's `stable` status like their siblings.

## Non-breaking

Additive/annotation-only. No field is added, removed, renamed, or retyped; only the `experimental` marker is dropped. The deprecated `xdm:attributes` field is left untouched (kept `deprecated`).

rbuck23:
## Summary

Closes #2250

Removes the leftover `"meta:status": "experimental"` annotation from three per-signal fields in the Conversation Event field group (`xdm:name`, `xdm:type`, `xdm:values`) so they inherit the field group's already-`stable` status, consistent with their sibling fields. Non-breaking (annotation-only).

## Motivation

The Conversation Event field group is `meta:status: stable`, and the sibling fields on each `xdm:signals[]` entry carry no per-field status. Only `xdm:name`, `xdm:type`, and `xdm:values` were still flagged `experimental` — a leftover from when the flat per-signal model was being introduced. That model has settled, so the markers are now inconsistent with the rest of the field group.

## Changes

- `components/fieldgroups/agentic/conversation-event.schema.json` — removed `"meta:status": "experimental"` from `xdm:name`, `xdm:type`, and `xdm:values`. The deprecated `xdm:attributes` field is left unchanged (kept `deprecated`).

## Validation

- `npm test` — 2417 passing
- `npm run lint` — clean (prettier introduced no reformatting)
- `npm run incompatibility-check` — clean (exit 0)
- `npm run validate` — not run locally; this is an annotation-only change (no constraint/type/enum change) and examples do not reference `meta:status`, so it cannot introduce a new validation failure.

## Breaking changes

None.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/components/fieldgroups/agentic/conversation-event.schema.json
+++ b/components/fieldgroups/agentic/conversation-event.schema.json
@@ -85,7 +85,6 @@
                   },
                   "xdm:name": {
                     "title": "Signal Name",
-                    "meta:status": "experimental",
                     "type": "string",
                     "description": "Identifier for this signal (for example 'subjects', 'intents', 'tones', 'sentiment', or any producer-defined signal). Producers can add new signal types without a schema change.",
                     "meta:titleId": "conversationevent##xdm:name##title##94081",
@@ -93,7 +92,6 @@
                   },
                   "xdm:type": {
                     "title": "Signal Value Type",
-                    "meta:status": "experimental",
                     "type": "string",
                     "description": "Data type of this signal's values; tells consumers which typed value field is populated on each entry of values.",
                     "enum": ["string", "number", "boolean"],
@@ -112,7 +110,6 @@
                   },
                   "xdm:values": {
                     "title": "Signal Values",
-                    "meta:status": "experimental",
                     "type": "array",
                     "description": "One or more values for this signal.",
                     "items": {


clubycoder (approved):
👍

kiranmca04 (approved):

PR #2256, opened 2026-09-11
Add access-log summary field groups (Summary Metrics)

### LINKED ISSUE #2258: Need to add Summary Metrics field groups to expose CDN Log Data & Classifications
We need to add field groups that extend the XDM Summary Metrics class to model aggregated access-log summaries — CDN weblog, mobile, and thin-app requests — for delivery to Customer Journey Analytics as summary data. 

The summary-metrics/cdn-requests-summary schema exists as stable fields, and will be utilized. ($ref and composed). We need a composition that combines:

- CDN Weblog
- Geo Place Lookup
- Device Lookup
- Consent Results
- Bot name, scores, and classification

Geo/device/environment shapes need to reuse the standard placecontext, device, and environment datatypes at their canonical paths, so the summary model matches what the Edge orchestration layer emits on interactive ExperienceEvents — one consistent enrichment model across interactive and summary datasets.

The existing CDN Requests Summary field group is adopted as-is (unmodified) for transport dimensions and the request count.

sandersonb:
# Add access-log summary field groups (Summary Metrics)

Resolves #2258

**Breaking changes:** None. This change is purely additive — three new field groups plus a reference composite schema; no existing component is modified.

## Summary

Adds field groups that extend the **XDM Summary Metrics** class to model aggregated **access-log** summaries — CDN weblog, mobile, and thin-app requests — for delivery to Customer Journey Analytics as summary data. Covers geo/device/environment, bot classification, and data-collection consent.

Geo/device/environment shapes reuse the standard `placecontext`, `device`, and `environment` datatypes at their canonical paths, so the summary model matches what the Edge orchestration layer emits on interactive ExperienceEvents — one consistent enrichment model across interactive and summary datasets.

The existing **CDN Requests Summary** field group is adopted **as-is (unmodified)** for transport dimensions and the request count.

## New field groups (`meta:intendedToExtend: summarymetrics`, `meta:status: experimental`)

- **`access-log-environment-summary`** — `xdm:placeContext`, `xdm:device`, `xdm:environment` via `$ref` to the standard datatypes (geo, IANA timezone, IP-local offset; device hardware; ISP/carrier/domain/connection & proxy type; OS and browser details). The normalized device type is carried in the standard `device.type` soft-enum field (`mobile`/`tablet`/`desktop`).
- **`access-log-bot-detection-summary`** — bot classification as three mutually exclusive, additive request counts: `botRequests` / `uncertainRequests` / `notBotRequests`, derived upstream from customer-chosen thresholds (equal thresholds collapse to a bot / not-bot split), plus an optional `name`. Counts re-aggregate correctly at any CJA grain.
- **`access-log-consent-summary`** — data-collection consent at the standard path `xdm:consents.xdm:collect` via `$ref` to the standard `consent-field` datatype. `xdm:val` uses the Adobe consent choice values (`y`/`n`/`p`/`u`/`dy`/`dn`), used as a GROUP BY dimension so consented vs non-consented traffic can be broken down. `dy`/`dn` (default opt-in / default opt-out) represent the customer's configured default consent, applied to requests with no explicit consent signal.

## Reference composite schema

- **`schemas/access-log/access-log-summary`** (+ example) — composes the Summary Metrics class with the three field groups above and the existing (unmodified) CDN Requests Summary field group; the example is one fully-populated summary record and doubles as the Data Prep mapping target. It is a reference/example schema, not an attachable field group.

## Design notes

- **Class fit.** Summary Metrics is used because the source is pre-aggregated (GROUP BY dimensions + a request count). Aggregation grain is set per flow by the upstream GROUP BY; numeric/continuous fields are never group-by dimensions, so the schema is a permissive superset and the grain lives upstream, not in the schema.
- **Reuse over redefinition.** The interactive `experienceevent-*` field groups can't be reused directly (they `intendedToExtend` ExperienceEvent), but the underlying datatypes (`placecontext`, `device`, `environment`, `consent-field`) are class-agnostic and are reused here at their standard paths.
- **Bot counts, not scores.** A raw or pre-averaged bot score can't be re-aggregated across summary rows; mutually exclusive counts can. Thresholds are applied upstream.
- **Consent reuses the standard.** No new consent vocabulary — the standard `choice-value` enum already includes `dy`/`dn` for default consent. The per-IMSOrg default is applied in the Ghostwire mapping layer, not the schema.
- **PII / IP.** Client IP is intentionally excluded from summary output. The shared `environment` datatype still declares `ipV4`/`ipV6` (fields can't be stripped from a `$ref`'d standard datatype); exclusion is enforced at the Data Prep mapping layer by never mapping source IP into those paths.
- **Soft enums** (`device.type`) accept new values without a schema change, matching existing XDM practice.
- **`meta:titleId`/`meta:descriptionId`** are omitted on new fields, matching current contribution practice; they are generated by the localization pipeline.

## Testing

`npm test` — all passing (schema meta-validity, example presence, and example validity for every new schema, including the composite reference schema).

## Deferred / follow-ups

- **Byte metrics** (`requestBytes`/`responseBytes`) were considered as additions to CDN Requests Summary but are **deferred** to avoid modifying another team's field group; may be revisited later (would need that team's sign-off at that time).
- **Consent scope** is data collection (`collect`) only; `share` can be added if data-sharing consent is needed.
- **Raw device hardware type** is not carried in the summary (the normalized `device.type` is the intended low-cardinality dimension); revisit only if both raw and normalized are needed in the same row.


clubycoder (approved):
👍

